PR #4, opened 2018-10-09
[FIX] Some stylistic comments to #1 and nit picks

oesteban:


oesteban:
@satra I can't assign @djarecka as a reviewer

engine/auxiliary.py
```diff
@@ -2,7 +2,7 @@
 import inspect, os
 import logging
 logger = logging.getLogger('nipype.workflow')
-from nipype import Node
+from .node import Node
```
djarecka: This is the `Node` current nipype and I don't have `node.py` 

This is only used to my wrapper for the current interfaces. I'll work now to connect the current interfaces directly, but for now it `CurrentInterface` is using nipype.Node

pydra/engine/node.py
```diff
@@ -1,39 +1,55 @@
...
+
+        Parameters
+        ----------
+
+        name : str
+            Unique name of this node
+        mapper : str or (list or tuple of (str or mappers))
+            Whether inputs should be mapped at run time
+        inputs : str or list of str
+            States this node's input names
```
djarecka: this is now dictionary, but I can update it with other things

djarecka:
@oesteban - I've fixed all tests and created a PR to your fork

effigies:
Merged master to get coverage reports.

PR #5, opened 2018-10-09
worker improvements + rm python2 things + workingdir

kaczmarj:
This PR proposes 
- adding a function `get_worker_class` to the `workers` namespace in order to get the worker based on requested plugin.
- removing python2-specific things (e.g., `__future__`)
- setting a default `workingdir` which is not `None`. using `None` raises errors in things like `os.path.join`.
- renaming `ready2run` to `ready_to_run`. the former makes it sound like you are converting `ready` to `run` (nitpicking, sure).
- enabling workers to be context managers (i.e., used with the `with` statement). i realized that the workers have `close` methods, and it would be helpful to use workers in a `with` statement so the worker can close itself on any errors.

For `workingdir`, instead of `'.'`, I might suggest making a temporary directory if the user does not provide `workingdir`. this is what tensorflow does and it seems intuitive to me.


effigies:
Merged master for tests and coverage.

effigies:
@kaczmarj Can you merge master, and revert the `workingdir` changes for now? @djarecka says they're causing problems, so they should probably be their own PR.

kaczmarj:
@djarecka - i will create a different branch and open a new PR specifically for the working dir change.

i will probably open a separate pr to make the workers context managers so we can use them with `with` (and so the worker instances can clean themselves up on errors and exiting).

PR #8, opened 2018-10-09
CI: Add coverage/codecov

effigies:


pydra/engine/tests/test_newnode_neuro.py
```diff
@@ -42,6 +46,7 @@ def select_target(subject_id, space):
     """ Given a source subject ID and a target space, get the target subject ID """
     return subject_id if space == 'fsnative' else space
 
+@pytest.mark.skipif(no_fmriprep, "No fmriprep")
```
djarecka: this should be `reason="No fmriprep"`

PR #11, opened 2018-10-10
MNT: Drop Python 2 vestiges

effigies:
Trying to figure out why #5 is failing. Checking Python 2 fixes.


effigies:
Wasn't that. Merging.

djarecka:
@effigies I think that #5 is failing because some inconsistency in `self.workingdir`, but haven't had time to debug it, will try tomorrow.

Issue #10, opened 2018-10-10
CI: Add PEP8 checks


effigies:
Would be good to start off on the right foot, here.

satra:
have been using black, but haven't automated the process.

effigies:
Looks like `black --check` can be used in place of `flake8`.

satra:
let's remove `yapf` configuration and switch to black. it should only be a development dependency not a runtime dependency.

djarecka:
@satra - i added 1pred-commit-config` in this pr: https://github.com/nipype/pydra/pull/58

I'm using `pre-commit`,  so `black` is running on my laptop for every single commit

PR #12, opened 2018-10-10
[WIP] STY: Flake8 cleanup

effigies:
This is purely a style cleanup. No refactors or changes in functionality are being proposed.

First step for #10.

### DIFF
--- a/pydra/engine/auxiliary.py
+++ b/pydra/engine/auxiliary.py
@@ -1,8 +1,10 @@
-import pdb
-import inspect, os
+# import pdb
+import os
+import inspect
 import logging
-logger = logging.getLogger('nipype.workflow')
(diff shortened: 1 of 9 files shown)


effigies:
This is going to break all other PRs. @oesteban if you want to get yours in first, that's fine with me.

rmarkello:
Thoughts on adding `flake8` to the test suite as a requirement for new contributions? As you said, would be good to start off on the right foot!

effigies:
@rmarkello Yup, that would be great, but figured we should fix the issues first.

satra:
should we just ask folks to use `yapf`, like we do for nipype?

effigies:
Closed in favor of #15.

PR #13, opened 2018-10-10
MNT: Revert to Codecov defaults for thresholds, etc

effigies:


effigies:
Doesn't seem to cause problems. Merging.

Issue #14, opened 2018-10-10
Remove hard-coded data paths


effigies:
See, e.g.,

https://github.com/nipype/pydra/blob/ce19c71ef8413e8987ac2e4733fb5a0cff4238cf/pydra/engine/tests/test_newnode.py#L1295

https://github.com/nipype/pydra/blob/ce19c71ef8413e8987ac2e4733fb5a0cff4238cf/pydra/engine/tests/test_newnode_neuro.py#L34-L40

We should use datalad or OSF to retrieve data to run tests on.

djarecka:
yes, we probably should create a docker with freesurfer and some data

effigies:
Not fully resolved by #19. Reopening.

satra:
@djarecka - have we removed all tests that depend on fsl and other tools. we should at this point strip out any dependency on nipype or any other neuroimaging package.

djarecka:
and where do you see these dependencies?

satra:
see the original post

djarecka:
but we don't have these tests anymore

PR #15, opened 2018-10-10
[ENH] yapf-ify code + add .gitignore

kaczmarj:
This PR proposes:
- adding a `.style.yapf` file and updating the code according to `yapf`
- adding a `.gitignore` file

The `.style.yapf` file is an explicit listing of pep8 style. The changes that `yapf` made are reasonable, but please let me know if you don't like any of the style.

to run `yapf`, I ran this on the command line:
```
yapf -rip pydra
```

This will change files in-place and in parallel.


kaczmarj:
This relates to https://github.com/nipype/pydra/pull/12#issuecomment-428570385

Adding the `.style.yapf` file will prevent most (all?) user-specific differences in yapf style.

pydra/_version.py
```diff
@@ -1,4 +1,3 @@
-
```
effigies: Can we exclude this from `yapf`? This is an auto-generated file.

kaczmarj:
when merged, can these commits please be squashed?

effigies:
Is this ready to add a flake8 check in Travis? I think we might need #14 first.

PR #19, opened 2018-10-14
TEST: Get data files for Node/Workflow tests

### LINKED ISSUE #14: Remove hard-coded data paths

effigies:
* [x] Get ds000114 files for `test_newnode`.
* [ ] Get fMRIPrep intermediate files for `test_newnode_neuro`.

Closes #14.


effigies:
@djarecka Could you tar up the files you're using for testing the surface sampling workflow?

effigies:
Looks like `test_current_wf_2a` is timing out. Any thoughts?

effigies:
I've pulled down the relevant files for surface sampling. I'm getting the following printed in a loop:

```
_NODES_CHECK INPUT set_xfm_source False
_NODES_CHECK INPUT sampler False
_NODES_CHECK INPUT sampler False
_NODES_CHECK INPUT sampler False
_NODES_CHECK INPUT sampler False
```

I'm assuming this shouldn't happen dozens of times?

djarecka:
@effigies - `_NODES_CHECK INPUT sampler False` can be printed multiple times. Do you still want em to create a tar from my files?

effigies:
Might as well. Using my own data wasn't working for me.

djarecka:
@effigies - I kept the directory structure, but included only the files that are used in the test:
https://www.dropbox.com/s/rrxc5kegeuzkbtr/fmriprep_test.tar?dl=0

effigies:
Thanks. I'll try to have a look tonight. I think if we want to release today, what we should do is settle on the theoretical, environment-variable-based locations of these files, make sure that tests skip on the absence of the data, and just hit publish. Then we can make fixing the tests a 0.1.0 target.

effigies:
(Also, in meetings all day, so very limited ability to do more than comment.)

djarecka:
I'll try to look later why `test_current_wf_2a` is failing for `serial`. Hardly ever `serial` fails and others work... Want to also finish my PR

djarecka:
also I can create a datalad repo from the fmriprep files, so we can test it with travis, but want to concentrate on other things first, will create an issue to remember

effigies:
Yeah, we can skip datalad on that for now. I also want to reduce the FreeSurfer files to the minimum necessary to perform sampling.

djarecka:
I don't know how to recreate problem with `test_current_wf_2a`..:(

pydra/engine/tests/test_node.py
```diff
@@ -1441,8 +1449,7 @@ def test_workflow_16a(plugin, change_dir):
 # testing CurrentInterface that is a temporary wrapper for current interfaces
 
 
-@pytest.mark.skipif(
-    not os.path.exists("/Users/dorota/nipype_workshop/data/ds000114"), reason="adding data")
+@pytest.mark.skipif(not DS114_DIR.exists(), reason="Missing $PYDRA_DATA_DIR/ds000114")
```
djarecka: you called the dir `PYDRA_TEST_DATA`

djarecka:
ok, I know what is wrong with this test, you didn't change paths: https://github.com/nipype/pydra/pull/19/files#diff-2f4c9a06e6e1c89b63af0c7ec864cbe5R1659
not sure why travis doesn't return the trait error

I was wrong that it worked for mp/cf/dask, travis didn't even try. Of course, it worked on my laptop ;-)

djarecka:
@effigies - are you still working on this. If not, I'll merge it and rest of the test we can do in a new PR. We will have  a lot of conflict anyway, since all PRs are working on the same files, so might be better to keep PR relatively small if possible

effigies:
Sounds good. We can do a separate PR.

PR #22, opened 2018-10-15
ENH: absolute imports + dependency checker + style

kaczmarj:
This PR proposes

- using absolute imports
- adding a function decorator `pydra.engine.utils._check_dependencies` to verify that python dependencies are installed (e.g., dask and distributed for `DaskWorker`)
- function `pydra.engine.workers._get_worker` to get a worker object from a string (cleans up the code in `Submitter`).
- changes to style (e.g., using `isinstance(x, y) instead of `type(x) is y`.
- tests for `pydra.engine.utils._check_dependencies` and `pydra.engine.workers._get_worker`.

### DIFF
--- a/pydra/engine/auxiliary.py
+++ b/pydra/engine/auxiliary.py
@@ -1,7 +1,10 @@
-import pdb
-import inspect, os
+"""Auxiliary elements."""
+import inspect
+import os
 import logging
+
 from nipype import Node as Nipype1Node
+
 logger = logging.getLogger('nipype.workflow')
 
 # dj: might create a new class or move to State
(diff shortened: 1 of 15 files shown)


pydra/engine/auxiliary.py
```diff
@@ -1,7 +1,10 @@
-import pdb
```
djarecka: I like keeping this on the files, at least when we are changing the code so much.

pydra/engine/tests/test_newnode.py
```diff
@@ -25,8 +27,7 @@ def move2orig():
     request.addfinalizer(move2orig)
 
 
-Plugins = ["serial"]
-Plugins = ["serial", "mp", "cf", "dask"]
+plugins = ["serial", "mp", "cf", "dask"]
```
djarecka: I like using capital letter for global var, we can always use all caps names if people prefer

pydra/engine/tests/test_newnode_neuro.py
```diff
@@ -28,12 +29,10 @@ def move2orig():
...
 
-import pdb
-
-Name = "example"
+name = "example"
 DEFAULT_MEMORY_MIN_GB = None
-# TODO, adding fields to Inputs (subject_id)
-Inputs = {
+# TODO, adding fields to inputs (subject_id)
+inputs = {
```
djarecka: the same as higher

pydra/engine/workers.py
```diff
@@ -68,8 +92,12 @@ def close(self):
 
 
 class DaskWorker(Worker):
+    @utils._check_dependencies("dask", "distributed")
```
djarecka: this might be also useful for mp and cf

kaczmarj: `multiprocessing` and `concurrent.futures` are part of the standard python3 library so everyone should have them

pydra/engine/tests/test_newnode.py
kaczmarj: i would prefer all caps for global. title case makes it look like a class. i'll change it to all caps.

pydra/engine/auxiliary.py
kaczmarj: if you're using python 3.7, we can use the new builtin `breakpoint()` instead of `import pdb ... pdb.set_trace()`. then we don't have to import `pdb` and have that be committed. what do you think?

pydra/engine/workers.py
djarecka: makes sense

pydra/engine/tests/test_newnode.py
djarecka: ok

pydra/engine/auxiliary.py
djarecka: I'm using various python. you can remove it if you want, but I won't promise that I won't add it (and forget to remove) again, I'm using pdb all the time. 

And I'm happy to remove before release

effigies:
Is there a good reason to use absolute imports? I would generally move the other direction.

djarecka:
I used relative imports because this is used in nipype. I like absolute imports, but my understanding is that for testing might be a bit safer in case you might have some conflicts. @effigies - is that why you prefer to keep relative imports or something else?

satra:
@djarecka  and @kaczmarj - explicit relative imports are what we do prefer within a package. much fewer chances of clash.

pydra/engine/auxiliary.py
effigies: I'll put in a vote for removing `pdb` and `set_trace` (and `breakpoint`, in Py37) from the main tree.

pydra/engine/node.py
```diff
@@ -1,19 +1,38 @@
-"""Basic compute graph elements"""
+"""Basic compute graph elements."""
```
effigies: This isn't really a sentence, more of a header. I don't think we need punctuation.

We will probably want to expand the docs at some point.

pydra/engine/auxiliary.py
djarecka: just a note, we don't want to have py37 only. So I'm testing on py36 or py35 and I think that's fine

effigies:
Please tag this with whether you want to get this in for the initial release (0.0.1) or 0.1.0. Even if it's 0.1.0, if it's ready to merge before 0.0.1 is closed, we can still do it.

kaczmarj:
I will close this.

PR #28, opened 2018-12-07
[WIP] enh: create a base task class with help from oesteban and effigies

satra:
moved to PR - so we can track contributions. google drive stores only the last 100 edits.

TODO:
- [ ] Requirements
  - [ ] Shell
  - [ ] Container
   - [ ] Docker
   - [ ] Singularity
- [ ] Versioning
- [x] Resource monitoring
- [x] Provenance
- [ ] TaskExtensions
  - [x] Shell
  - [ ] MATLAB
  - [ ] Container
   - [ ] Docker
   - [ ] Singularity
- [ ] Checksums
  - [X] Input
  - [x] Input with functions
  - [ ] Input with files
  - [ ] Task object
  - [ ] RuntimeRequirements


pydra/engine/task.py
```diff
@@ -0,0 +1,443 @@
...
+    return dt.datetime.utcnow().isoformat(timespec='microseconds')
+
+
+prov_context = "https://openprovenance.org/prov.jsonld"
+schema_context = "https://schema.org/docs/jsonldcontext.json"
+pydra_context = {"pydra": "https://uuid.pydra.org/"}
+
+# audit flags
+AUDIT_PROV = 0x01
+AUDIT_RESOURCE = 0x02
```
effigies: If we're 3.6+, we can use [`enum.Flag`](https://docs.python.org/3/library/enum.html#flag).

```Python
class AuditFlag(enum.Flag):
    PROV = enum.auto()  # 0x01
    RESOURCE = enum.auto()  # 0x02
```

pydra/engine/task.py
```diff
@@ -0,0 +1,443 @@
...
+    output: ty.Optional[ty.Any] = None
+
+
+class BaseTask:
+    """This is a base class for Task objects.
+    """
+
+    _api_version: str = "0.0.1"  # Should generally not be touched by subclasses
+    _task_version: ty.Optional[str] = None  # Task writers encouraged to define and increment when implementation changes sufficiently
+    _version: str  # Version of tool being wrapped
```
effigies: Should this be optional, or potentially a range of versions? Or is the idea that tasks *should* be highly constrained to a specific version, and my way of indicating that this task still works for a new version is:

```Python
class FLIRT6(FLIRT509):
    _version = '6.0.0
```

pydra/engine/task.py
```diff
@@ -0,0 +1,443 @@
...
+        odir = self.cache_dir / inputs_hash
+        odir.mkdir(parents=True, exist_ok=True)
+        #check_runtime(self._runtime_requirements)
+        #isolate inputs if files
+        #cwd = os.getcwd()
+        #id = record_provenance(self, env)
+        #resources = start_monitor()
+        result = Result(output=None)
+        try:
+            result.output = self._run_interface(**kwargs)
```
effigies: Do we still pass a `runtime` object (or `_runtime_requirements`) to `_run_interface`? I'm thinking in terms of the existing `_pre_run_hook` and `_post_run_hook`, but perhaps that's a problem for a later day.

pydra/engine/task.py
```diff
@@ -0,0 +1,443 @@
...
+    _runtime_requirements = RuntimeSpec()
+    _runtime_hints = None
+
+    _input_sets = None  # Dictionaries of predefined input settings
+    _cache_dir = None  # Working directory in which to operate
+    _references = None  # List of references for a task
+    
+
+    def __init__(self, inputs: ty.Optional[ty.Text]=None,
+                 audit_flags: bool=False,
```
effigies: I think this should probably be `int` (or `AuditFlag`, if we go with that).

pydra/engine/task.py
satra: implemented.

pydra/engine/task.py
```diff
@@ -0,0 +1,412 @@
...
+        # start recording provenance
+        aid = gen_uuid()
+        self.audit({"@id": "pydra:{}".format(aid),
+                    "startedAtTime": now()},
+                   AuditFlag.PROV)
+          
+        # Not cached        
+        if self._cache_dir is None:
+            self.cache_dir = mkdtemp()
+        odir = self.cache_dir / inputs_hash
```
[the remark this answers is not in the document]
satra: i'm still contemplating this. but the checksum will replace this and will be a function of inputs + task + (environment?).

pydra/engine/task.py
```diff
@@ -0,0 +1,412 @@
...
+import os
+import cloudpickle as cp
+from tempfile import mkdtemp
+import typing as ty
+import inspect
+import asyncio
+from pathlib import Path
+
+File = ty.NewType('File', Path)
+Directory = ty.NewType('Directory', Path)
```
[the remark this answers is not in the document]
satra: i know @oesteban - changed them. i'm not completely sure yet. but i think this would work.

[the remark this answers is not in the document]
oesteban: Why `str`s? `Path` does not enforce the existence of the path, but will complain if it is not a valid path.

pydra/engine/task.py
```diff
@@ -0,0 +1,393 @@
...
+    @property
+    def output_names(self):
+        return [f.name for f in dc.fields(self.output_spec)]
+
+    @property
+    def version(self):
+        return self._version
+    
+    def save_set(self, name, inputs, force=False):
+        if name in self._input_sets and not force:
```
effigies: Here's a place where you assume `_input_sets is not None`.

pydra/utils/profiler.py
```diff
@@ -0,0 +1,319 @@
...
+_MB = 1024.0**2
+
+
+class ResourceMonitor(threading.Thread):
+    """
+    A ``Thread`` to monitor a specific PID with a certain frequence
+    to a file
+    """
+
+    def __init__(self, pid, freq=5, logdir=None, fname=None, python=True):
```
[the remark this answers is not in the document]
effigies: In this context, I'd find interval more intuitive than period. But either way.

```suggestion
    def __init__(self, pid, interval=5, logdir=None, fname=None, python=True):
```

pydra/utils/messenger.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+class AuditFlag(enum.Flag):
+    NONE = 0
+    PROV = enum.auto()  # 0x01
+    RESOURCE = enum.auto()  # 0x02
+    ALL = PROV | RESOURCE
+
+
+class RuntimeHooks(enum.IntEnum):
```
effigies: The only reason to use `IntEnum` over `Enum` is that `IntEnum` is comparable with `int`s outside the class. Is that the goal here?

See the note at the bottom of [enum.Flag](https://docs.python.org/3/library/enum.html#flag).

pydra/utils/messenger.py
```diff
@@ -0,0 +1,111 @@
...
+    fl = glob(str(message_path / '*.jsonld'))
+    data = []
+    for f in fl:
+        with open(f, 'rt') as fp:
+            data.append(json.load(fp))
+    if data:
+        records = getattr(pld.jsonld, ld_op)(
+            pld.jsonld.from_rdf(pld.jsonld.to_rdf(data,
+                                                  {})),
+            data[0])
```
effigies: At least the `pld.jsonld.from_rdf(...)` call should all be able to fit on one line. We may need to update something to make flake8 accept longer lines, but I believe the consensus was 99 character (+ `\n`) max?

pydra/utils/profiler.py
```diff
@@ -0,0 +1,319 @@
...
+        '''
+        if freq < 0.2:
+            raise RuntimeError(
+                'Frequency (%0.2fs) cannot be lower than 0.2s' % freq)
+        '''
+
+        if fname is None:
+            fname = 'proc-%d_time-%s_freq-%0.2f.log' % (pid, time(), freq)
+        if logdir is None:
+            logdir = Path(os.getcwd())
```
effigies:
```suggestion
            logdir = Path.cwd()
```

pydra/utils/profiler.py
```diff
@@ -0,0 +1,319 @@
...
+        self._fname = logdir / fname
+        self._logfile = open(self._fname, 'w')
+        self._freq = freq
+        self._python = python
+
+        # Leave process initialized and make first sample
+        self._process = psutil.Process(pid)
+        self._sample(cpu_interval=0.2)
```
effigies: Should this `cpu_interval` be hard-coded?

pydra/utils/profiler.py
```diff
@@ -0,0 +1,319 @@
...
+        # Iterate through child processes and get number of their threads
+        try:
+            children = self._process.children(recursive=True)
+        except psutil.NoSuchProcess:
+            children = []
+
+        for child in children:
+            try:
+                with child.oneshot():
+                    cpu += child.cpu_percent()
```
effigies: Should `interval=cpu_interval` be passed here, as well?

pydra/__about__.py
```diff
@@ -55,6 +55,8 @@
     'nipype>=1.1.3',
     'networkx>=%s' % NETWORKX_MIN_VERSION,
     'pytest>=%s' % PYTEST_MIN_VERSION,
+    'dataclasses; python_version < "3.7"',
+    'pyld'
```
mgxd: Some missing imports when running tests locally
```suggestion
    'pyld',
    'cloudpickle',
    'dask',
    'distributed',
```

effigies: I think `pyld` is only needed for `provenance`, and `dask` and `distributed` shouldn't be required at this stage, unless I've missed something. Perhaps move the non-core dependencies into a `'test'` extra for now, and we can think about splitting them out later?

mgxd: sounds good to me - I realized `dask` and `distributed` are already under `plugins`. we could move `pyld` to `dev`, but we should keep `cloudpickle` in our core

satra: at present pyld is only required to collect messages, so it can be moved to testing.

satra:
closed - see #30 instead.

PR #44, opened 2019-03-15
fix: submitting with futures

satra:


satra:
> so i Understand that returning in CF Worker.run_el self.pool.submit(interface, **kwargs) instead of self.pool.submit(interface, **kwargs).results() was enough to return future and to run truly parallel

i believe so, but there is not test for testing this still.

PR #48, opened 2019-04-07
[WIP] changes in wf, showing issue with cf

djarecka:
@satra - I'm having issue with pickling/deepcopy/cf in `the test_10a`, maybe you know what is going on. First I was having errors that result is None, and that the `load_result` is having issue, looks like  even if file exists `cp.loads` can give an error when the file is empty.

Now, my issue is that I don't know why I can't do `deepcopy`in `to_task` for the second node. This is not exactly the error when the tests stops, but when I started reading carefully the output I have `can't pickle _thread.RLock objects`

Do you have any thoughts?

### DIFF
--- a/pydra/engine/tests/test_node.py
+++ b/pydra/engine/tests/test_node.py
@@ -223,8 +223,9 @@ def test_task_nostate_1_cachedir(plugin, tmpdir):
 
 
 @pytest.mark.parametrize("plugin", Plugins)
-def test_task_nostate_1_cachedir_relativepath(plugin):
+def test_task_nostate_1_cachedir_relativepath(tmpdir, plugin):
     """Node with provided cache_dir as relative path"""
+    cwd = tmpdir.chdir()
     cache_dir = "test_task_nostate"
     nn = fun_addtwo(name="NA", a=3, cache_dir=cache_dir)
     assert np.allclose(nn.inputs.a, [3])
--- a/pydra/engine/tests/test_node.py
+++ b/pydra/engine/tests/test_node.py
@@ -562,7 +562,7 @@ def test_task_spl_comb_2(
     with Submitter(plugin=plugin) as sub:
         sub.run(nn)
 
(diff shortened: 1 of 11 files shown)


pydra/engine/tests/test_satraexample.py
```diff
@@ -227,16 +360,38 @@ def test_8(plugin):
     assert results[0][1].output.out == 6
 
 
+# @pytest.mark.parametrize("plugin", Plugins)
+# def test_10a(plugin):
+#     """Test workflow with node level splitters and combiners"""
+#     wf = Workflow(name="test10", input_spec=["x"])
+#     wf.add(add2(name="add2", x=wf.lzin.x).split("x"))
```
satra: just a note that for this situation, the final state of the workflow itself would need to be updated, such that subsequent nodes can do the right thing.

pydra/engine/tests/test_node.py
```diff
@@ -206,6 +206,95 @@ def test_task_nostate_1(plugin):
...
+    with Submitter(plugin=plugin) as sub:
+        sub.run(nn)
+
+    # checking the results
+    results = nn.result()
+    assert results.output.out == 5
+
+
+@pytest.mark.parametrize("plugin", Plugins)
+def test_task_nostate_1_cachedir_relativepath(plugin):
```
satra:
```suggestion
def test_task_nostate_1_cachedir_relativepath(plugin, tmpdir):
```

pydra/engine/tests/test_node.py
```diff
@@ -206,6 +206,95 @@ def test_task_nostate_1(plugin):
...
+    """Node with provided cache_dir as relative path"""
+    cache_dir = "test_task_nostate"
```
satra:
```suggestion
    cwd = tmpdir.chdir()
    cache_dir = "test_task_nostate"
```

PR #51, opened 2019-04-18
ENH: asynchronous execution

### LINKED ISSUE #39: Plugin - concurrent.futures
- [ ] Enable resource-available scheduling

mgxd:
This PR targets pydra execution, with significant speedup using asynchronous job submission execution. 

Closes #39.

Some preliminary work was done in a gist: https://gist.github.com/mgxd/9a00cba44ece7c92deccb8f88ae6cdea


pydra/engine/node.py
```diff
@@ -236,6 +240,12 @@ def cache_dir(self, location):
             self._cache_dir = mkdtemp()
             self._cache_dir = Path(self._cache_dir)
 
+    @property
+    def is_runnable(self):
```
satra: @mgxd - this is the kind of thing i would add a function to check and use the function in the workflow class. the function can take a set of dependencies (this will come in handy when you start dealing with splits).

try keeping functions out of classes when it's not necessary. also, please do test, because adding task pointers may make the class not pickle properly.

pydra/engine/core.py
```diff
@@ -563,25 +574,14 @@ def add(self, task):
...
-                task.run()
-            else:
-                from .submitter import Submitter
-
-                with Submitter(self.plugin) as sub:
-                    sub.run(task)
-                while not task.done:
-                    sleep(1)
+        # logic in submitter.run
+        # TODO: enable lockfile for workflow execution
```
satra: this should automatically happen, since `workflow.run()` or `workflow()` will be called somewhere. shouldn't have to worry about that.

pydra/engine/submitter.py
```diff
@@ -51,134 +78,36 @@ def run(self, runnable, cache_locations=None):
                 job.results_dict[None] = (None, checksum)
                 if cache_locations:
                     job.cache_locations = cache_locations
-                res = self.worker.run_el(job)
```
satra: there should be a check here that the job could be a workflow job.

pydra/engine/submitter.py
```diff
@@ -51,134 +78,36 @@ def run(self, runnable, cache_locations=None):
...
-
-    def _run_workflow_el(self, workflow, ind, collect_inp=False):
-        """running one internal workflow (if workflow has a splitter)"""
-        # TODO: can I simplify and remove collect inp? where should it be?
-        if collect_inp:
-            st_inputs, wf_inputs = workflow.get_input_el(ind)
+
+        # job = runnable
+
+        if is_workflow(job):  # expand out
```
satra: i think this should be indented into the above `else`

djarecka:
@mgxd - I saw that you're adding tests `test_concurrent_wf` and `test_wf_in_wf` to `test_submitter`. Are they specifically for the submitter or do you think they can be moved to `test_workflow`?

Are the current test are working fine for you? Please let me know if the tests need more explanations etc.

mgxd:
@djarecka - I was putting those workflows there since we are explicitly calling their execution through a `Submitter`, rather than just `Workflow()` or `Workflow().run()`. But this brings up something I'm not sure of - are planning on always running `Task`s through a context manager?

satra:
@mgxd - what's the status of this? is there an eta on when this will be ready for review?

djarecka:
@mgxd - have you ever tried testing the `asyncio` code with `pytest.asyncio`? heard a couple of talks about testing `asyncio`, and wondering if we should use it

mgxd:
@satra i'm hoping to finish up the state bits tomorrow

@djarecka i have not, but would be good to test some of the async functions in `Submitter`/`Worker`

djarecka:
@mgxd - will try to learn about it more today or tomorrow

djarecka:
@mgxd - do the previos tests from `test_node_task` and `test_workflow` work for you? are failing on my osx

djarecka:
`test_submitter` are also failing for me :(
are they just not ready yet, or you didn't commit some changes?

pydra/engine/core.py
```diff
@@ -346,6 +346,9 @@ def run(self, **kwargs):
                         {"@id": aid, "endedAtTime": now(), "errored": result.errored},
                         AuditFlag.PROV,
                     )
+            logger.debug("Completed %s.run", self)
+            if return_self:
```
djarecka: why do we need it?

pydra/engine/submitter.py
```diff
@@ -33,152 +29,108 @@ def __enter__(self):
...
+        Await worker for any finished tasks.
+        If any task is a workflow, return it.
+        """
+        done = await self.worker.fetch_finished()
+        for fut in done:
+            task = await fut
+            self.worker._remove_pending(fut)
+            logger.debug("Task: %s completed", task)
+            if is_workflow(task):
+                wf.name2obj.get(task.name).__dict__.update(task.__dict__)
```
mgxd: @djarecka because a workflow will expand and execute in a different thread, memory is not shared - this part ensures the object is updated in the preceding thread.

satra: @mgxd - since a workflow is a task, the only task that the preceding thread should care about is the workflow task, not its internals. as long as the workflow task returns a future, shouldn't that be sufficient?

pydra/engine/core.py
```diff
@@ -608,3 +617,14 @@ def is_task(obj):
 
 def is_workflow(obj):
     return isinstance(obj, Workflow)
+
+
+def is_runnable(graph, obj):
+    """Check if a task within a graph is runnable"""
+    if (not is_task(obj)) or not hasattr(graph, "predecessors"):
```
djarecka: I don't understand this if statement, and none of the tests are coming to this part. 

If you're trying to check that the obj is a `Workflow`, `not is_task(obj)` won't do the job

pydra/engine/submitter.py
djarecka: @mgxd - not sure if I asked about this part yet...

mgxd: ideally, but without this `update` call, the inner workflow's task results are not shared between threads

satra: why would they need to be shared between threads? since we can read results from disk anytime.

pydra/engine/core.py
mgxd: `obj` can be a `Task` or `Workflow`

djarecka: so when do you expect to execute the statement in the `if` part?

pydra/engine/submitter.py
mgxd: that sounds good to me - we would still have to load the results, check if the inputs have changed, and then assign to the downstream task

pydra/engine/core.py
mgxd: this can be removed, was essentially a catch for incorrect usage

pydra/engine/submitter.py
satra: doesn't each task do this on its own? the only place we were assigning things was at the beginning of a workflow task. when we called `inputs.retrieve_values.`

pydra/engine/core.py
djarecka: if this is the motivation, we should raise an exception, returning `False` is misleading

pydra/engine/submitter.py
mgxd: just to clarify: if this `update` is excluded, the inner workflow result will still have a `result()`. However, the tasks within the inner workflow will not, and execution will be stuck (inner workflow will not be marked as `done` since its tasks have no `result()`).

pydra/engine/workers.py
```diff
@@ -66,22 +75,42 @@ def close(self):
 
 
 class ConcurrentFuturesWorker(Worker):
-    def __init__(self, nr_proc=4):
-        self.nr_proc = nr_proc
+    def __init__(self, nr_proc=2, loop=None):
+        super(ConcurrentFuturesWorker, self).__init__()
+        self.nr_proc = nr_proc or mp.cpu_count()
```
djarecka: this change is not needed, `nr_proc` could be `None`. I was planning to change it to `None` in `__init__` as default.

mgxd: yep, this was my testing, i'll just set `nr_proc`'s default to `None`. (though we should provide some api for the user to alter this)

pydra/engine/submitter.py
```diff
@@ -33,152 +29,108 @@ def __enter__(self):
...
+                # instead, allow them to spawn a job
+                await self._run_task(task, wait_on_results=False)
+
+            # wait for at least one of the tasks to finish
+            wf = await self.fetch_pending(wf)
+
+        logger.debug("Finished workflow %s", wf)
+        return wf  # return the run workflow
+
+    def run(self, runnable, cache_locations=None):
```
djarecka: just as a reminder: we have to bring `cache_locations` back

mgxd:
Tests are passing with `-n 1`, but some fail as soon as multiple processes are introduced. This should be ready for a review though while I track this down

djarecka:
@mgxd - do you see this problem on your laptop? Not able to reproduce on osx...

mgxd:
yeah, though only spreading the tests across multiple processes. and even then, it isn't consistent - sometimes all pass. it does seem exclusive to the `test_wf_ndst_X` series though.

kaczmarj:
@djarecka - i can reproduce the error on my linux laptop. when testing using parallelized pytest, there is always at least one test that fails. when testing using sequential pytest, a test will sometimes fail, but often times all will pass.

i'm using python 3.7 on debian 10.

djarecka:
yes, I'm running it in some random nipype container and it fails even without `-n`

djarecka:
@mgxd - i was just planning to push the same, but it doesn't solve all the problems. now I have errors from `get_value`

djarecka:
it's even worse, when I was testing with a temporary loop to try to read the results twice, see [here](https://github.com/djarecka/pydra/commit/dc2283f92ea876ac49c5b0da9adb47dae20ba31c)
it can read wrong results... i.e. repeats the results from the first element `Result(output=Output(out=[[39, 42], [39, 42], [39, 42]]), runtime=None, errored=False)`

anibalsolon:
Hi @mgxd ! I was discussing w @djarecka here at the hackathon, and I was wondering why is there an asyncio event loop on the submitter and not in the ConcurrentFuturesWorker, since it (at least looks like) is a worker-dependant object. She brought about maybe a problem of having multiple event loops opened, so maybe a singleton-style would solve this?

djarecka:
i was hacking on this for a few hours and i added a few workarounds that are pretty ugly, but was just trying to figure what is not working. Unfortunately after dome "fixes" the code fails even less often, but it doesn't mean it's fine. I believe I still can get the exception from `Submitter`(l.63: "Nothing queued or todo - something went wrong").

You can take a look, there are multiple print statements, etc. Maybe the "fixes" I introduced will suggest something... 
https://github.com/djarecka/pydra/pull/4

mgxd:
Hi @anibalsolon - from what I understand, the goal of a `Submitter` is to initiate `Task` execution, by sending it to a `Worker` pool (CF, SLURM, etc) or in the case of a `Workflow`, handling the graph expansion as well. Since this expansion is happening asynchronously, we'd like to have the event loop handled by the `Submitter`. There shouldn't ever be multiple event loops spawned, is there an instance where this can happen? 

I'm going to push some changes that hopefully make loop behavior a little more straightforward

pydra/engine/core.py
```diff
@@ -567,31 +583,126 @@ def add(self, task):
...
-            task.inputs.retrieve_values(self)
-            if task.state and not hasattr(task.state, "states_ind"):
-                task.state.prepare_states(inputs=task.inputs)
-            if task.state and not hasattr(task.state, "inputs_ind"):
-                task.state.prepare_inputs()
-            if self.plugin is None:
-                task.run()
-            else:
-                from .submitter import Submitter
+    async def run(self, submitter=None, **kwargs):
```
satra: i think the base task should be turned async.

satra: i.e this function should not be in workflow.

mgxd: @satra this would affect a few things
1) all `Task._run_task()` methods would have to be made async as well
2) `Task`s would no longer be callable
3) Adding to the process pool would not become so straightforward (I don't believe `loop.run_in_executor()` works with coroutines)

djarecka:
@mgxd - I created a pr to your fork: https://github.com/mgxd/pydra/pull/5
hopefully that solves issues with workflows

pydra/engine/helpers.py
```diff
@@ -146,9 +146,7 @@ def execute(cmd):
         loop = asyncio.ProactorEventLoop()  # for subprocess' pipes on Windows
```
djarecka: are we actually support windows? anyone tested it?

satra: we haven't, but pydra should try to be cross-platform

satra:
if all tests are passing, let's merge this.

djarecka:
@satra - yesterday I added one small property to clarify and started working on the dictionary updates: https://github.com/nipype/pydra/pull/51/files#diff-de54d574bfa96684a83af82301489e88R704

I was able to eliminate for `runnable.state`, but don't understand why I need it for `runnable` without state, any idea: https://github.com/mgxd/pydra/pull/7

satra:
woohoo!!!

PR #52, opened 2019-04-20
cleaning, reviewing tests

djarecka:
- [x] reviewing  old test that had an old syntax

- [x] reviewing and merging tests for single tasks from test_node and test_satraexample (many were testing the same parts of codes, so removed)

- [x] adding tests for workflows inside a workflow

- [x] cleaning  State and tests

- [x] cleaning auxiliary and adding tests

- [x] cleaning Node


satra:
let's remove the nipype stuff out of pydra:
- [ ] remove `test_nipype1interfaces.py`
- [ ] clean up travis (remove fsl install, remove datalad download)
- [ ] remove nipype dependency in setup and elsewhere (old nipype Node)

djarecka:
@satra - is this is the [syntax](https://github.com/nipype/pydra/pull/52/commits/5b5053c4cffb3411690373dbfccca43fd3ac92a1#diff-61004b9f88d892cca950a805bfa61526R366) that you had in mind for wf as a node?

Because the `LazyField` only keeps the name of node (in this example `add2`) when set the first time (with `wfnd`), later `wf` is trying to find `add2` and not `wfnd.add2`. Let me know if you had something different in mind, otherwise I will change `LazyField` or/and `set_output`

pydra/engine/tests/test_workflow.py
```diff
@@ -0,0 +1,376 @@
...
+    """
+    wfnd = Workflow(name="wfnd", input_spec=["x"])
+    wfnd.add(add2(name="add2", x=wfnd.lzin.x))
+    wfnd.set_output([("out", wfnd.add2.lzout.out)])
+    wfnd.inputs.x = 2
+
+    wf = Workflow(name="wf", input_spec=["x"])
+    wf.add(wfnd)
+    # TODO: is this the syntax
+    wf.set_output([("out", wf.wfnd.add2.lzout.out)])
```
satra:
```suggestion
    wf.set_output([("out", wf.wfnd.lzout.out)])
```

satra: workflow nodes have inputs and outputs like anything else. the external workflow would not have access to any of the internal nodes outputs directly without calling `set_output` (i.e. line 360).

djarecka: yes, I was thinking about `wf.wfnd.lzout.out`, but `Workflow` object has no attribute `lzout`. So this should be changed

satra: since a workflow is a task, it should have all attributes of a task. hence it should work for a workflow object as well. lzout is done in getattr in the base class.

satra: it would only have it once `set_output` has been called.

satra: easy fix: workflow `getattr` should return this:

`return super().__getattr__(name)`

djarecka: but it doesn't work... I'm getting the error exactly when calling `set_output` (`wf.set_output([("out", wf.wfnd.lzout.out)]`). I will look at this later today

satra: after fixing workflow `getattr`?

djarecka: github site didn't refresh when I was writing, let me try the fix

djarecka: it works! Thank you! I will add more tests later today

satra:
those things should start with something other than `Test`, otherwise pytest thinks it needs to collect them.
  
```
pydra/engine/tests/test_specs.py:71
pydra/engine/tests/test_specs.py:71
  /home/travis/build/nipype/pydra/pydra/engine/tests/test_specs.py:71: PytestWarning: cannot collect test class 'TestNode' because it has a __init__ constructor
    class TestNode:
pydra/engine/tests/test_specs.py:99
pydra/engine/tests/test_specs.py:99
  /home/travis/build/nipype/pydra/pydra/engine/tests/test_specs.py:99: PytestWarning: cannot collect test class 'TestWorkflow' because it has a __init__ constructor
    class TestWorkflow:
-- Docs: https://docs.pytest.org/en/latest/warnings.html
```

pydra/engine/tests/test_workflow.py
```diff
@@ -0,0 +1,374 @@
...
+    with Submitter(plugin=plugin) as sub:
+        sub.run(wf)
+
+    # checking the results
+    while not wf.done:
+        sleep(1)
+    results = wf.result()
+
+    assert len(results.output.out) == 6
+    assert results.output.out == [39, 42, 52, 56, 65, 70]
```
satra: this looks like the workflow is auto-combining without a `combine`. is that intentional?

djarecka: no, they are not combined, just added additional tests with a combiner to show you the difference

djarecka: you can see all tests with names *_6/7/8/9, to see the difference in results with different combiners

djarecka:
@satra -  I'm sure the code requires more cleaning, but feel free to review and merge; I finished what I mostly wanted to do - cleaning tests related to task/wf (plus adding more and fixing more bugs..)
removed or commented some functions that are not used anymore

djarecka:
@satra - we should probably have more tests for `task.py` and `helpers.py`, but this could be done in a separate pr.
will also try to increase the coverage of `core.py`

satra:
thank you @djarecka - this is a great step forward. we should merge this and then move forward on specific issues. this will be helpful for me think about conditional nodes/workflows

PR #59, opened 2019-05-20
[wip] testing pypy

djarecka:
testing if pydra works for pypy3.6 (there is no pyp3.7 for now)

not sure why some tests are failing...


effigies:
Maybe we can try again later.

PR #60, opened 2019-05-22
adding a graph class (possible replacement for networkx)

djarecka:
@satra , @mgxd - i've created a simple class for graph that can be sorted. I'm sure it might be more efficient, but it seems to work (see tests). It also has `removing` that removes the node and sort the list again if needed (addresses what we were discussing yesterday, that even if we sort the graph with `nx` at the beginning, we might have to jump over some nodes anyway)

let me know if I should finish this and replace the `nx`, would have to add methods  like `add_node`, `add_edge`, etc. and probably more tests

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -1,11 +1,8 @@
 """Basic compute graph elements"""
 import abc
-from collections import OrderedDict
 import dataclasses as dc
-import itertools
 import json
 import logging
-import networkx as nx
 import os
 import pdb
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


djarecka:
@satra  - I've changed the class, so can work for `Workflow`, tell me what you think

djarecka:
@satra - can do it, but i'm curious how do you want to use it in pydra

satra:
it's not urgent, but would be used to optimize which nodes to run. for optimization we will implement a graph scheduler eventually.

djarecka:
@satra still don't see it, but will implement a simple version

djarecka:
@satra , @mgxd - let me know if you want me to merge it to master before the asyncio PR. Some of the feature might be useful in `Submitter` (e.g. `remove_node` with sorting), but I'm ok with merging this later

djarecka:
open a new PR that has already asyncio changes (#66)

PR #61, opened 2019-05-24
Hash

djarecka:
- adding nodes hashes to the wf hash (including sorted graph in the wf input)
- using `cahce_locations` for wf (it was not set in the init)

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -484,6 +484,7 @@ class Workflow(TaskBase):
         messengers=None,
         messenger_args=None,
         cache_dir=None,
+        cache_locations=None,
         **kwargs,
     ):
         if input_spec:
@@ -506,6 +507,7 @@ class Workflow(TaskBase):
             name=name,
             inputs=kwargs,
             cache_dir=cache_dir,
+            cache_locations=cache_locations,
             audit_flags=audit_flags,
             messengers=messengers,
             messenger_args=messenger_args,
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -96,7 +96,9 @@ class TaskBase:
             }
         )
         self.input_names = [
-            field.name for field in dc.fields(klass) if field.name not in ["_func"]
+            field.name
+            for field in dc.fields(klass)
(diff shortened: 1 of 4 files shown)


pydra/engine/core.py
```diff
@@ -563,6 +570,7 @@ def add(self, task):
             else:
                 task.state = state.State(task.name, other_states=other_states)
         self.node_names.append(task.name)
+        self.graph_sorted
```
satra: perhaps the semantics of this can be improved. this looks weird in python. i would turn this into a function with a better name.

satra:
```suggestion
        self.inputs._graph = self.graph_sorted
```

djarecka:
@satra  - which property you would like to rename to `update_input_graph()`? you mean `graph_sorted`? the new part of `hash`?

pydra/engine/core.py
```diff
@@ -503,13 +507,15 @@ def __init__(
...
+            cache_locations=cache_locations,
             audit_flags=audit_flags,
             messengers=messengers,
             messenger_args=messenger_args,
         )
 
         self.graph = nx.DiGraph()
-
+        self._graph_sorted = []
+        self.inputs._graph = self._graph_sorted
```
satra: just remove the above two lines.

pydra/engine/core.py
```diff
@@ -531,7 +537,8 @@ def nodes(self):
 
     @property
     def graph_sorted(self):
-        return list(nx.topological_sort(self.graph))
+        self._graph_sorted[:] = list(nx.topological_sort(self.graph))
+        return self._graph_sorted
```
satra: keep this the original property

pydra/engine/core.py
satra: make this change

satra:
@djarecka - i commented at three places.

pydra/engine/core.py
djarecka: I had to do this way, because didn't work properly without the additional assignment, had to be sure that it doesn't create a new object. Right now I can't recall details, can check later

PR #64, opened 2019-06-20
enh: SLURM worker

mgxd:
Builds off #51's implementation of the `Submitter()`.
A few things that still need fixing:
- [x] `cache_dir` is set to /tmp by default, which is not guaranteed to be shared across compute nodes on HPCs - provide a warning in this case?
- [x] `cache_dir` not propagating to `Tasks` within `Workflow`
- [x] submission scripts (pyscript, batchscript) are currently disappearing?
- [x] better `sbatch_arg` support
- [x] ensure `slurm-%j.out` files are output to `cache_dir`


satra:
@mgxd - please merge master

pydra/engine/helpers.py
```diff
@@ -57,9 +57,28 @@ def load_result(checksum, cache_locations):
...
+    task_path : Path
+        Write directory
+    result : Result
+        Result to pickle and write
+    task : Task
+        Task to pickle and write
+    """
+    if task is None and result is None:
+        raise ValueError("Nothing to be saved")
+    task_path.mkdir(parents=True, exist_ok=True)
```
satra: i would save tasks at a different location than the tasks output directory.

mgxd: i agree, i'd like to lump all slurm files (`task.pklz', `pyscript`, `batchscript`) in one location.

ci/none.sh
```diff
@@ -0,0 +1,58 @@
...
+    elif [ "$CHECK_TYPE" = "style" ]; then
+        pip install black
+    fi
+}
+
+function travis_script {
+    if [ "$CHECK_TYPE" = "test" ]; then
+       pytest -vs -n auto --cov pydra --cov-config .coveragerc --cov-report xml:cov.xml --doctest-modules pydra
+    elif [ "$CHECK_TYPE" = "style" ]; then
+        black pydra setup.py
```
effigies: I screwed this up before. This should be:

```suggestion
        black --check pydra setup.py
```

satra:
@mgxd - since you are running slurm in a container, you have to save or send the code-coverage report. look at how nipype does this.

.travis.yml
```diff
@@ -29,66 +33,29 @@ env:
 # Useful for testing un-released upstream fixes
 matrix:
   include:
-  - python: 3.8-dev
+  - python: "3.7"
```
effigies: Let's either do quoted or not. We have some that are and some that aren't.

mgxd:
all green + ready for review

satra:
@mgxd - while i think the slurm plugin will require quite a few more bells and whistles, this is a good start for now.

djarecka:
we decreased the test coverage by 17%... could we please bring it back in the nearest future ;)

djarecka:
looks like [my pr](https://github.com/nipype/pydra/pull/106) brings back the coverage, so it looks like codecov got lost here?

PR #68, opened 2019-07-09
Graph clas

djarecka:
adding a simple graph class and removing networkx


pydra/engine/core.py
```diff
@@ -504,13 +504,8 @@ def add(self, task):
             if isinstance(val, LazyField):
                 # adding an edge to the graph if task id expecting output from a different task
                 if val.name != self.name:
+                    self.graph.add_edges((getattr(self, val.name), task))
```
satra: this doesn't appear to be storing the specific edge info. the output and input names. where is that being stored?

pydra/engine/core.py
```diff
@@ -457,7 +457,7 @@ def __init__(
             messenger_args=messenger_args,
         )
 
-        self.graph = nx.DiGraph()
+        self.graph = Graph()
```
satra: can we make the class networkx compatible?

djarecka: and by this you mean?

satra: `DiGraph()`

pydra/engine/core.py
```diff
@@ -608,8 +603,8 @@ def is_workflow(obj):
 
 def is_runnable(graph, obj):
     """Check if a task within a graph is runnable"""
-    if graph.predecessors(obj):
-        for pred in graph.predecessors(obj):
+    if graph.connections_pred[obj.name]:
```
satra: use `predecessors`

djarecka: but for sorting graph etc. we don't need  this info. right?

pydra/engine/core.py
djarecka: but for sorting graph etc. we don't need this info. right?

pydra/engine/core.py
djarecka: i've changed some names, but not all, since I sometimes combine two methods from nx into one like `add_node` and `add_nodes_from` into `add_nodes` that works for one element or for a list

pydra/engine/core.py
satra: since the graph is the workflow, it would be useful to store that information here rather than somewhere else. the link to the nodes feeding information to a node is right here. if not this information is also being stored in the node. seems redundant.

djarecka: ok, can store it with the graph and see which wf attributes could be removed

djarecka: we agreed that it's not needed. everything is stored in LazyField

PR #70, opened 2019-07-12
changing input_spec to allow tasks without input

djarecka:
tasks without inputs didn't work

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -54,11 +54,6 @@ class TaskBase:
     _cache_dir = None  # Working directory in which to operate
     _references = None  # List of references for a task
 
-    # dj: do we need it??
-    input_spec = BaseSpec
-    output_spec = BaseSpec
-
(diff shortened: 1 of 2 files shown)


satra:
i'm inclined not to support this. can you think of a situation in a dataflow for a task not to have an input?

djarecka:
@satra - as you want. I noticed this when I worked on notebooks and I thought that we should allow it, but no strong feelings...

satra:
leave this as an orphaned PR, but i would prefer not.

djarecka:
@satra - and what if we specify input for task, and not for wf? 

```
wf1 = pydra.Workflow(name="wf1")
wf1.add(add_two(name="add_two", x=1))
```

satra:
> and what if we specify input for task, and not for wf?

a `Workflow` is a `Task` hence has to have the same consideration. i.e. there has to be at least one task that receives a workflow input

djarecka:
ok

PR #76, opened 2019-07-16
[wip, fix] small changes and fixes

djarecka:
various small changes: 
- adding imports to __init__ files; 
- adding exception for no `input_spec` (Assuming that #70 won't be accepted); 
- removing loop.close when Jupyter in os.environ (I'm just adding this variable in the tutorials, other suggestions to run pydra in jupyter welcome); 
- adding `output_names` to Task, so it doesn't have to be "out"
- adding tests 

tests that xfail (will be fixing later):
- tasks wit splitter can't be run using __call__
- if `output_names` is not specified we can't save as an output multiple variables


pydra/engine/submitter.py
```diff
@@ -149,7 +150,9 @@ def __exit__(self, type, value, traceback):
             return futures
 
     def close(self):
-        self.loop.close()
+        # jupyter is also using cf, so can't close the loop
+        if not "Jupyter" in os.environ:
```
satra: this seems something that Jupyter should handle, not us. i don't want pydra to be dependent on whether it's used in a specific environment.

pydra/engine/task.py
```diff
@@ -60,6 +60,7 @@ def __init__(
         self,
         func: ty.Callable,
         output_spec: ty.Optional[BaseSpec] = None,
+        output_names=None,
```
satra: no `output_names` - if needed to be specified they should use output_spec

satra: see this example: https://github.com/nipype/pydra/blob/master/pydra/engine/tests/test_task.py#L27

satra: perhaps an example we should work towards is how to annotate an existing function via the `to_task` decorator

pydra/engine/submitter.py
djarecka: The short answer is - it also surprised me, but Jupyter doesn't handle... Note, that it's not only this, I still had to import `nest_asyncio` to the notebook!

I agree that we don't want to change pydra, the if statement seemed to me the only "acceptable" option, but I will search more.

djarecka: yesterday I was thinking about forgetting about jupyter, but I'm guessing it's not an option that will be approved ;-)

pydra/engine/task.py
djarecka: I was thinking that `output_names` could be the easier option

satra: i agree it would be easier, but we want them to do it through proper annotation. since we will want people to provide type hints as well eventually.

so we should see if we can add `__annotations__` to the function using the decorator and adding arguments to the decorator.

pydra/engine/submitter.py
effigies: Do we have config?

```suggestion
        if not config.getboolean("execution", "nested_asyncio"):
```

Then, Jupyter notebooks could use a modified config file, or users could add at import: 

```Python
import pydra
pydra.config.set("execution", "nested_asyncio", True)
```

djarecka: but the problem is that we also want pydra to change the behavior, i.e. do not close the loop. At least that was the solution we found yesterday

satra: one possibility is for pydra to check if there is a loop and reuse it rather than creating a new one? and, in such an event, not close it.

djarecka: we were thinking about it yesterday, but we were afraid that this might happen also in pydra Workflow itself, and has nothing to do with jupyter.

Just to be clear all tests pass without closing the loop at all, it's "just" a best practice, and I'm not sure when this might be a problem.

@mgxd can comment as well

pydra/engine/task.py
djarecka: ok, will work on it

pydra/engine/submitter.py
satra: a) a loop should be opened only when the program is not already in a loop
b) a loop should be closed only if the loop has been opened by the program

mgxd: we can easily detect if the available loop is currently running, and avoid closing it, but we'll still need to import / activate `nest_asyncio` to allow running tasks through jupyter

djarecka: @satra - looks like this is not issue create by jupyter, but newer version of tornado, and it looks like this is still not fixed. It works fine (at least for my simple example, will be testing more) if older version of tornado is used `tornado==4.5.3`

djarecka: didn't read all posts, but you can find more here: https://github.com/jupyter/notebook/issues/3397

pydra/engine/core.py
```diff
@@ -441,6 +437,8 @@ def __init__(
                     + [("_graph", ty.Any)],
                     bases=(BaseSpec,),
                 )
+        else:
+            raise Exception("Workflow has to have input_spec")
```
effigies: I think you can just do this with types:

```Diff
 class Workflow(TaskBase):
     def __init__(
         self,
         name,
-        input_spec: ty.Union[ty.List[ty.Text], BaseSpec, None] = None,
+        *,
+        input_spec: ty.Union[ty.List[ty.Text], BaseSpec],
         output_spec: ty.Optional[BaseSpec] = None,
         audit_flags: AuditFlag = AuditFlag.NONE,
```

Keyword-only arguments don't have to have defaults.

satra:
should this be closed?

PR #78, opened 2019-07-16
Fix: loop management

mgxd:
Only close event loops that we create.

This should make it possible to submit Pydra tasks with a Jupyter notebook, after patching nested event loops with:

```python
import nest_asyncio
nest_asyncio.apply()
```

### DIFF
--- a/pydra/engine/submitter.py
+++ b/pydra/engine/submitter.py
@@ -13,7 +13,7 @@ class Submitter:
     # TODO: runnable in init or run
     def __init__(self, plugin):
         self.loop = get_open_loop()
-        self._running_loop = self.loop.is_running()
+        self._own_loop = not self.loop.is_running()
         self.plugin = plugin
         if self.plugin == "serial":
             self.worker = SerialWorker()
--- a/pydra/engine/submitter.py
+++ b/pydra/engine/submitter.py
@@ -13,6 +13,7 @@ class Submitter:
     # TODO: runnable in init or run
     def __init__(self, plugin):
         self.loop = get_open_loop()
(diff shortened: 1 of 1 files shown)


pydra/engine/submitter.py
```diff
@@ -13,6 +13,7 @@ class Submitter:
     # TODO: runnable in init or run
     def __init__(self, plugin):
         self.loop = get_open_loop()
+        self._running_loop = self.loop.is_running()
```
effigies: Perhaps a more intuitive way of thinking about it:

```suggestion
        self._own_loop = not self.loop.is_running()
```

pydra/engine/submitter.py
```diff
@@ -147,7 +148,9 @@ def __exit__(self, type, value, traceback):
             return futures
 
     def close(self):
-        self.loop.close()
+        # do not close previously running loop
+        if not self._running_loop:
```
effigies:
```suggestion
        if self._own_loop:
```

djarecka:
we actually don't need it. Looks like if you install `tornado==4.5.3` jupyter is able to manage the loops. I was checking this on my OSX

mgxd:
@djarecka I think we should still play nice with existing loops, and ensure we aren't closing them

pydra/engine/submitter.py
effigies: I don't think this got applied.

djarecka:
@mgxd - oh, sure, do whatever is better, I didn't read the pr carefully.

I'm just saying that this is not to make it work in Jupyter, and we don't need `nest_asyncio`

PR #81, opened 2019-07-17
MAINT: Move to setup.cfg + pyproject.toml

### LINKED ISSUE #80: problems with requirements
I'm having issue with installing the package and the way we specify the requirements. I wasn't paying attention to this, I know that we used to have `requirements.txt`, but it has disappeared and now everything is in `__about__`.  It was fine, but I'm doing something wrong in #76, where I've started  exposing some objects to the user in the `__init__` files. Suggestions will be appreciated!

effigies:
Setuptools 30.3.0+ supports `setup.cfg`-based setup, which obviates the need for an `__about__.py` file, to the extent that most things can be set statically.

The main thing that needed to be dynamically generated is `DOWNLOAD_URL`, but it's not really clear what that provides. My reading of some PyPA threads is that it used to affect installation tools' behavior by providing a download location, but `pip` now ignores it and PyPI just hosts it as metadata. So I removed it. Please make noise if this disturbs you.

Getting pretty good at doing this right on the first try, but I should still add Travis tests to make sure multiple install methods work as expected.

Closes #80.


effigies:
@djarecka Why are we testing workflows 5 times?

mgxd:
@effigies that was added due to some flaky test results emerging in (#51).

@djarecka do you still want to keep that? It looks like we have solved that problem now

djarecka:
it's temporally, the reason was simple - at some point we were having errors that did not show up on osx and on linux os was raised not every single time. I could have all tests passing without fixing the problem. 

we will remove it at some point, but right now I feel saver with it. We could also change the number from 6 to 3 or something.

djarecka:
@mgxd - yes, I still want to keep it, maybe 3 not 6. We could remove it once we have stable submitter and workers.

effigies:
I'm going to be adding multiple runs anyway. Will that be sufficient, or is a loop specifically desired?

djarecka:
@effigies - no, there is nothing special about the loop, multiply runs will be fine

setup.cfg
```diff
@@ -1,5 +1,62 @@
...
+    License :: OSI Approved :: Apache Software License
+    Operating System :: MacOS :: MacOS X
+    Operating System :: POSIX :: Linux
+    Programming Language :: Python :: 3.7
+    Topic :: Scientific/Engineering
+
+[options]
+python_requires = >= 3.7
+install_requires =
+    pytest >= 4.4.0
```
effigies: @satra @djarecka It's not clear that this is actually a dependency, as pytest isn't imported outside of `test_*.py` files. Should I move it to the extras_require?

setup.cfg
```diff
@@ -1,5 +1,62 @@
...
+maintainer = nipype developers
+maintainer_email = neuroimaging@python.org
+description = Pydra dataflow engine
+long_description = file:long_description.rst
+long_description_content_type = text/x-rst; charset=UTF-8
+license = Apache License, 2.0
+provides =
+    pydra
+classifiers =
+    Development Status :: 2 - Pre-Alpha
```
effigies: Is the next release going to be alpha? It looks like we're getting there...

```suggestion
    Development Status :: 3 - Alpha
```

setup.cfg
```diff
@@ -1,5 +1,62 @@
...
+    Environment :: Console
+    Intended Audience :: Science/Research
+    License :: OSI Approved :: Apache Software License
+    Operating System :: MacOS :: MacOS X
```
effigies: We aren't testing on OSX. Probably should if we want to make this claim.

effigies:
> @effigies - no, there is nothing special about the loop, multiply runs will be fine

Ah, you commented while I was commenting. Thanks.

setup.cfg
djarecka: i think we can move it to extra_require

setup.cfg
djarecka: we probably should add to travis. Of course, it is tested on my osx, but nothing more

effigies: Okay. That feels like a lot to put in this already kind of big PR. I'll open an issue.

satra: we should i think also add a windows build via appveyor.

effigies: Given how obnoxiously slow Appveyor is, I would propose giving Azure a shot.

satra: @effigies - the addition of testing on windows. @djarecka - we could ask hoda to look into this as part of getting her feet wet with all the tools.

effigies: Created #82 and #83.

djarecka: you mean to install and test on her os, or adding windows to CI?

satra: both can be separately added based on the issues you created.

satra:
i'm going to merge this.

PR #85, opened 2019-07-18
Fix/checksum

djarecka:
changes in `TaskBase.checksum` and `BaseSpec.hash`. 
My changes introduced in [this commit](https://github.com/nipype/pydra/commit/86467b7fa9335ec60bee051227188ffa73f798a6) that were fixing problem with no existing output_dir, didn't work for situations from `test_annotated_func` (test was xfailed..).

Main changes:
- all checksum can be calculated before running (also for state) 
- hash property is changed to a method that takes dictionary with state values as an optional argument (so can replace values for inputs that are part of the splitter)
- to calculate has of `_graph`, `nd.inputs.hash()` is calculated, so `TaskBase.checksum` diesn't have to deal with lazy inputs

### DIFF
--- a/pydra/engine/specs.py
+++ b/pydra/engine/specs.py
@@ -18,18 +18,19 @@ class SpecInfo:
 class BaseSpec:
     """The base dataclass specs for all inputs and outputs"""
 
-    @property
-    def hash(self):
+    def hash(self, replace=None):
         """Compute a basic hash for any given set of fields"""
         inp_dict = {
             field.name: getattr(self, field.name)
             for field in dc.fields(self)
             if field.name not in ["_graph"]
         }
+        if replace:
+            inp_dict.update(replace)
         inp_hash = self._hash(inp_dict)
         if hasattr(self, "_graph"):
-
-            graph_hash = [nd.checksum for nd in self._graph]
+            # calculating inputs.hash for every node inside the graph
+            graph_hash = [nd.inputs.hash() for nd in self._graph]
             return self._hash((inp_hash, graph_hash))
         else:
             return inp_hash
@@ -44,7 +45,6 @@ class BaseSpec:
             if isinstance(value, LazyField):
                 value = value.get_value(wf, state_index=state_index)
                 temp_values[field.name] = value
-        # dj: changes the outputd_dir (node is updating input)
         for field, value in temp_values.items():
             setattr(self, field, value)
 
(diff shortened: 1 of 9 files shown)


pydra/engine/specs.py
```diff
@@ -18,18 +18,19 @@ class SpecInfo:
...
             if field.name not in ["_graph"]
         }
+        if replace:
+            inp_dict.update(replace)
         inp_hash = self._hash(inp_dict)
         if hasattr(self, "_graph"):
-
-            graph_hash = [nd.checksum for nd in self._graph]
+            # calculating inputs.hash for every node inside the graph
+            graph_hash = [nd.inputs.hash() for nd in self._graph]
```
satra: if an internal node has splits, how would you know whether splits have changed or not.

pydra/engine/tests/test_task.py
```diff
@@ -171,7 +162,7 @@ def test_shell_cmd(tmpdir):
     assert res.output.stdout == " ".join(cmd[1:]) + "\n"
 
     # separate command into exec + args
-    shelly = ShellCommandTask(executable=cmd[0], args=cmd[1:])
+    shelly = ShellCommandTask(name="test", executable=cmd[0], args=cmd[1:])
```
satra: should not be necessary.

djarecka:
@satra - I added the changes after discussion. I've added some tests, including one for workflows that have the same inputs and nodes, but different spliter in one of the node.

Unfortunately, one test is still failing, looks like `inputs["_func"]` gives different hash for my osx and linux. will check it tomorrow, but let me know if you have any idea, looks like I had to introduce the error in this PR...

satra:
@djarecka - since we are using cloudpickle it's quite possible that the dependencies are different between your setup and travis. may be worthwhile to test in a container with the same environment. but this is an issue we will need to address better. (perhaps in another PR).

djarecka:
@satra - yes, `cp.dumps(func)` returns different values. And I didn't introduce it in this PR, I've just added this assert in the PR... I moved it to `test_task` that is xfailing and opened #87

pydra/engine/tests/test_task.py
djarecka: was needed because `ShellCommandTask` didn't have default value for `name`. Added None

satra:
i'm merging this together with a few things to be sorted later. also i would like to see if we can avoid the `deepcopy(self.graph_sorted)`

djarecka:
@satra - see #92

PR #86, opened 2019-07-19
maint: license + pr template

mgxd:
Makes licensing information easily available and adds PR template with acknowledgment.


satra:
@mgxd - i don't know why the event loop error is happening again.

PR #88, opened 2019-07-19
changing PR template

djarecka:
I've suggested some changes to the template, mostly adding checklist. 

Also, I don't see the template to have any effect for new PRs, it shouldn't be called `PULL_REQUEST_TEMPLATE.md`??

**edit**
renamed the file

Some of my instructions should be in the contributor guidelines, but added here temporally to the template (see #89)


mgxd:
I wanted to make the behavior similar to the ISSUE_TEMPLATES, allowing for multiple templates, but it looks like currently the only way to access different templates is to include `?template=basic.md` in the URL :disappointed: 

Feel free to remove the `PULL_REQUEST_TEMPLATES` directory and rename the file to `.github/PULL_REQUEST_TEMPLATE.md`

PR #92, opened 2019-07-19
removing graph from inputs, adding checksums

djarecka:
## Checklist

- [ x] All tests passing
- [ x] I have added tests to cover my changes
- [ x] My code follows the code style of this project 

## Summary

saving checksums of nodes from the graph instead of copy of entire graph in `wf.inputs` (we only need it for calculating checksum anyway)

## Acknowledgment
- [x ] I acknowledge that this contribution will be available under the Apache 2 license.

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -94,7 +94,7 @@ class TaskBase:
         self.input_names = [
             field.name
             for field in dc.fields(klass)
-            if field.name not in ["_func", "_graph"]
+            if field.name not in ["_func", "_graph_checksums"]
         ]
         self.state = None
         self._output = {}
@@ -446,7 +446,7 @@ class Workflow(TaskBase):
                 self.input_spec = SpecInfo(
                     name="Inputs",
                     fields=[(name, ty.Any) for name in input_spec]
-                    + [("_graph", ty.Any)],
+                    + [("_graph_checksums", ty.Any)],
                     bases=(BaseSpec,),
                 )
         if output_spec is None:
@@ -508,7 +508,7 @@ class Workflow(TaskBase):
         self._last_added = task
         self.create_connections(task)
         logger.debug(f"Added {task}")
-        self.inputs._graph = deepcopy(self.graph_sorted)
+        self.inputs._graph_checksums = [nd.checksum for nd in self.graph_sorted]
         return self
 
     def create_connections(self, task):
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -506,7 +506,6 @@ class Workflow(TaskBase):
         self.graph.add_nodes(task)
         self.name2obj[task.name] = task
         self._last_added = task
-        self.create_connections(task)
         logger.debug(f"Added {task}")
         return self
 
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -508,7 +508,6 @@ class Workflow(TaskBase):
         self._last_added = task
         self.create_connections(task)
         logger.debug(f"Added {task}")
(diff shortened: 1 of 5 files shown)


djarecka:
this assumes (as in master) that we're not changing nodes after adding it to wf

pydra/engine/core.py
```diff
@@ -508,7 +508,7 @@ def add(self, task):
         self._last_added = task
         self.create_connections(task)
         logger.debug(f"Added {task}")
-        self.inputs._graph = deepcopy(self.graph_sorted)
+        self.inputs._graph_checksums = [nd.checksum for nd in self.graph_sorted]
```
satra: this would compute graph and node checksum multiple times for each add.

pydra/engine/specs.py
```diff
@@ -25,13 +25,11 @@ def hash(self):
...
-            if field.name not in ["_graph"]
+            if field.name not in ["_graph_checksums"]
         }
         inp_hash = hash_function(inp_dict)
-        if hasattr(self, "_graph"):
-            # calculating checksum for every node inside the graph
-            graph_hash = [nd.checksum for nd in self._graph]
-            return hash_function((inp_hash, graph_hash))
+        if hasattr(self, "_graph_checksums"):
+            return hash_function((inp_hash, self._graph_checksums))
```
satra: i would compute the graph checksum at this step.

pydra/engine/core.py
djarecka: the graph has to be recomputed (it's sorted)

pydra/engine/specs.py
djarecka: but that was in the current master...

satra: in the current master `graph_hash = [nd.checksum for nd in self._graph]` this computes the hash for every node at the point of checksum computation, not in the `add` function.

pydra/engine/core.py
satra: all i am saying is compute the checksum only when it is needed.

pydra/engine/specs.py
djarecka: in that case I don't know how to remove deepcopy. Nodes are changing, so if we don't create copy and we calculate whenever we want, than the checksum will be changing during execution, and we don't want it.

satra:
i don't think we want to deepcopy in the `add` function, but when the graph checksum is being computed.

satra:
is this now ready?

djarecka:
@satra - it might be good if you review the last changes. If you like it, you can merge

pydra/engine/tests/test_workflow.py
```diff
@@ -1438,6 +1632,68 @@ def test_wf_ndstate_cachelocations(plugin, tmpdir):
...
+    with Submitter(plugin=plugin) as sub:
+        sub(wf2)
+    t2 = time.time() - t0
+
+    results2 = wf2.result()
+    assert results2.output.out == [8, 82]
+
+    # checking execution time
+    assert t1 > 3
+    assert t2 < 0.1
```
satra: i'm going to merge this, but the way i think we should evaluate whether the correct tasks are rerun is using messaging. (see #9)

djarecka: are you sure that you pointed me to the right pr?

satra: oops - updated the link

djarecka: yes, we should add it, but just fyi, the time is not the only check in the tests, I'm also checking if the outputdir is created.

PR #94, opened 2019-07-22
Output spec

### LINKED ISSUE #75: functions with multiple values in return statement
If default `out` for output is used (without defining `output_names` - will be in pr) and a function returns two values, only one value is saved. See:

```
@pydra.to_task
def add_vector(x1, y1, x2, y2):
    return (x1 + x2, y1 + y2)

task5 = add_vector(name="add_vect", 
                   x1=[10, 20], y1=[1, 2], x2=[10, 20, 30], y2=[10, 20, 30])
task5.split(splitter=[("x1", "y1"), ("x2", "y2")])
task5(plugin="cf")
task5.result()
```

djarecka:
## Types of changes

closes  #75

## Summary
- some changes to FunctionTask.output_spec and additional tests for using `__annotations__` to specify multiple outputs etc.
**edited**
issue #75 is fixed and multiple values could be return from functions with and without annotation

- also, I was checking one of the test from test_task that is xfailing, it's about the audit. I split the test to two tests, one is passing, but `test_audit_all` is failing with [the pickling error](https://travis-ci.org/djarecka/pydra/jobs/562287591). @satra - do you know how to fix it? 

## Checklist

- [x] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -132,9 +132,16 @@ class FunctionTask(TaskBase):
         del inputs["_func"]
         self.output_ = None
         output = cp.loads(self.inputs._func)(**inputs)
-        if not isinstance(output, tuple):
-            output = (output,)
-        self.output_ = list(output)
+        if len(self.output_spec.fields) > 1:
+            if len(self.output_spec.fields) == len(output):
+                self.output_ = list(output)
+            else:
+                raise Exception(
+                    f"expected {len(self.output_spec.fields)} elements, "
+                    f"but {len(output)} were returned"
+                )
+        else:  # if only one element in the fields, everything should be returned together
+            self.output_ = output
 
     def _list_outputs(self):
         return self.output_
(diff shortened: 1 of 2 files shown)


djarecka:
@satra - the failing test was in my older pr: https://github.com/nipype/pydra/pull/76/files#diff-871957728874435d337bd031b6bcb53cR266

I didn't add it at the end, because I understand that the annotation should be the default options.

I don't see any option to do it using `to_task`, but I guess we could add an option to `add_task`  and `FunctionTask.__init__`

satra:
@djarecka - that test should still work. the most default option is no annotation.

djarecka:
@satra  - ok, I found what was wrong and fixed it

djarecka:
@satra - regarding you question about annotating an existing function - see [this PR](https://github.com/djarecka/pydra/pull/7). If this is what you had in mind I could finish it

djarecka:
@satra - you didn't comment on the `Audit.ALL` issue, created #98 to not forget

PR #95, opened 2019-07-23
[wip] to_task decorator with arguments

djarecka:
## Types of changes

## Summary

testing `to_task` decorator with arguments for providing output spec (instead of annotating the function).

we should think if you want to have a separate decorator for it, right now `@to_task` would have to be changed to `@to_task()` in all our tests 

## Checklist

- [ ] All tests passing

## Acknowledgment

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,7 +78,7 @@ before_script:
 script:
   - |
     if [ "$CHECK_TYPE" = "test" ]; then
-        py.test -vs -n auto --cov pydra/engine/tests/test_task.py --cov-config .coveragerc --cov-report xml:cov.xml --doctest-modules pydra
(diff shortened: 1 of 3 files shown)


effigies:
Just as a thought, we could follow the `pytest.mark` model and separate out into some discrete, useful decorators. So the following would be equivalent:

```Python
@pydra.mark.task
@pydra.mark.annotate(a=int, b=float, return={'ret1': int, 'ret2': float})
def f(a, b):
    return a, b
```

```Python
@pydra.mark.task
def f(a: int, b: str) -> {'ret1': int, 'ret2': str}:
    return a, b

djarecka:
@effigies - I like the idea!

djarecka:
new pr #101

PR #100, opened 2019-07-25
Fix/cache propagation

mgxd:
## Types of changes

## Summary

If a `Workflow`'s `cache_dir` is not `None`, it should be propagated to downstream `Task`s. This behavior was discovered in #64. 

- Propagates a Workflow's `cache_dir` to tasks within, unless the task explicitly sets the `allow_cache_override` to `False`

## Checklist

- [x] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -205,9 +205,11 @@ class TaskBase:
     @cache_dir.setter
     def cache_dir(self, location):
         if location is not None:
+            self._custom_cache_dir = True
             self._cache_dir = Path(location).resolve()
             self._cache_dir.mkdir(parents=False, exist_ok=True)
         else:
+            self._custom_cache_dir = False
             self._cache_dir = mkdtemp()
             self._cache_dir = Path(self._cache_dir)
 
@@ -502,6 +504,8 @@ class Workflow(TaskBase):
         """adding a task to the workflow"""
         if not is_task(task):
             raise ValueError("Unknown workflow element: {!r}".format(task))
+        if self._custom_cache_dir:
+            task.cache_dir = self.cache_dir
         self.graph.add_nodes(task)
         self.name2obj[task.name] = task
         self._last_added = task
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -206,11 +206,9 @@ class TaskBase:
     @cache_dir.setter
     def cache_dir(self, location):
         if location is not None:
-            self._custom_cache_dir = True
             self._cache_dir = Path(location).resolve()
             self._cache_dir.mkdir(parents=False, exist_ok=True)
         else:
-            self._custom_cache_dir = False
             self._cache_dir = mkdtemp()
             self._cache_dir = Path(self._cache_dir)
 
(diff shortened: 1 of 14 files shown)


satra:
@mgxd - a few notes.

1. the `cache_dir` should be set at runtime not add time. since someone can change the workflow cache_dir at any time.
2. instead of this approach, add a TaskBase variable called `allow_cache_override`. this would be `True` by default, in which case when the workflow submits jobs, it can update all the cache_dirs of all its tasks.
3. if override is set to `False`, then the workflow would not override the Task's location

satra:
@mgxd - what this may allow us to do is to consider different levels of cache. if we think a particular workflow or task is time consuming and we should retain the cache, we may save those tasks in one cache, and other quick tasks could go to a different cache that can be wiped out. the current architecture gives us many possibilities. so it would be good to have a clean api for cache control.

mgxd:
@satra I agree - runtime resolution seems to be the most maintainable way moving forward. I added a few tests that hopefully converge to your thoughts above

pydra/engine/core.py
```diff
@@ -166,12 +167,16 @@ def checksum(self):
         """
         input_hash = self.inputs.hash
         if self.state is None:
-            self._checksum = create_checksum(self.__class__.__name__, input_hash)
+            self._checksum = create_checksum(
+                self.__class__.__name__, self.name, input_hash
```
satra: why should the name change the checksum? name does not carry any compute information, which are in inputs, workflow graph/function.

mgxd: It would be nice to make it easily available to users which directories pertain to which task, but you are right that it doesn't carry compute information. 

I'll revert for now but open up an issue regarding this.

satra: the cache will no longer behave like current nipype. this mapping exists and is exposed with `output_dir` and `result` at the task/workflow level. @djarecka - has tests for this.

pydra/engine/core.py
```diff
@@ -205,9 +206,11 @@ def cache_dir(self):
     @cache_dir.setter
     def cache_dir(self, location):
         if location is not None:
+            self._custom_cache_dir = True
```
satra: do you still need the `_custom_cache_dir`?

mgxd: yes, it's to prevent tasks nesting within a workflow's default `/tmp/path` `cache_dir`

satra: wouldn't the other clause automatically override the task's `cache_dir` ?

pydra/engine/submitter.py
```diff
@@ -30,6 +30,8 @@ def __call__(self, runnable, cache_locations=None):
         if is_workflow(runnable):
             for nd in runnable.graph.nodes:
                 runnable.create_connections(nd)
+                if runnable._custom_cache_dir and nd.allow_cache_override:
```
satra:
```suggestion
                if nd.allow_cache_override:
```

pydra/engine/core.py
djarecka: yes, `output_dir` should contain all directories with results

djarecka: i'm checking `output_dir` in all/most workflow tests. If `self.state` than the `output_dir` contains multiple paths

PR #101, opened 2019-07-26
ENH: Add pydra.mark module with initial annotate and task decorators

effigies:
## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Summary
As suggested in https://github.com/nipype/pydra/pull/95#issuecomment-514299672, we can use a `pytest.mark`-inspired namespace to provide a library of useful decorators that can keep the functionality of each decorator tightly focused.

This replaces the `pydra.engine.task.to_task` decorator with `pydra.mark.task`, which feels a little more declarative, IMO.

## Checklist

- [x] All tests passing

## Acknowledgment

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -54,6 +54,7 @@ tests =
     %(test)s
 dev = 
     %(test)s
+    black
     pre-commit
 all =
     %(dev)s
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,9 +85,9 @@ before_script:
 script:
   - |
     if [ "$CHECK_TYPE" = "test" ]; then
-       pytest -vs -n auto --cov pydra --cov-config .coveragerc --cov-report xml:cov.xml --doctest-modules pydra
+        pytest -vs -n auto --cov pydra --cov-config .coveragerc --cov-report xml:cov.xml --doctest-modules pydra
     elif [ "$CHECK_TYPE" = "style" ]; then
-        black pydra setup.py
+        black --check pydra setup.py
     fi
 
 after_script:
--- a/pydra/engine/audit.py
+++ b/pydra/engine/audit.py
@@ -6,9 +6,7 @@ from .helpers import ensure_list, gather_runtime_info
 
 
 class Audit:
-    def __init__(
-        self, audit_flags, messengers, messenger_args, develop=None
-    ):
+    def __init__(self, audit_flags, messengers, messenger_args, develop=None):
         self.audit_flags = audit_flags
         self.messengers = ensure_list(messengers)
         self.messenger_args = messenger_args
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -147,14 +147,6 @@ class FunctionTask(TaskBase):
         return self.output_
 
 
-def to_task(func_to_decorate):
-    def create_func(**original_kwargs):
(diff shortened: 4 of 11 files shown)


djarecka:
@effigies - i was just working on it,  so will just abandon and review your PR. Don't have opinion regarding renaming the decorator

effigies:
Sorry. I should have given you a heads up when I started a day or two ago...

djarecka:
@effigies - no worries, I talked about it with Satra yesterday and I've only started changing my branch

setup.cfg
```diff
@@ -54,6 +54,7 @@ tests =
     %(test)s
 dev = 
     %(test)s
+    black
```
djarecka: i've just double checked that `pre commit` will install `black` during the first commit, but it's fine if you want to add it here

djarecka:
I think it would be nice to add tests not only for functions annotations, but how pydra deals with the `OutputSpec`. I can add them if you want

effigies:
Shouldn't that be part of testing `FunctionTask`?

djarecka:
@effigies - yes, I was thinking about adding tests to `test_task`. 

And I also realized that we were thinking about slightly different use case for `annotate`, we would like to use it to create a name for the output (instead of using just `return`), you could see [my example](https://github.com/nipype/pydra/pull/95/files#diff-7c80889b6dbc9a56fcd4821c277e3e95R255)  or [this one for multiple outputs](https://github.com/nipype/pydra/pull/95/files#diff-7c80889b6dbc9a56fcd4821c277e3e95R271)

So in this version with two layers of decorators, I think we would like to create [this kind of custom  annotation](https://github.com/nipype/pydra/blob/master/pydra/engine/tests/test_task.py#L36)

effigies:
I think by annotating `{'return': {'out1': type_info_1, 'out2': type_info_2}}` we can do this.

djarecka:
yes, I think this would work. We could either use the names and be sure that `FunctionTask` will end up [here](https://github.com/nipype/pydra/blob/master/pydra/engine/task.py#L109) or without names (half annotation) and be sure that we end up [here](https://github.com/nipype/pydra/blob/master/pydra/engine/task.py#L115)

Unless we want to remove "half annotations", @satra - any thoughts?

effigies:
Want to check out the latest commit? It might be best split out into a separate PR where we can discuss the full range of valid return annotations and individually test them.

satra:
@djarecka - i think @effigies approach should work fine.

djarecka:
@satra - my question was if we want to keep half annotation in general. but could ignore this question for now

satra:
@djarecka - what do you mean by `half annotation`?

i think these are all the possibilities, and we should support all.

1. inputs annotated
2. outputs annotated
3. both annotated
4 neither annotated

djarecka:
@satra I believe you started using term "half annotation" in the `test_task`. It means that you annotate the type (and number of elements), but without providing the names.

I've just created a small PRs with two new test, one with "full annotation", the other with "half annotation":
https://github.com/effigies/pydra/pull/1

djarecka:
we don't have to think about removing the half annotation now, It's working, so I'm merging and we can come back to this another time!

satra:
thanks @djarecka for clarifying half annotation

Issue #90, opened 2019-07-19
[CI] Flaky execution


mgxd:
Rarely, event loop is missing from travis thread, see:
https://travis-ci.org/nipype/pydra/jobs/560805001

effigies:
I wonder if this is a `pytest-xdist` interaction.

satra:
it seems like this may have been resolved with changes made by @mgxd in #121

PR #105, opened 2019-07-29
CI: Check minimum dependencies

effigies:
## Types of changes

## Summary
Adds a CI entry to check that our listed minimum dependencies are accurate. Using a `min-requirements.txt` to keep the number of places to check minimal, though open to alternatives.

## Checklist

- [x] All tests passing

## Acknowledgment


satra:
would the pre-option still check other versions?

effigies:
Interpreting that as "Does the `PIP_ARGS=--pre` entry still check new versions?". If that's the question, yes, though I'll check the outputs to be sure.

effigies:
Looks like it, though there aren't pre-releases of either of our two dependencies right now. Except on this new entry, we install cloudpickle 1.2.1 and filelock 3.0.12.

The [failing test](https://travis-ci.org/nipype/pydra/jobs/565008427) is #90. I don't know if there's anything interesting to be divined from those logs, though...

effigies:
RE filelock, I don't have a good reason for picking [3.0.0](https://github.com/benediktschmitt/py-filelock/releases/tag/v3.0.0), except it's older than both pydra and the filelock docs.

As to going with more recent release, I don't see any critical failures being reported (looking through issues and PRs) in this window, so it's not obvious we need to be more restrictive.

Could be less restrictive, but not clear how much benefit there is to finding the absolute oldest version that does what we need.

In any case, I think it's a good idea to put a minimum version in the `install_requires` before merge. I was also thinking of writing a quick script to synchronize `min-requirements.txt` from `setup.cfg`.

Issue #35, opened 2018-12-21
Implement registered action hooks


satra:
Hooks will allow automations to kick in for specific events. 

Possible enumerated locations for hooks:
task_init_pre
task_init_post
run_pre
run_post
run_task_pre
run_task_post
workflow_init_pre
workflow_init_post
workflow_run_pre
workflow_run_post

Hooks can be registered through the Task/Node API.

satra:
closed by #109

PR #109, opened 2019-07-31
ENH: action hooks

mgxd:
## Types of changes

## Summary

Kicks off #35 - adds a `TaskHook` class which contains 4 registered task hooks:
- `pre_run`: before auditing starts
- `pre_run_task`: before task's `_run_task` is called
- `post_run_task`: after task's `_run_task` is called and results are collected
- `post_run`: at the end of a task's `_run` method

Sample use case:
https://gist.github.com/mgxd/e3992eee57e8e24aa25d3f7c5db94b36

## Checklist

- [x] All tests passing

## Acknowledgment


djarecka:
I wondering if `TaskHook` shouldn't take arguments that would define the functions for `pre_run`, etc., instead of setting this functions later (as in the [test](https://github.com/nipype/pydra/pull/109/files#diff-7c80889b6dbc9a56fcd4821c277e3e95R379))

PR #113, opened 2019-08-02
adding slurm to plugins in tests

djarecka:
## Summary

I've realized that slurm was not added to the tests as a new plugin. Not sure if it was any specific reason, but I'm actually having problems with it, travis has timeouts on various tests, e.g.:
https://travis-ci.org/djarecka/pydra/jobs/566758907

## Checklist

- [ ] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/tests/test_node_task.py
+++ b/pydra/engine/tests/test_node_task.py
@@ -8,7 +8,10 @@ from .utils import fun_addtwo, fun_addvar, moment, fun_div
 from ..core import TaskBase
 from ..submitter import Submitter
 
-Plugins = ["cf"]
+if bool(shutil.which("sbatch")):
+    Plugins = ["cf", "slurm"]
+else:
+    Plugins = ["cf"]
 
 
 @pytest.fixture(scope="module")
--- a/pydra/engine/tests/test_workflow.py
+++ b/pydra/engine/tests/test_workflow.py
@@ -9,7 +9,10 @@ from ..core import Workflow
 from ... import mark
 
 
-Plugins = ["cf"]
+if bool(shutil.which("sbatch")):
+    Plugins = ["cf", "slurm"]
+else:
+    Plugins = ["cf"]
 
 
 @pytest.mark.parametrize("plugin", Plugins)


satra:
@mgxd - can we not oversubscribe in the slurm config? our jobs are not intensive, so they should run. 

also we should have a slurm option to limit the maximum number of jobs that the worker will submit.

mgxd:
@satra sure, we could add infinite? nodes to the slurm config. But one thing to consider when extending the slurm testing is spawning all these jobs in an environment with actual limits (ie, running pydra tests on openmind)

re: limiting max jobs - definitely something we want to implement, though this may be tricky in this current state, will have to play around with this for a bit

satra:
> one thing to consider when extending the slurm testing is spawning all these jobs in an environment with actual limits (ie, running pydra tests on openmind)

that will require the slurm worker to understand cluster configurations. also there seems to be a more fundamental problem as to why the tests would fail with 5 nodes. the tests should run with 1 node or 5 nodes or 40 nodes.

so i feel there is something deeper that's causing the problems.

djarecka:
after merging Mathias PR, there is no timeouts, but some tests are failing:
- tests that have wf as a node and state
- some test that uses cached results. They are failing when comparing time of the execution, so maybe the test should be changed, but would have to double check.

mgxd:
> the tests should run with 1 node or 5 nodes or 40 nodes.

From my understanding, we wanted to allow the execution of each task as a separate job. So consider the following workflow

```
wf = Workflow()
wf.add(task1, x=wf.x)
wf.add(task2, x=task1.x)
```

The slurm worker will first submit the `wf` as a job. If the `wf` doesn't have a plugin defined, the slurm plugin will be propagated to the tasks, and task1 will be submitted as well.

If there are no available cores for task1 to run on, we'll reach a deadlock, since `wf` is executing but waiting for task1 to finish.

satra:
> we wanted to allow the execution of each task as a separate job. 

not blindly though. in your example the workflow should not be submitted to slurm, unless the workflow plugin is non-slurm. for slurm plugin, the workflow should continue to be handled/run on the primary loop. i.e. just the workflow execution should not take up a slurm resource.

satra:
@mgxd - the slurm worker should work with a slurm cluster containing 1 core. otherwise you are always going to run into this deadlock under many situations. so the deadlock issue has to be resolved. it's not a resource issue, this is a programmatic one.

djarecka:
@satra @mgxd - any reason to not add the tests for slurm? They are passing now, but were not added in the previous PRs

PR #114, opened 2019-08-02
tst: increase available nodes in slurm image

mgxd:
## Types of changes

## Summary

Increases available compute nodes in the slurm docker image from 5 to 40.

## Checklist

- [ ] All tests passing

## Acknowledgment

### DIFF
--- a/ci/slurm/Dockerfile
+++ b/ci/slurm/Dockerfile
@@ -10,3 +10,4 @@ RUN conda update -yq --all conda \
     && conda config --set always_yes yes \
     && conda config --add channels conda-forge \
     && rm ~/miniconda.sh
+COPY ./slurm.conf /etc/slurm/slurm.conf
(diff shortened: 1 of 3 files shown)


satra:
is this still needed?

mgxd:
this just reflects the version of the image we're pulling (`mgxd/slurm:19.05.1`) and makes the test a bit less flaky.

In the future we should probably shift to building the image and pushing whenever the dockerfile is updated to confirm it matches up to the docker image, but this should be fine for now

satra:
> makes the test a bit less flaky.

what helps make it less flaky?

mgxd:
sacct was occasionally returning `Unknown` for some jobs' "Eligible" times. I'm not 100%, but my guess is this happens when the slurmd / slurmdbd gets overloaded with requests.

PR #129, opened 2019-08-16
[WIP] example with workflow reruns

djarecka:
@satra - see [the example](https://github.com/nipype/pydra/compare/master...djarecka:wf_reset?expand=1#diff-61004b9f88d892cca950a805bfa61526R1913) I added 
It actually does not create a new directory


djarecka:
will create a new pr with more changes

PR #130, opened 2019-08-17
changes in checksums/results and some fixes (fixes#107)

### LINKED ISSUE #107: workflow_as_node does not update the input (if it was run already)
If I create a `wfnd` first and later use within `wf` that has different input it works properly only if `wfnd` was not run before `wf`. See the tests:

```
@pytest.mark.parametrize("plugin", Plugins)
def test_wfasnd_wfndupdate(plugin):
    """ workflow as a node
        workflow-node with one task and no splitter
        wfasnode input is updated to use the main workflow input
    """

    wfnd = Workflow(name="wfnd", input_spec=["x"], x=2)
    wfnd.add(add2(name="add2", x=wfnd.lzin.x))
    wfnd.set_output([("out", wfnd.add2.lzout.out)])

    wf = Workflow(name="wf", input_spec=["x"], x=3)
    wfnd.inputs.x = wf.lzin.x
    wf.add(wfnd)
    wf.set_output([("out", wf.wfnd.lzout.out)])
    wf.plugin = plugin

    with Submitter(plugin=plugin) as sub:
        sub(wf)

    results = wf.result()
    assert results.output.out == 5
    assert wf.output_dir.exists()


@pytest.mark.xfail(reason="wfnd is not updating input for it's nodes")
@pytest.mark.parametrize("plugin", Plugins)
def test_wfasnd_wfndupdate_rerun(plugin):
    """ workflow as a node
        workflow-node with one task and no splitter
        wfasnode is run first and later is
        updated to use the main workflow input
    """

    wfnd = Workflow(name="wfnd", input_spec=["x"], x=2)
    wfnd.add(add2(name="add2", x=wfnd.lzin.x))
    wfnd.set_output([("out", wfnd.add2.lzout.out)])
    with Submitter(plugin=plugin) as sub:
        sub(wfnd)

     wf = Workflow(name="wf", input_spec=["x"], x=3)
    # trying to set before
    wfnd.inputs.x = wf.lzin.x
    wf.add(wfnd)
    # trying to set after add...
    wf.wfnd.inputs.x = wf.lzin.x
    wf.set_output([("out", wf.wfnd.lzout.out)])
    wf.plugin = plugin

    results = wf.result()
    assert results.output.out == 5
    assert wf.output_dir.exists()
```

Haven't debug too much, but it looks like in the second test the input `wf.wfnd.inputs.x` is not taken from the `wf.inputs` (it shows `x=LF('wf', 'x')`), so the `wf.wfnd.add2` takes the previously set input, i.e. 2

djarecka:
## Types of changes

## Summary

- [x] removing results_dict and using a new method `checksum_states` to get all checksums

- [x] restarting connections after running workflows

  -  test for workflows that are run twice (and use results right away)

  - closes #107

## Checklist

- [x] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -165,6 +165,10 @@ class TaskBase:
     def checksum(self):
         """calculating checksum
         """
+        # if checksum is called before run the _graph_checksums is not ready
+        if is_workflow(self) and self.inputs._graph_checksums is None:
+            self.inputs._graph_checksums = [nd.checksum for nd in self.graph_sorted]
+
         input_hash = self.inputs.hash
         if self.state is None:
             self._checksum = create_checksum(self.__class__.__name__, input_hash)
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -281,7 +281,7 @@ class TaskBase:
         return res
 
     def _run(self, **kwargs):
-        # self.inputs = dc.replace(self.inputs, **kwargs) don't need it?
+        self.inputs = dc.replace(self.inputs, **kwargs)
         checksum = self.checksum
         lockfile = self.cache_dir / (checksum + ".lock")
         # Eagerly retrieve cached
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -543,7 +543,6 @@ class Workflow(TaskBase):
         """adding a task to the workflow"""
         if not is_task(task):
             raise ValueError("Unknown workflow element: {!r}".format(task))
-        task._reset()
         self.graph.add_nodes(task)
         self.name2obj[task.name] = task
         self._last_added = task
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -95,6 +95,8 @@ class TaskBase:
             for field in dc.fields(klass)
             if field.name not in ["_func", "_graph_checksums"]
         ]
+        # dictionary to save the connections with lazy fields
+        self.inp_lf = {}
         self.state = None
         self._output = {}
         self._result = {}
@@ -454,6 +456,15 @@ class TaskBase:
             result = load_result(checksum, self.cache_locations)
             return result
 
+    def _reset(self):
+        """resetting the connections between inputs and LazyFields"""
+        for field in dc.fields(self.inputs):
+            if field.name in self.inp_lf:
+                setattr(self.inputs, field.name, self.inp_lf[field.name])
+        if is_workflow(self):
+            for task in self.graph.nodes:
+                task._reset()
+
 
 class Workflow(TaskBase):
     def __init__(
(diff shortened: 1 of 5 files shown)


pydra/engine/core.py
```diff
@@ -260,7 +277,7 @@ def __call__(self, submitter=None, plugin=None, **kwargs):
         return res
 
     def _run(self, **kwargs):
-        self.inputs = dc.replace(self.inputs, **kwargs)
+        # self.inputs = dc.replace(self.inputs, **kwargs) don't need it?
```
satra: this was there to ensure that inputs could be set via the run command (when it was not a private function) or as a function call. as long as this is the case within `__call__` it would be ok to remove this. i.e. a user should be able to set inputs during the call to a task (much like a python function). 

if `**kwargs` is not being used, it should be removed.

pydra/engine/core.py
```diff
@@ -522,6 +543,7 @@ def add(self, task):
         """adding a task to the workflow"""
         if not is_task(task):
             raise ValueError("Unknown workflow element: {!r}".format(task))
+        task._reset()
```
satra: does this need to happen at every `add` call? or could this be done just prior to `run` or perhaps more preferably at the end of a `run`?

pydra/engine/core.py
```diff
@@ -558,7 +582,7 @@ def create_connections(self, task):
                 task.state = state.State(task.name, other_states=other_states)
 
     async def _run(self, submitter=None, **kwargs):
-        self.inputs = dc.replace(self.inputs, **kwargs)
+        # self.inputs = dc.replace(self.inputs, **kwargs) don't need it?
```
satra: same question as before.

satra:
overall looks good. the only question i have is whether the reset can be called at the end of a run by the submitter, instead of during `add`. if so it may also help with the checksum issue.

```python
checksum1  = task.checksum
result = task(param1=2, param2=3)
checksum2 = task.checksum
assert checksum1 == checksum2
```

pydra/engine/core.py
djarecka: oh, i see, we have `kwargs` in the `__call__` that is passing to `_run`, but had no tests for it, so I couldn't find who is using this. will keep it and add a test

djarecka:
@satra - I'm just working on moving the `reset` to `Submitter`, so running workflow twice doesn't create a new output_dir, but other tests just started failing... might be related to `asyncio`, will try to think about it a bit more

djarecka:
@satra - at the end I just moved `wf._reset` to the end of `Submitter.__call__`. Wanted to put in more specific place but was having some issues.

about comparing the  `checksum`, in addition to the issue with resetting the connections, there was one more thing - we added `_graph_checksums` to inputs, and this is usually created in the submitter when the graph has all the nodes and it can sort it.
I've added a line to `checksum` that allow to set this before running.

pydra/engine/core.py
djarecka: we do not allow running a workflow without a submitter, so I believe this can't be used

pydra/engine/core.py
djarecka: added [a test](https://github.com/nipype/pydra/pull/130/files#diff-871957728874435d337bd031b6bcb53cR216)

PR #136, opened 2019-08-19
adding description to State, some cleaning

djarecka:
## Types of changes

## Summary
- adding description to State,
- renaming auxiliary
- some cleaning, removing functions that are not used anymore

## Checklist

- [x] All tests passing

## Acknowledgment


pydra/engine/helpers_state.py
```diff
@@ -1,3 +1,5 @@
+""" additional functions used mostly by the State class """
```
mgxd: Sorry for the late review here:

@djarecka any reason for separating helpers into their own file instead of including them in `state.py`? 
Not necessarily as part of the `State` class, but within the same file.

djarecka: nothing special, the same reason way we have other files with helping functions.

PR #139, opened 2019-08-22
ShellCommandTask

djarecka:
## Types of changes

## Summary

- fixing `__getstate__` in `Result`
- fixing  execution so it can run with submitter (adding read_and_display that is not a coroutine)
- adding tests

## Checklist

- [x] All tests passing

## Acknowledgment


djarecka:
I'm trying to explore `ShellCommandTask` and add tests for it, but I'm having issue when using `Submitter` even for the simplest task, see `test_node_shelltask.py::test_shell_cmd_1a `
 
@mgxd - any idea?

mgxd:
@djarecka I think the problem lies with
https://github.com/nipype/pydra/blob/85e7b0e625fd633a1401d3c535f9467609d4fe61/pydra/engine/task.py#L202-L206

specifically `execute()`
https://github.com/nipype/pydra/blob/85e7b0e625fd633a1401d3c535f9467609d4fe61/pydra/engine/helpers.py#L165-L168

It is sending a coroutine to an already running loop - we may want to add a synchronous subprocess call that mimics `execute()` and let submitter / workers handle multiple processes.
cc @satra

djarecka:
@satra - Mathias helped me with the execution and it works now with submitter, so I added examples for tasks with `State` and also a simple `Workflow`. Could you please check if the tests make sense for you. I decided to have splitter over `executables` and `args`, but maybe should have also other exampels

pydra/engine/helpers.py
```diff
@@ -158,13 +159,46 @@ def make_klass(spec):
...
+    else:
+        return rc, stdout, stderr
+
+
+def read_and_display(*cmd, strip=False, hide_display=False):
+    """Capture cmd's stdout
 
+    """
+    try:
+        process = sp.run(cmd, stdout=sp.PIPE, stderr=sp.PIPE)
```
satra: subprocess should take in environment settings, which should be settable at the shelltask level.

pydra/engine/helpers.py
```diff
@@ -158,13 +159,46 @@ def make_klass(spec):
...
+        )
+
+
+# run the event loop with coroutine read_and_display_async
+# or run read_and_display if a loop is already running
+def execute(cmd, strip=False):
     loop = get_open_loop()
-    rc, stdout, stderr = loop.run_until_complete(read_and_display(*cmd))
+    if loop.is_running():
+        rc, stdout, stderr = read_and_display(*cmd, strip=strip)
```
satra: is it not possible to just await here?

pydra/engine/tests/test_shelltask.py
```diff
@@ -0,0 +1,320 @@
...
+
+def test_shell_cmd_1_nosubm(tmpdir):
+    """ simple command, no arguments
+        no submitter
+    """
+    cmd = ["pwd"]
+    shelly = ShellCommandTask(name="shelly", executable=cmd)
+    assert shelly.cmdline == " ".join(cmd)
+    res = shelly()
+    assert res.output.stdout == str(shelly.output_dir) + "\n"
```
satra:
```suggestion
    assert res.output.stdout.strip() == str(shelly.output_dir)
```

pydra/engine/tests/test_shelltask.py
```diff
@@ -0,0 +1,320 @@
...
+    assert res.output.stderr == ""
+
+
+@pytest.mark.parametrize("plugin", Plugins)
+def test_shell_cmd_1_strip(plugin):
+    """ simple command, no arguments
+        strip option to remove \n at the end os stdout
+    """
+    cmd = ["pwd"]
+    shelly = ShellCommandTask(name="shelly", executable=cmd, strip=True)
```
satra: does strip remove just the last one?

djarecka: yes, but can changed

pydra/engine/helpers.py
mgxd: not without a refactor to make `Task._run_task()` async.

satra:
@mgxd and @djarecka - i restarted the slurm build again. do make sure that build works before merging this. it would be good to track down the loop closed error. yes, it appears stochastic, but that's what would be good to figure out.

djarecka:
@satra - this is a different error than the loop error we sometimes have. 

here we have problems with slurm even in old tests, so it looks I broke something slurm related

djarecka:
@satra - tests are passing so if we want, we can merge it, but there are some issues that we have to think about. For example the `cmdline` doesn't return correct command when a splitter is used

satra:
perhaps cmdline can behave like `output_directory` provide cmdline if result is available but not otherwise.

Issue #141, opened 2019-08-24
TODO for Shell Task and ContainerTask


djarecka:
There are some issues tak don't have to be fixed right away, but should be done soon:
(will be updated)

- [x] cmdline for Shell: doesn't make sense for tasks with splitter

- [x] inputs for Containers: 
      - checking if has image and it's proper image, raising exceptions; 
      - bindings: allowing for tuple (not only list), tuple should be 3 or 2 elements (without `ro`)
      - checking if the local directory exists

- [x] adding an option to `Task` that converts it to the `DockerTask`: see @satra comments in #144 

- [x] provide option to have position for arguments

satra:
is this now taken care of?

djarecka:
not all

djarecka:
addressed in #184, #186

PR #142, opened 2019-08-24
tests for docker task

djarecka:
## Types of changes

## Summary
- adding tests for DockerTasks: tasks, tasks with State, tasks in wf

decided to not fix some of the issues for now, but created #141 
## Checklist

- [x] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -218,6 +218,7 @@ class ContainerTask(ShellCommandTask):
         messengers=None,
         messenger_args=None,
         cache_dir=None,
+        strip=False,
         **kwargs,
     ):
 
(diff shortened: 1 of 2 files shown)


satra:
lgtm - any idea why coverage is decreasing?

djarecka:
@satra - it annoyed me too and I was checking it earlier. Not completely sure, but IMO what you see here might be not updated, since when go to codecov and see the report, it looks like this:
https://codecov.io/gh/nipype/pydra/pull/142

I'm guessing that codecov decreased the the coverage at the beginning because slurm travis was failing. 

It's still a bit mystery to me, since only some new tests were failing for slurm, but this is the best explanations I could think about.

Don't also know why it's not updating.

djarecka:
exactly the same tests give different coverage report here:
https://github.com/nipype/pydra/pull/143

djarecka:
was hoping that after merge the codecov will use a newer report, but no... it will have to wait for another PR

PR #144, opened 2019-08-26
tests for docker tasks that runs python scripts

djarecka:
## Types of changes

## Summary

adding tests for docker tasks that runs python scripts to save and load file

## Checklist

- [ ] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/tests/data_tests/loading.py
+++ b/pydra/engine/tests/data_tests/loading.py
@@ -1,0 +1,17 @@
+import os
+
+
+def loading(filename):
+    with open(filename, "r") as f:
+        txt = f.read()
+    print(txt)
+
+
+if __name__ == "__main__":
+    from argparse import ArgumentParser, RawTextHelpFormatter
+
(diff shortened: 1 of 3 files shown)


satra:
@djarecka - while you are thinking of this, one of the things yet to implement is proper file and environment isolation.

consider a situation where a script like your python code above can be run locally, but now i want it to run in a docker environment. it should be easy for me to:
- set some attribute of the task to run in docker or 
- convert it to a dockertask without having to create a new dockertask or
- pipe in a new environment into the task from another node (in a workflow context)

more generally this could apply to commands available on the host (e.g., BET from FSL) that i now want to run via some docker image. so same function/commandline, but different environments. (this also aligns with needs for testkraken as well).

from a user perspective it would be like splitting over execution environments.

djarecka:
@satra - yes, I was also thinking that an option to run any `FunctionTask` or `ShellCmdTask` in docker could be nice. For now, I added it to my general issue #141

PR #152, opened 2019-09-23
[wip] adding customized ShellOutputSpec

djarecka:
## Types of changes

## Summary

adding/testing customized `ShellOutputSpec`

## Checklist

- [ ] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/tests/test_shelltask.py
+++ b/pydra/engine/tests/test_shelltask.py
@@ -3,11 +3,12 @@
 import typing as ty
 import os, shutil
 import pytest
-
+import dataclasses as dc
 
 from ..task import ShellCommandTask
 from ..submitter import Submitter
 from ..core import Workflow
+from ..specs import ShellOutSpec, SpecInfo
 
 if bool(shutil.which("sbatch")):
     Plugins = ["cf", "slurm"]
@@ -15,7 +16,7 @@ else:
     Plugins = ["cf"]
 
 
-def test_shell_cmd_1_nosubm(tmpdir):
+def test_shell_cmd_1_nosubm():
     """ simple command, no arguments
         no submitter
     """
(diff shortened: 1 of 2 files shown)


djarecka:
@satra - I know you didn't want me to specify another argument to save files paths (like [here](https://github.com/djarecka/pydra/commit/f412b219de023ceeb227f7026b4c424f46b049b6)), and user should defined his/her own spec,  but  not sure if I understood you well

pydra/engine/tests/test_shelltask.py
```diff
@@ -320,3 +321,30 @@ def test_wf_shell_cmd_1(plugin):
...
+def test_shell_cmd_outputspec_1(tmpdir):
+    """
+        customised OutputSpec
+    """
+    cmd = ["touch"]
+    args = ["newfile_tmp.txt"]
+
+    @dc.dataclass
+    class MyOutput(ShellOutSpec):
+        newfile: str = "newfile_tmp.txt"
```
satra: should be of type `File`

djarecka: could change this, but in `task` I assumed that this is a relative path and I'm adding `self.output_dir`

satra:
@djarecka - your example of specifying the exact filename would be one of 4 options, here are the others.

option 2. glob metadata:

```python
        newfile: File = field(default='newfile*.txt')
```
this would automatically check if the default is a glob pattern or not and use it.

option 3. callable for key

```python
        newfile: File = field(metadata={'callable': gather_output})
```
where `gather_output(keyname, **other_args)` could have such a signature

option 4. input spec is modified

note this is in input spec.

option 1.
```python
        outfile: str = field(metadata={'name_template': f'{infile}_out.txt'})
```
this should also create a new output field called `outfile`. in addition, `name_template` could have slightly different forms: `f'{infile | reffile}_out.txt'`

option 2.
```python
        outfile: str = field(metadata={'name_template': f'{infile}_out.txt'}, 'out_field': 'namedout'})
```
this should also create a new output field called `namedout`

satra:
and we can also add `keep_extension=True` as a metadata option to keep the original file's extension.

djarecka:
the way how we use the `output_spec` is not optimal for asking users to create a new dataclasses for spec. I'm planning to come back to my previous idea to ask uses to provide just additional fields. Will open a new PR

satra:
@djarecka - this is true for inputs and outputs. the reason for `SpecInfo` is to allow pickling of the specs for generating a `dataclass`. 

i would focus first on making sure the spec piece works with the right metadata before simplifying the api to create it.

djarecka:
@satra - I'm not sure what do you mean "making sure the spec piece works with the right metadata". for now I want something that is working and can be used in testkraken.

 For now I have a simple way of providing additional fields that works, and more complicated (this RP) that it doesn't work, so I like the simple way more and more...

satra:
all of the options i listed above have to work. that is not an option to discard them. so instead of a simple solution that only addresses a very specific use case, we have to move more systematically towards making sure we can use pydra in nipype.

djarecka:
@satra - the way I tried to do in this PR does not work with submitter (i didn't event got to the 4 options), and I don't know how to easily fix it. If you have specific idea, please let me know. 

But I also strongly believe that we should not ask users to create a dataclass only to add a file to the output of `ShellTask`

PR #153, opened 2019-09-25
customized input_spec and output_spec for ShellTask

djarecka:
## Types of changes

## Summary

this is instead of  #152, I'm testing option that user simply has to add `fields` to `SpecInfo` and base stay the same: `bases=(ShellOutSpec,)`. 

## Checklist

- [x] All tests passing

## Acknowledgment

### DIFF
--- a/ci/none.sh
+++ b/ci/none.sh
@@ -41,7 +41,7 @@ function travis_before_script {
             pip install ".[test]"
         fi
     elif [ "$CHECK_TYPE" = "style" ]; then
-        pip install black
+        pip install black==19.3b0
     fi
 }
 
--- a/ci/none.sh
+++ b/ci/none.sh
@@ -47,7 +47,7 @@ function travis_before_script {
 
 function travis_script {
     if [ "$CHECK_TYPE" = "test" ]; then
-        pytest -vs -n auto --cov pydra tutorial --cov-config .coveragerc --cov-report xml:cov.xml --doctest-modules pydra
+        pytest -vs -n auto --cov pydra tutorial --cov-config .coveragerc --cov-report xml:cov.xml pydra
     elif [ "$CHECK_TYPE" = "style" ]; then
         black --check pydra tools setup.py
     fi
--- a/ci/none.sh
+++ b/ci/none.sh
@@ -49,7 +49,7 @@ function travis_script {
     if [ "$CHECK_TYPE" = "test" ]; then
         pytest -vs -n auto --cov pydra tutorial --cov-config .coveragerc --cov-report xml:cov.xml --doctest-modules pydra
     elif [ "$CHECK_TYPE" = "style" ]; then
-        black -py37 --check pydra tools setup.py
+        black --py37 --check pydra tools setup.py
     fi
 }
 
--- a/ci/none.sh
+++ b/ci/none.sh
@@ -49,7 +49,7 @@ function travis_script {
     if [ "$CHECK_TYPE" = "test" ]; then
         pytest -vs -n auto --cov pydra tutorial --cov-config .coveragerc --cov-report xml:cov.xml --doctest-modules pydra
     elif [ "$CHECK_TYPE" = "style" ]; then
-        black --py37 --check pydra tools setup.py
+        black --check pydra tools setup.py
     fi
 }
 
--- a/ci/none.sh
+++ b/ci/none.sh
@@ -47,7 +47,7 @@ function travis_before_script {
 
 function travis_script {
     if [ "$CHECK_TYPE" = "test" ]; then
-        pytest -vs -n auto --cov pydra tutorial --cov-config .coveragerc --cov-report xml:cov.xml pydra
+        pytest -vs -n auto --cov pydra tutorial --cov-config .coveragerc --cov-report xml:cov.xml --doctest-modules pydra
     elif [ "$CHECK_TYPE" = "style" ]; then
         black --check pydra tools setup.py
     fi
--- a/ci/none.sh
+++ b/ci/none.sh
@@ -49,7 +49,7 @@ function travis_script {
     if [ "$CHECK_TYPE" = "test" ]; then
         pytest -vs -n auto --cov pydra tutorial --cov-config .coveragerc --cov-report xml:cov.xml --doctest-modules pydra
     elif [ "$CHECK_TYPE" = "style" ]; then
-        black --check pydra tools setup.py
+        black -py37 --check pydra tools setup.py
     fi
 }
 
--- a/pydra/engine/helpers_file.py
+++ b/pydra/engine/helpers_file.py
@@ -114,17 +114,17 @@ def hash_infile(afile, chunk_len=8192, crypto=hashlib.md5, raise_notfound=False)
     """
     Computes hash of a file using 'crypto' module
 
-    >>> hash_infile('smri_ants_registration_settings.json')
-    'f225785dfb0db9032aa5a0e4f2c730ad'
-
-    >>> hash_infile('surf01.vtk')
-    'fdf1cf359b4e346034372cdeb58f9a88'
-
-    >>> hash_infile('spminfo')
-    '0dc55e3888c98a182dab179b976dfffc'
-
-    >>> hash_infile('fsl_motion_outliers_fd.txt')
-    'defd1812c22405b1ee4431aac5bbdd73'
+    # >>> hash_infile('smri_ants_registration_settings.json')
+    # 'f225785dfb0db9032aa5a0e4f2c730ad'
+    #
+    # >>> hash_infile('surf01.vtk')
+    # 'fdf1cf359b4e346034372cdeb58f9a88'
+    #
+    # >>> hash_infile('spminfo')
+    # '0dc55e3888c98a182dab179b976dfffc'
+    #
+    # >>> hash_infile('fsl_motion_outliers_fd.txt')
+    # 'defd1812c22405b1ee4431aac5bbdd73'
 
 
     """
--- a/pydra/engine/helpers_file.py
+++ b/pydra/engine/helpers_file.py
@@ -1,0 +1,479 @@
+# -*- coding: utf-8 -*-
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+
+""" functions copied from nipype, removed parts that were related to py2"""
+
+import subprocess as sp
+import hashlib
+from hashlib import md5
+import os
+import os.path as op
+import re
+import shutil
+import posixpath
+from builtins import str, bytes, open
+
+
+related_filetype_sets = [(".hdr", ".img", ".mat"), (".nii", ".mat"), (".BRIK", ".HEAD")]
+
+
+import logging
+
+logger = logging.getLogger("pydra")
+
+
+def split_filename(fname):
+    """Split a filename into parts: path, base filename and extension.
+
+    Parameters
+    ----------
+    fname : str
+        file or path name
+
+    Returns
+    -------
+    pth : str
+        base path from fname
+    fname : str
+        filename from fname, without extension
+    ext : str
+        file extension from fname
+
+    Examples
+    --------
+    >>> from pydra.engine.helpers_file import split_filename
+    >>> pth, fname, ext = split_filename('/home/data/subject.nii.gz')
+    >>> pth
+    '/home/data'
+
+    >>> fname
+    'subject'
+
+    >>> ext
+    '.nii.gz'
+
+    """
+
+    special_extensions = [".nii.gz", ".tar.gz", ".niml.dset"]
+
+    pth = op.dirname(fname)
+    fname = op.basename(fname)
+
+    ext = None
+    for special_ext in special_extensions:
+        ext_len = len(special_ext)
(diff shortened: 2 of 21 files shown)


djarecka:
@satra - please see if this way of creating customized `output_spec` can work. For now I only created two types from your previous comment (see #152), so you can use [specific names](https://github.com/nipype/pydra/pull/153/files#diff-d34220e5398ccac7bb1d867bbf76a63bR349) or [use a wildcard](https://github.com/nipype/pydra/pull/153/files#diff-d34220e5398ccac7bb1d867bbf76a63bR430)

satra:
@djarecka - looks good - may want to check the slurm output in the docker container.

djarecka:
yes, I've been debugging slurm this morning. slurm doesn't pass pydra exceptions

@mgxd - maybe we can help me with this. I realized that pydra exception is saved in a file 
`.../SlurmWorker_scripts/.../slurm-??.out`, that is part of [`sargs` (in `--output`)](https://github.com/nipype/pydra/blob/master/pydra/engine/workers.py#L209)
So when [read_and_display_async](https://github.com/nipype/pydra/blob/master/pydra/engine/helpers.py#L138) runs everything is fine, [stdout](https://github.com/nipype/pydra/blob/master/pydra/engine/helpers.py#L152) return sometime "FAILED", but nothing is raised. 

Maybe you remember why this is implemented this way, and what would be the best way of fixing it.
Do you think that [read_stream_and_display](https://github.com/nipype/pydra/blob/master/pydra/engine/helpers.py#L123) should be change in some specific way?

djarecka:
@satra - I've decided to work on `input_spec` as well to implement the other option you asked for. So now, you can have an arbitrary number of options/arguments to the command, but you just need to provide a position - I decided to use `metadata` from `dc.field` to specify it, see [test here](https://github.com/nipype/pydra/pull/153/commits/2dabb02b920ef4b5683116dec249cdbf4a7985e7#diff-d34220e5398ccac7bb1d867bbf76a63bR398)

regarding your option described in #152, I'm not sure if I understood the option 3 and 4 correctly:
- for option 3, did you have [something like this](https://github.com/nipype/pydra/pull/153/commits/f92e3ba5be41e3204490f76257365931b273baf5#diff-d34220e5398ccac7bb1d867bbf76a63bR497) in mind?

- for option 4, I created `input_spec` like [this](https://github.com/nipype/pydra/pull/153/commits/d2564611a76f798f8e548f659041e51fca67f708#diff-d34220e5398ccac7bb1d867bbf76a63bR631), and fields that have `output=True` I'm adding to the `output_spec`

But I'm not sure why we have to do it in `input_spec`, see [a very similar example](https://github.com/nipype/pydra/pull/153/commits/9e93d98e2c46a8d86602cdf030c799759a640a0d#diff-d34220e5398ccac7bb1d867bbf76a63bR695) that uses `output_spec`

pydra/engine/task.py
```diff
@@ -184,14 +183,34 @@ def __init__(
...
-        return args
+                pos_args.append((pos, ensure_list(value)))
+        # sorting all elements of the command
+        pos_args.sort()
+        # if args available, they should be moved at the of the list
+        if pos_args[0][0] == -1:
+            pos_args.append(pos_args.pop(0))
+        # dropping the position index
+        cmd_args = []
+        [cmd_args.extend(el[1]) for el in pos_args]
```
satra: ??

pydra/engine/tests/test_shelltask.py
```diff
@@ -320,3 +323,457 @@ def test_wf_shell_cmd_1(plugin):
...
+
+@pytest.mark.parametrize("plugin", Plugins)
+def test_shell_cmd_inputspec_1_nosubm(plugin):
+    """ a command with executable, args and one command opt,
+        using a customized input_spec to add the opt to the command
+        in the right place that is specified in metadata["cmd_pos"]
+    """
+    cmd_exec = "echo"
+    cmd_opt = "-n"
+    cmd_args = "hello from pydra"
```
satra: is there a reason to separate `opt` from `args`. aren't all pieces args?

pydra/engine/tests/test_shelltask.py
```diff
@@ -627,6 +627,70 @@ def gather_output(keyname, output_dir):
...
+    """
+    cmd = "touch"
+    args = "newfile_tmp.txt"
+
+    my_input_spec = SpecInfo(
+        name="Input",
+        fields=[
+            (
+                "out1",
+                File,
```
satra: this would be a string on the input side, but a File on the output side.

satra:
> for option 3, did you have something like this in mind?

yes

> But I'm not sure why we have to do it in input_spec, see a very similar example that uses output_spec

there are many command line programs where the output name is specified in the input (for example the `touch` command you have). in such cases, we could simply add metadata to input to create the output spec.

pydra/engine/task.py
djarecka: I'm just removing all positions and adding all list of arguments. can rewrite

pydra/engine/tests/test_shelltask.py
djarecka: ok, makes sense

pydra/engine/task.py
```diff
@@ -209,7 +209,8 @@ def command_args(self):
             pos_args.append(pos_args.pop(0))
         # dropping the position index
         cmd_args = []
-        [cmd_args.extend(el[1]) for el in pos_args]
+        for el in pos_args:
+            cmd_args += el[1]
```
satra: i think `+=` appends, so if `el[1]` is a list, it will append a list, but the previous command would have appended all the elements of the list. so just making sure that's your intent.

djarecka: `+=` is not the same as `append`, `ll=[1]; ll+=[2,3]; all==[1,2,3]`

djarecka:
@mgxd - would you have time to look at the issue with failing test (tried to explained it a post above)

mgxd:
so this boils down to when the error is happening - because shell task is designed to be handled by slurm, all exceptions will be caught by slurm as well. atm, pydra raises a general error if any queued jobs fail

we could build some infrastructure on top of that to handle propagation between slurm / pydra. for instance, separating stdout / stderr in the `sbatch` submission  (with -o and -e respectively) and then reading and dumping the stderr file to supplement "Job Failed".

we could also mark those tests to be skipped when executing with slurm. this all depends on how in depth we want pydra's slurm support to go.

djarecka:
@mgxd -for now we can skip, but I think it should be supported.

it might be a good idea to run `sbatch` with both `--error` and `--output`, and parse the error file. We could probably also just parse the output file, since the exception is written there even now.

Right now [stderr form here](https://github.com/nipype/pydra/blob/master/pydra/engine/helpers.py#L152) gives an empty string, so it's useless. Also I'm not sure if we can ever go to the [except block](https://github.com/nipype/pydra/blob/master/pydra/engine/helpers.py#L156) with the current implementation

satra:
in terms of slurm support. we should have all the capabilities of slurm worker from nipype, but will expand that further. in general though, a job could have not finished either because of a job execution error or because it was cancelled via slurm (e.g., user, preemption, etc.,.). 

if you skip the test, please file an issue with a new label `skipped_tests` so that we can finish it

djarecka:
@satra , @mgxd - I added a simple way of getting the error from the slurm job, not sure if it work well for all cases.

satra:
@djarecka - that should cover the reasons when the actual job fails. do note that since some jobs may not write out to stderr, you may have to consider both stdout and stderr in the future.  the other way a job could fail is because of slurm and then you would need to look it up with `sacct -l -j {jobid}`

as long as this works for now, we can move forward.

mgxd:
@satra yep, we can add different checks based on the JOB_REASON_CODE

>JOB STATE CODES
Jobs typically pass through several states in the course of their execution. The typical states are PENDING, RUNNING, SUSPENDED, COMPLETING, and COMPLETED. An explanation of each state follows.
BF BOOT_FAIL
Job terminated due to launch failure, typically due to a hardware failure (e.g. unable to boot the node or block and the job can not be requeued).
CA CANCELLED
Job was explicitly cancelled by the user or system administrator. The job may or may not have been initiated.
CD COMPLETED
Job has terminated all processes on all nodes with an exit code of zero.
CF CONFIGURING
Job has been allocated resources, but are waiting for them to become ready for use (e.g. booting).
CG COMPLETING
Job is in the process of completing. Some processes on some nodes may still be active.
DL DEADLINE
Job terminated on deadline.
F FAILED
Job terminated with non-zero exit code or other failure condition.
NF NODE_FAIL
Job terminated due to failure of one or more allocated nodes.
OOM OUT_OF_MEMORY
Job experienced out of memory error.
PD PENDING
Job is awaiting resource allocation.
PR PREEMPTED
Job terminated due to preemption.
R RUNNING
Job currently has an allocation.
RD RESV_DEL_HOLD
Job is held.
RF REQUEUE_FED
Job is being requeued by a federation.
RH REQUEUE_HOLD
Held job is being requeued.
RQ REQUEUED
Completing job is being requeued.
RS RESIZING
Job is about to change size.
RV REVOKED
Sibling was removed from cluster due to other cluster starting the job.
SI SIGNALING
Job is being signaled.
SE SPECIAL_EXIT
The job was requeued in a special state. This state can be set by users, typically in EpilogSlurmctld, if the job has terminated with a particular exit value.
SO STAGE_OUT
Job is staging out files.
ST STOPPED
Job has an allocation, but execution has been stopped with SIGSTOP signal. CPUS have been retained by this job.
S SUSPENDED
Job has an allocation, but execution has been suspended and CPUs have been released for other jobs.
TO TIMEOUT
Job terminated upon reaching its time limit.

@djarecka I don't think that will work since `SlurmWorker` can handle multiple jobs simultaneously. One workaround would be to make a dictionary of `jobid: path-to-stderr-file`, but this may increase the memory footprint substantially for larger workflows

djarecka:
@mgxd - not sue if I understand, `_verify_exit_code` is takes `jobid` as an argument

pydra/engine/specs.py
```diff
@@ -50,6 +58,139 @@ def retrieve_values(self, wf, state_index=None):
...
+            "mandatory",
+            "xor",
+            "requires",
+            "default_value",
+            "position",
+            "help_string",
+            "output_file_template",
+            "output_field_name",
+            "separate_ext",
+            "argstr",
```
satra: shouldn't argstr only get added by shellspec? is there a need for this in base/function?

pydra/engine/specs.py
```diff
@@ -50,6 +58,139 @@ def retrieve_values(self, wf, state_index=None):
         for field, value in temp_values.items():
             setattr(self, field, value)
 
+    def check_metadata(self):
```
satra: check_metadata could take a flag that can add supported_keys to a base list.

ci/none.sh
```diff
@@ -47,7 +47,7 @@ function travis_before_script {
 
 function travis_script {
     if [ "$CHECK_TYPE" = "test" ]; then
-        pytest -vs -n auto --cov pydra tutorial --cov-config .coveragerc --cov-report xml:cov.xml --doctest-modules pydra
+        pytest -vs -n auto --cov pydra tutorial --cov-config .coveragerc --cov-report xml:cov.xml pydra
```
satra: this seems like we are removing some tests. are those tests not necessary?

pydra/engine/specs.py
```diff
@@ -22,6 +27,9 @@ class SpecInfo:
 class BaseSpec:
     """The base dataclass specs for all inputs and outputs"""
 
+    def collect_additional_outputs(self, input_spec, inputs, output_dir):
+        return {}
+
     @property
     def hash(self):
```
satra: shouldn't the hash take `output_field_name` into account when computing a checksum?

pydra/engine/helpers_file.py
```diff
@@ -0,0 +1,479 @@
...
+from hashlib import md5
+import os
+import os.path as op
+import re
+import shutil
+import posixpath
+from builtins import str, bytes, open
+
+
+related_filetype_sets = [(".hdr", ".img", ".mat"), (".nii", ".mat"), (".BRIK", ".HEAD")]
```
satra: perhaps add a note here that these are imaging specific, but may need to be generalized for hash computation and for copyfile.

ci/none.sh
djarecka: there are doctests in functions from `nipype/utils/filemanip.py`, and they required some data files. I probably should just copied the files from nipype

pydra/engine/specs.py
djarecka: maybe it should...

pydra/engine/helpers_file.py
djarecka: ok

pydra/engine/specs.py
djarecka: I've moved this method to shellspec after yesterday meeting

pydra/engine/specs.py
djarecka: you mean an option to add something to the list defined in the function

djarecka:
@satra -i don't know how to fix the style tests, i should be using the same black version in the pre-commit. will not worry about this for now

djarecka:
@satra - I've added a set of tests from nipype, but codecov still is not happy. I'd prefer to fix the tests coverage later

satra:
@djarecka - that's fine. perhaps open an issue to keep track of it.

PR #168, opened 2020-01-05
ENH: Use Python's ``hash`` built-in to calculate checksums

### LINKED ISSUE #167: Use Python's __hash__ whenever possible
Dataclasses define a `__hash__` member (when the appropriate arguments are given in the decorator). We could/should use the `hash()` built-in to check hashes, instead of:

https://github.com/nipype/pydra/blob/51fdbbf27aebbd9c78b9726098804015701fa198/pydra/engine/specs.py#L40-L59

The whole method could go away, defining the proper `__hash__` in File:

https://github.com/nipype/pydra/blob/51fdbbf27aebbd9c78b9726098804015701fa198/pydra/engine/specs.py#L11-L12

And make sure we set `compare` or `hash` to `False` in the appropriate fields that should not be considered for comparison and/or hashing.

There is that last exception of ``field.name not in ["_graph_checksums", "bindings"]`` - I'm not sure if a field can dynamically change their `compare` or `hash` when they are added to `bindings`. For `_graph_checksums` then the `__hash__` member of the field should be mutated (if I understood correctly the property).

Any reason not to try to stick with the `hash` built-in? I believe that has been a source of unreliability in Nipype 1.0 (I can remember it was particularly painful during the migration to unicode strings).

oesteban:
Closes #167

## Types of changes

## Checklist

- [ ] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/specs.py
+++ b/pydra/engine/specs.py
@@ -1,4 +1,5 @@
 """Task I/O specifications."""
+from hashlib import sha256
 import dataclasses as dc
 from pathlib import Path
 import os
@@ -10,6 +11,23 @@ from .helpers_file import copyfile
 
 class File(Path):
     """An :obj:`os.pathlike` object, designating a file."""
+
+    hash_chunk_length = 8192
+
+    def __hash__(self):
+        """Calculate the hash of the file based on its content."""
+        if not self.is_file():
+            raise FileNotFoundError(
+                "Path %s does not point to a file or is a broken symlink" % self)
+
+        hashed = sha256()
+        with self.open("rb") as fp:
+            while True:
+                data = fp.read(self.hash_chunk_length)
+                if not data:
+                    break
+                hashed.update(data)
+        return int(hashed.hexdigest(), 16)
 
 
 class Directory(Path):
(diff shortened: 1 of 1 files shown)


oesteban:
@satra, haven't you hit https://bugs.python.org/issue24132 when using our `File` derived from `pathlib.Path`?

effigies:
If I'm not mistaken, the main effect of defining __hash__ is to make this object insertable into a dictionary or set. My feeling is that this is a pretty drastic reinterpretation of the API, and one that could have unintended consequences.

satra:
i haven't hit the bug. 

i also wouldn't override the built in hash. that function is created for immutable objects, and in this case one can assign a new value to that variable at any time. the built in hash function has some specific requirements for python, and there is no advantage gained i think by overriding it in this case.

anibalsolon:
The applicability scenario in Pydra of redefining the __hash__ function using hashlib and similar would be for multi-process executions. The native hash function has a nondeterministic behavior (by default in python >= 3.3) across processes, so hashes will not match if comparing in one process file hashes generated in another process.

https://docs.python.org/3.3/reference/datamodel.html#object.__hash__

oesteban:
Thanks for the feedback, this doesn't seem like a good idea.

Issue #173, opened 2020-01-11
Discussion: attrs vs dataclasses


effigies:
I've spent a bit of time reading through the [attrs documentation](https://www.attrs.org/en/stable), and it looks pretty straightforward and capable of providing much of what we use `traits` for in Nipype 1. And as a pure Python dependency, it seems quite light.

I don't fully understand the problems with dataclasses, though, so I wanted to start a discussion thread where we can work through these issues. The attrs [Why not Data Classes?](https://www.attrs.org/en/stable/why.html#data-classes) section is a bit light on details (though it may be quite clear for someone who's played with them...).

djarecka:
one issue is with pickling, described by Satra here: https://bugs.python.org/issue35510

djarecka:
the second issue I remember is inheritance - the example below does not work and that it's very annoying:
```
@dataclass
class Parent:
    a: str
    b: int = 3

@dataclass
class Child(Parent):
    c: bool

parent = Parent('string', 32)
child = Child('string', 12, c=True)
```

it doesn't even work if I try to overwrite `a`:
```
@dataclass
class Parent:
    a: str
    b: int = 3

@dataclass
class Child(Parent):
    a: str = "ala"
    c: bool

parent = Parent('string', 32)
child = Child('string', 12, c=True)
```

as far as I remember `attrs` has a slightly different way of creating inheritance

effigies:
Was able to do @satra's suggestion without problems using `attr`:

```Python
import attr
import pickle as pk

@attr.s
class A:
    pass

B = attr.make_class('B', [], bases=(A,))

pk.dumps(A)
pk.dumps(B)
```

So agreed that that's an argument for `attrs`.

djarecka:
here is a useful link: https://stackoverflow.com/questions/51575931/class-inheritance-in-python-3-7-dataclasses

`attrs` doesn't solve the problem completely, but it slightly better IMO

PR #174, opened 2020-01-11
Use attrs instead of dataclasses

satra:
## Types of changes

## Summary
Switch to attrs from dataclasses

## Checklist

- [ ] All tests passing

## Acknowledgment


satra:
@djarecka @effigies  - this is very much a WIP to go in parallel with #173. i have only run `test_specs.py` for now. was even able to un-xfail a few tests.

i do think even based on this, that attrs will simplify a lot of the code. in this particular instance i'm doing a very superficial change (keeping the pickling stuff for instance). but overall it seems quite straightforward to move over.

here is what i have used.
```
dc.dataclass --> @attr.s(auto_attribs=True, kw_only=True)
dc.fields(klass) --> attr.fields(klass)
dc.fields(instance) --> attr_fields(instance) #defined this function
dc.field --> attr.ib
dc.make_dataclass --> attr.make_class (with the fields having to be attr.ib)
```

allows better inheritance. everything is a keyword arg, and defaults work (@djarecka - so default_value may need to be removed or altered)

satra:
@djarecka - just a note that `default_value` will likely not be required. i've turned all the inputs/outputs into keyword only classes.  i haven't removed `default_value` yet.

djarecka:
@satra - great! I tried to read this quickly, but I'd need more time to understand all changes, will do it later

satra:
@djarecka - for me all tests pass under engine except 3 in `test_node_task.py`

djarecka:
@satra - i can take a look

satra:
@effigies  and @djarecka - i am opening this PR for discussion before we merge.

i am going to vote yes since attrs appears to have a nice approach to validation as well, which can be globally disabled. so this should allow us to do type validation quite easily.

as far as i could tell, attrs still does not address the pickling issue.

pydra/engine/task.py
```diff
@@ -385,7 +400,13 @@ def __init__(
     @property
     def cmdline(self):
         """Get the actual command line that will be submitted."""
-        return " ".join(self.container_args + self.command_args)
+        orig_inputs = attr.asdict(self.inputs)
+        modified_inputs = template_update(self.inputs)
+        if modified_inputs is not None:
+            self.inputs = attr.evolve(self.inputs, **modified_inputs)
+        cmdline = " ".join(self.container_args + self.command_args)
+        self.inputs = attr.evolve(self.inputs, **orig_inputs)
```
djarecka: was this needed to pass the existing tests?

djarecka:
sure, let's change it!

pydra/engine/task.py
satra: this was updated to reflect the fact that one can call `cmdline`  before calling run. and this property is not used for running.

but to answer your original question, i don't think there was a test that tested this.

effigies:
I'm okay with changing it. Also, do we have a precise description of what the pickling problem is?

satra:
> do we have a precise description of what the pickling problem is?

- https://bugs.python.org/issue31209
- https://github.com/cloudpipe/cloudpickle/issues/223

concurrent futures uses regular pickle, so cloudpickle cannot be used.

satra:
> I'm okay with changing it.

i'll let @djarecka  hit the merge button once she is ok with it. i'll make a separate dataclasses branch from master in case we wan't to easily find it.

README.md
```diff
@@ -10,7 +10,7 @@ A simple dataflow engine with scalable semantics.
 The goal of pydra is to provide a lightweight Python dataflow engine for DAG construction, manipulation, and distributed execution.
 
 Feature list:
-1. Python 3.7+ using type annotation and dataclasses
+1. Python 3.7+ using type annotation and [attrs](https://www.attrs.org/en/stable/)
```
effigies: Do we still use 3.7+ features, or can we relax to 3.6? (I know that at this point 3.7 isn't cutting edge, so maybe it doesn't matter.)

satra: my current intent was to get this merged, then do additional trimming/testing/relaxing. i did mention the 3.6 expansion to dorota when we met.

pydra/engine/task.py
djarecka: I'm trying to find situation when your modification changes anything. If you think about a single task and running `cmdline` before `run`, this worked without your changes. If you think about workflows, this doesn't work, there are still `LF`.  Could you give me some examples or revert the changes for now.

djarecka: if you prefer to keep it, I'll just make a note to think about it later, but couldn't find a good example

satra: you will also have to take the changes away from `run`. to test the issue, take the current master branch, print `cmdline` from a container task, run it, and then print `cmdline` again.

djarecka:
It run this is needed... i only argu that nothing changes in this place

pydra/engine/task.py
satra: in master this modified test will fail. (i'll push it to this branch)

```python
def test_docker_inputspec_3(plugin, tmpdir):
    """ input file is in the container, so metadata["container_path"]: True,
        the input will be treated as a str """
    filename = "/proc/1/cgroup"

    cmd = "cat"

    my_input_spec = SpecInfo(
        name="Input",
        fields=[
            (
                "file",
                File,
                dc.field(
                    metadata={
                        "mandatory": True,
                        "position": 1,
                        "help_string": "input file",
                        "container_path": True,
                    }
                ),
            )
        ],
        bases=(DockerSpec,),
    )

    docky = DockerTask(
        name="docky",
        image="busybox",
        executable=cmd,
        file=filename,
        input_spec=my_input_spec,
        strip=True,
    )

    cmdline = docky.cmdline # in master this will fail
    res = docky()
    assert "docker" in res.output.stdout
    assert cmdline == docky.cmdline
```

djarecka: I really don't see a difference between the old `cmdline` and the new one: you can see: https://github.com/satra/pydra/pull/12

PR #183, opened 2020-01-25
[wip] Enh/container flag

djarecka:
## Types of changes

## Summary

- adding an optional argument to `ShellCommandTask` - `container`, that can be "docker" or "singularity",if flag is used, than it will run the task in the containers; 
- many methods were moved from `ConatinerTask` to `ShellCommandTask`

## Checklist

- [ ] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/specs.py
+++ b/pydra/engine/specs.py
@@ -406,6 +406,8 @@ class ShellOutSpec(BaseSpec):
 class ContainerSpec(ShellSpec):
     """Refine the generic command-line specification to container execution."""
 
+    image: ty.Union[File, str] = attr.ib(metadata={"help_string": "image"})
+    """The image to be containerized."""
     container: ty.Union[File, str, None] = attr.ib(
         metadata={"help_string": "container"}
     )
--- a/pydra/engine/specs.py
+++ b/pydra/engine/specs.py
@@ -165,6 +165,18 @@ class ShellSpec(BaseSpec):
             "e.g. <file_name>, or a list"
         },
     )
+    image: ty.Union[File, str, None] = attr.ib(None, metadata={"help_string": "image"})
+    """The image to be containerized."""
+    bindings: ty.Optional[
+        ty.List[
+            ty.Tuple[
+                Path,  # local path
+                Path,  # container path
+                ty.Optional[str],  # mount mode
+            ]
+        ]
+    ] = attr.ib(default=None, metadata={"help_string": "bindings"})
+    """Mount points to be bound into the container."""
 
     def retrieve_values(self, wf, state_index=None):
         """Parse output results."""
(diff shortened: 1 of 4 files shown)


djarecka:
@satra - please take a look when you find time. I'm not sure if this is the best approach. I had to move some methods to `ShellCommandTask`, because using the `container` flag with `ShellCommandTask` effectively means that we want to call methods from the child class  (and this methods were calling another methods) so I moved them. 

also had to change some properties to methods to call them from another class.

satra:
@djarecka - i'm hesitating to think about `image` and `bindings` as inputs to `ShellTask`

perhaps one option is to think about a container_info input to a shell task which takes container/image/bindings and returns a container task.

```
a = ShellTask(...) # returns shelltask
b = ShellTask(..., container_info = ('singularity', 'busybox.sif', ...)) # returns SingularityTask
c = ShellTask(..., container_info = ('singularity', 'busybox.sif', ...)) # returns DockerTask
```

this could be implemented via a function `to_docker`, `to_singularity` which takes as input a `ShellTask`, and the container_info and returns the corresponding `[Container]Task`, thus minimizing the need to move things into `ShellTask` that i don't think is relevant to the vanilla `ShellTask`. 

perhaps such a function can be first implemented and then add as a method to `ShellTask`.

can you tell me the use case you are trying to address with this PR?

djarecka:
@satra - the main use case for doing this now is that you put me this on the list things that should be done in jan...

not sure if I understand your idea, the `__init__` method of `ShellCommandTask` should return either `DockerTask` or `RealShellCommandTask`?

satra:
> the main use case for doing this now is that you put me this on the list things that should be done in jan...

ah so it's my problem :) so the thing i would like to be able to do is run a workflow say locally on my laptop via dockertasks, and be able to quickly switch to singularity on our cluster. same workflow different container executing things. while this would be slick, if there are easier things to tackle, you should tackle those first !

> `__init__` method of `ShellCommandTask` should return either `DockerTask` or `RealShellCommandTask`?

yes. however, `__init__` cannot return anything so you will have to overwrite `__new__` (here is an example: https://stackoverflow.com/questions/3209233/how-to-replace-an-instance-in-init-with-a-different-object)

while we are on this thought process, let's also consider how we run a `FunctionTask` through a container. as a start let's assume that `pydra` is installed in the container, or that we can install `pydra` in a container on instance startup.

djarecka:
Yes, I completely agree that this would be nice, I was just explaining why I'm doing this right now...

I thought that this will be faster, but than I realized what I'm really trying to do, and couldn't find a reasonable way of doing it. I was thinking about overwriting self, but this seemed to be not very recommended idea, but didn't think about `__new__`...

satra:
@djarecka - i guess this should be closed.

djarecka:
closed with #184

PR #184, opened 2020-01-26
Enh/container flag new

djarecka:
## Types of changes

## Summary
this replaces #183

adding `container` argument to the `ShellCommandTask`, so it can run in a container 

## Checklist

- [x] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -210,21 +210,28 @@ class FunctionTask(TaskBase):
 class ShellCommandTask(TaskBase):
     """Wrap a shell command as a task element."""
 
-    def __new__(cls, container=None, *args, **kwargs):
-        if container == "docker":
-            if "image" not in kwargs:
-                raise Exception(
-                    "when task run in a container, the image has to be specified"
-                )
-            return DockerTask(*args, **kwargs)
-        elif container == "singularity":
-            if "image" not in kwargs:
-                raise Exception(
-                    "when task run in a container, the image has to be specified"
-                )
-            return SingularityTask(*args, **kwargs)
-        else:
+    def __new__(cls, container_info=None, *args, **kwargs):
+        if not container_info:
             return super(ShellCommandTask, cls).__new__(cls)
+
+        if len(container_info) == 3:
+            type, image, bind = container_info
+        elif len(container_info) == 2:
+            type, image, bind = container_info + (None,)
+        else:
+            raise Exception(
+                f"container_info has to have 2 or 3 elements, but {container_info} provided"
+            )
+
+        if type == "docker":
+            return DockerTask(image=image, bindings=bind, *args, **kwargs)
+        elif type == "singularity":
+            return SingularityTask(image=image, bindings=bind, *args, **kwargs)
+        else:
+            raise Exception(
+                f"first element of container_info has to be "
+                f"docker or singularity, but {container_info[0]} provided"
+            )
 
     def __init__(
         self,
@@ -473,6 +480,8 @@ class ContainerTask(ShellCommandTask):
 class DockerTask(ContainerTask):
     """Extend shell command task for containerized execution with the Docker Engine."""
 
+    init = False
+
     def __init__(
         self,
         name,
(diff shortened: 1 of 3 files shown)


djarecka:
@satra - you're right, that with `__new__` is much better, but I failed to use `container_info` as you suggested, had to provide `image` as a separate argument. There has been something important I'm missing about `__new__`...

satra:
@djarecka - perhaps this demo can help

```python
In [10]: class A: 
    ...:     def __new__(cls, *args, **kwargs): 
    ...:         if 'info' in kwargs: 
    ...:             print('Info: {0}'.format(kwargs['info'])) 
    ...:         return super(A, cls).__new__(cls) 
    ...:     def __init__(self, info=None): 
    ...:         pass 
    ...:                                                                                       

In [11]: a = A()                                                                               

In [12]: a = A(info=('a', 'b'))                                                                
Info: ('a', 'b')

In [13]: a = A(info=None)                                                                      
Info: None
```

so in the shell task info could be a dictionary.

djarecka:
@satra - no, this doesn't help. I modified kwargs, to have image etc., but somehow it goes twice to `DockerTask.__init__` and the second time there is no `image` in the `input`.

djarecka:
I was trying something like [this](https://github.com/djarecka/pydra/blob/enh/container_flag_new_tmp/pydra/engine/task.py#L213), but it doesn't work.

You can try to run: pydra/engine/tests/test_dockertask.py::test_docker_1_dockerflag

satra:
> somehow it goes twice to DockerTask.__init__

that's because dockertask is a subclass of shelltask.

```python
In [29]: class A: 
    ...:     def __new__(cls, *args, **kwargs): 
    ...:         if 'info' in kwargs: 
    ...:             print('Info: {0}'.format(kwargs['info'])) 
    ...:             instance =  B(image=kwargs['info']['image']) 
    ...:             print(type(instance)) 
    ...:             return instance 
    ...:         return super(A, cls).__new__(cls) 
    ...:     def __init__(self, info=None): 
    ...:         print('A init called') 
    ...: class B(A): 
    ...:     def __init__(self, image=None, **kwargs): 
    ...:         print('B entry') 
    ...:         self.image = image 
    ...:         print(kwargs) 
    ...:         print('B exit') 
    ...:                                                                                       

In [30]: a = A(info={'image': 'test'})                                                         
Info: {'image': 'test'}
B entry
{}
B exit
<class '__main__.B'>
B entry
{'info': {'image': 'test'}}
B exit
```

as you can see first the instance of B is created which calls init, and then since new is called inside the original A, another init is invoked with the original arguments. you could avoid it by trapping to see if the instance has already been initialized. of course in the proper implementation you would want to pass the full set of args/kwargs (except info) to B

```python
In [31]: class A: 
    ...:     def __new__(cls, *args, **kwargs): 
    ...:         if 'info' in kwargs: 
    ...:             print('Info: {0}'.format(kwargs['info'])) 
    ...:             instance =  B(image=kwargs['info']['image']) 
    ...:             print(type(instance)) 
    ...:             return instance 
    ...:         return super(A, cls).__new__(cls) 
    ...:     def __init__(self, info=None): 
    ...:         print('A init called') 
    ...: class B(A): 
    ...:     init = False 
    ...:     def __init__(self, image=None, **kwargs): 
    ...:         print('B entry') 
    ...:         if not self.init: 
    ...:             self.image = image 
    ...:             print(kwargs) 
    ...:             self.init = True 
    ...:         print('B exit') 
    ...:                                                                                       

In [32]: a = A(info={'image': 'test'})                                                         
Info: {'image': 'test'}
B entry
{}
B exit
<class '__main__.B'>
B entry
B exit
```

djarecka:
@satra i think I tried similar things, but I will give it another try, but not today. thanks

djarecka:
@satra - on sat didn't read your answer carefully, indeed adding init to DockerTask can solve the problem, thanks!

djarecka:
@satra - can i merge it?

pydra/engine/task.py
```diff
@@ -210,6 +210,29 @@ def _run_task(self):
 class ShellCommandTask(TaskBase):
     """Wrap a shell command as a task element."""
 
+    def __new__(cls, container_info=None, *args, **kwargs):
+        if not container_info:
+            return super(ShellCommandTask, cls).__new__(cls)
+
+        if len(container_info) == 3:
+            type, image, bind = container_info
```
satra: perhaps not use `type` which is a python keyword.

pydra/engine/task.py
```diff
@@ -319,21 +342,28 @@ def command_args(self):
     def command_args(self, args: ty.Dict):
         self.inputs = attr.evolve(self.inputs, **args)
 
+    @property
+    def container_args(self):
```
satra: is this still required.

pydra/engine/task.py
```diff
@@ -319,21 +342,28 @@ def command_args(self):
...
+
     @property
     def cmdline(self):
         """Get the actual command line that will be submitted."""
         orig_inputs = attr.asdict(self.inputs)
         modified_inputs = template_update(self.inputs)
         if modified_inputs is not None:
             self.inputs = attr.evolve(self.inputs, **modified_inputs)
-        cmdline = " ".join(self.command_args)
+        cmdline = " ".join(self.container_args + self.command_args)
```
satra: should this and the container cmdline be reverted.

pydra/engine/task.py
djarecka: I added this when I decided to remove `run_task` and `smndline` from `ContainerTask`. They were very similar

djarecka: but i can modify the methods and add an if statement

pydra/engine/task.py
djarecka: not sure if i follow, looks good to me

satra:
lgtm - you can merge once tests pass.

PR #186, opened 2020-01-30
[fix] fixing cmdline (closes #141)

### LINKED ISSUE #141: TODO for Shell Task and ContainerTask

djarecka:
## Types of changes

## Summary

- fixing `cmdline`, `command_args` and `container_args` so they work if the task has a state (return a list if state) 
- adding `is_lazy` to be able to check if inputs has lazy fields - prevents to run `results()`, and `checksum_states` if inputs is not ready

## Checklist

- [x] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -44,7 +44,7 @@ import inspect
 import typing as ty
 from pathlib import Path
 
-from .core import TaskBase
+from .core import TaskBase, is_lazy
 from ..utils.messenger import AuditFlag
 from .specs import (
     File,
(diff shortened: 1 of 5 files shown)


pydra/engine/task.py
```diff
@@ -288,7 +288,20 @@ def __init__(
 
     @property
     def command_args(self):
-        """Get command line arguments."""
+        """Get command line arguments, returns a list if task has a state"""
+        if is_lazy(self.inputs):
+            raise Exception("can't return cmdline, self.inputs has LazyFields")
+        if self.state:
+            command_args_list = []
+            self.state.prepare_states(self.inputs)
```
satra: what happens if you call this multiple times? will it still work?

djarecka: it works. I agree that it's not the best solution, and should be review where `prepare_States` is called

PR #194, opened 2020-01-31
using hash_file for lists and tuples (closes #178)

### LINKED ISSUE #178: list of Files as an input
`BaseSpec.hash` will not work properly for list of `File`s.

djarecka:
## Types of changes

## Summary

calling `hash_file` if type is ty.List[File] or ty.Tuple[File], `hash_file` can handle lists/tuples

## Checklist

- [ ] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/helpers_file.py
+++ b/pydra/engine/helpers_file.py
@@ -113,7 +113,7 @@ def hash_file(afile, chunk_len=8192, crypto=sha256, raise_notfound=True):
     from .helpers import hash_function
 
     # adding option for tasks with splitter over list of files
-    if isinstance(afile, list):
+    if isinstance(afile, (list, tuple)):
         return hash_function([hash_file(el) for el in afile])
 
     if afile is None or isinstance(afile, LazyField) or isinstance(afile, list):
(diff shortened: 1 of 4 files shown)


pydra/engine/helpers_file.py
```diff
@@ -113,7 +113,7 @@ def hash_file(afile, chunk_len=8192, crypto=sha256, raise_notfound=True):
     from .helpers import hash_function
 
     # adding option for tasks with splitter over list of files
-    if isinstance(afile, list):
+    if isinstance(afile, (list, tuple)):
```
satra: how do you know this list or tuple contains files?

satra: nevermind

pydra/engine/specs.py
```diff
@@ -50,6 +50,8 @@ def hash(self):
                 continue
             if field.type == File and "container_path" not in field.metadata:
                 inp_dict[field.name] = hash_file(getattr(self, field.name))
+            elif field.type == ty.List[File] or field.type == ty.Tuple[File]:
```
satra: what if it was a `List[List[File]`?

it's likely that this bit has to be refactored into a recursive function call. also should support dictionaries whose values are `File`.  this is the function from nipype:

https://github.com/nipy/nipype/blob/master/nipype/interfaces/base/specs.py#L273

i don't think some of the elements are necessary with py3, but the general idea is still something to consider here.

djarecka:
created #195 instead

PR #195, opened 2020-02-01
hash for containers (closes #178)

### LINKED ISSUE #178: list of Files as an input

djarecka:
## Types of changes

## Summary

adding function to return value or calculate hash recursively for dict, tuple and lists
(instead of #194) 

## Checklist

- [ ] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/helpers_file.py
+++ b/pydra/engine/helpers_file.py
@@ -514,12 +514,9 @@ def is_local_file(f):
     return f.type is File and "container_path" not in f.metadata
 
 
-def is_existed_file(f):
-    """ checking if object is an existed file: using multiple path classes"""
-    from py import path
-
-    # for windows might need more path options
-    if isinstance(f, (str, Path, path.local)) and Path(f).exists():
-        return True
-    else:
+def is_existing_file(f):
+    """ checking if an object is an existing file"""
+    try:
+        return Path(f).exists()
+    except TypeError:
         return False
--- a/pydra/engine/helpers_file.py
+++ b/pydra/engine/helpers_file.py
@@ -110,11 +110,6 @@ def fname_presuffix(fname, prefix="", suffix="", newpath=None, use_ext=True):
 def hash_file(afile, chunk_len=8192, crypto=sha256, raise_notfound=True):
     """Compute hash of a file using 'crypto' module."""
     from .specs import LazyField
-    from .helpers import hash_function
-
-    # adding option for tasks with splitter over list of files
-    if isinstance(afile, list):
-        return hash_function([hash_file(el) for el in afile])
 
     if afile is None or isinstance(afile, LazyField) or isinstance(afile, list):
         return None
--- a/pydra/engine/helpers_file.py
+++ b/pydra/engine/helpers_file.py
@@ -512,3 +512,14 @@ def is_local_file(f):
     from .specs import File
 
     return f.type is File and "container_path" not in f.metadata
+
+
+def is_existed_file(f):
+    """ checking if object is an existed file: using multiple path classes"""
+    from py import path
+
+    # for windows might need more path options
+    if isinstance(f, (str, Path, path.local)) and Path(f).exists():
+        return True
+    else:
+        return False
--- a/pydra/engine/specs.py
+++ b/pydra/engine/specs.py
@@ -2,6 +2,7 @@
 import attr
 from pathlib import Path
 import typing as ty
+from .helpers_file import hash_file, is_existed_file
 
 
 def attr_fields(x):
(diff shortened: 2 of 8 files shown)


pydra/engine/specs.py
```diff
@@ -48,18 +48,33 @@ def hash(self):
...
             return hash_function((inp_hash, self._graph_checksums))
         else:
             return inp_hash
 
+    def hash_value(self, value, type=None, metadata=None):
+        """calculating hash or returning values recursively"""
+        from .helpers_file import hash_file
+
+        if isinstance(value, (tuple, list)):
+            return [self.hash_value(el, type, metadata) for el in value]
```
satra: shouldn't something be inspecting the type of `el`?

djarecka: it should be check when the `hash_value` is called

djarecka: if `el` is not container anymore, it should go to line 71

djarecka:
i've modified check for the `type is File` before hashing the content: https://github.com/nipype/pydra/pull/195/files#diff-e620dfe0aa8c1cbe3f271d08b820c89fR73

Not sure how to do better, but I want to allow for this kind of spec: https://github.com/nipype/pydra/pull/195/files#diff-d34220e5398ccac7bb1d867bbf76a63bR934

satra:
this may still not be considering different use cases:

1. the type is a nested container (e.g., `List[List[Union[File, int]]`)

ex1: https://github.com/nipy/nipype/blob/8fe05afcd/nipype/interfaces/spm/preprocess.py#L355
ex2: https://github.com/nipy/nipype/blob/8fe05afcd/nipype/interfaces/fsl/model.py#L77

2. the type of a field is a class that contains attributes which have different types including nested types. this doesn't have an example in nipype 1, but is easy to do with `attrs`.

djarecka:
@satra - will check the links later, but in general I believe the nested containers will work in my implementation, but will add a test for this.  
What will not work right now is a list/container that has both files and ints, so might add additional check to `tp is File or "pydra.engine.specs.File" in str(tp)` that checks if `value.exists`

satra:
> What will not work right now is a list/container that has both files and ints,

indeed, you may need to consider both the spec and the value to determine the type. however, this may require some extra work with Optional and Union types. and some specs may become tricky if someone uses it e.g., ty.Union[File, str]. we may want to raise an exception, but i do know there are situations where this happens (e.g., in FSL) and will need special validation.

so let's do some level of nesting for now with this PR, and open an issue that says more complex situations need to be handled.

djarecka:
@satra - please see the examples that i added to test_specs

pydra/engine/tests/test_specs.py
```diff
@@ -172,3 +172,97 @@ def test_file_hash(tmpdir):
...
+    with open(file, "w") as f:
+        f.write("hello")
+
+    input_spec = SpecInfo(
+        name="Inputs", fields=[("in_file", ty.Union[File, int])], bases=(BaseSpec,)
+    )
+    inputs = make_klass(input_spec)
+
+    # checking specific hash value
+    hash1 = inputs(in_file=file).hash
```
satra: perhaps also check when the input is an `int` here and something else in the next test. overall the tests at least cover some basic cases.

djarecka: I added int in the next test:  https://github.com/nipype/pydra/pull/195/files/01d0c1bc084a1d43562b7dc5432539d846730deb#diff-e0f87d47951f1d013292bae836ba2c75R235
but let me add a few more tests

djarecka:
@satra - I'd merge it once the tests are done, let me know if you want more changes. 
I opened an issue #196

pydra/engine/helpers_file.py
```diff
@@ -517,3 +512,14 @@ def is_local_file(f):
     from .specs import File
 
     return f.type is File and "container_path" not in f.metadata
+
+
+def is_existed_file(f):
```
effigies: What is the difference between this and `os.path.exists`?

pydra/engine/specs.py
```diff
@@ -40,26 +41,44 @@ def collect_additional_outputs(self, input_spec, inputs, output_dir):
...
+            inp_dict[field.name] = self.hash_value(
+                value=getattr(self, field.name), tp=field.type, metadata=field.metadata
+            )
         inp_hash = hash_function(inp_dict)
         if hasattr(self, "_graph_checksums"):
             return hash_function((inp_hash, self._graph_checksums))
         else:
             return inp_hash
 
+    def hash_value(self, value, tp=None, metadata=None):
```
effigies: Is this intended to permit subclasses overriding the `hash_value` method? If not, I think it would be simpler for this to be a function.

djarecka: ok, will move it

pydra/engine/helpers_file.py
djarecka: one problem with `os.path.exists` is that gives `True` for `os.path.exists(4)`

effigies: Gotcha. Maybe just:

```Python
try:
    return Path(f).exists()
except TypeError:
    return False
```

Then we don't have to worry about specifically supporting `py.path` or any other `PathLike`?

Issue #179, opened 2020-01-17
Thoughts from attr conversion - TODOs for next release


satra:
- [x] the hashing code currently assumes flat inputs and will not hash collections containing files properly. see #178, but this is more about `Files` in any type of container (tuple, list, dict) and their nested forms.

- [x] `default_value` metadata should be removed.

- [x] tutorials should be updated before a new release

- [ ] our current dependent projects should be checked (e.g., testkraken)

- [x] we should try to ensure that we are not changing inputs when we run a task/workflow. 

- [ ] push code coverage to 90+ %

-~~[ ] the pickle setstates need some TLC for nested objects. (see ResultSpec)~~

-~~[ ] dask scheduler support~~

- [x] determine how we deal with Undefined vs None.

- ~~[ ] improve sphinx doc using this theme: https://pandas-sphinx-theme.readthedocs.io/en/latest/index.html~~

oesteban:
Just a heads-up

>  improve sphinx doc using this theme

I don't think sphinx versioning would work with that theme

effigies:
Is there a reason we don't want to just stick with ReadTheDocs versioned builds?

djarecka:
I don't have experience with documentation - it would be great if you can point me to the one that should work best (easiest to use or whatever metric is relevant) in your opinion

oesteban:
> Is there a reason we don't want to just stick with ReadTheDocs versioned builds?

As of #161 and #162 if we are to "stick with" something, that would be sphinx-versioning since that is the choice I made to build the documentation (https://nipype.github.io/pydra/). Other than that, there are a couple of equally flimsy reasons:

  - For the last 3 years I've been hearing ReadTheDocs is closing shop next month and that never seems to happen.
  - After the work of Ross getting sphinx-versioning to work well it seems useful to stick with that, and have ownership of the whole deployment via GH pages.

I apologize since I attracted all the attention to probably the lowest priority item in @satra's list.

djarecka:
@oesteban - thank you for the answer! And for the documentation of course! And I'd be very happy  sticking with your choice. Also, if I don't see any strong objections I would also keep the standard theme - maybe I'm too conservative, but I like it :)

djarecka:
@satra 

> * [ ]  we should try to ensure that we are not changing inputs when we run a task/workflow.

do you have a specific example with this issue that you can point me to?

> * [ ]  determine how we deal with Undefined vs None.

[...]

PR #198, opened 2020-02-06
[ci] checking that the input doesn't change when the node/wf is run

djarecka:
## Types of changes

## Summary

addresses one point from #179

## Checklist

- [ ] All tests passing

## Acknowledgment


djarecka:
this is actually not needed, I do check the `checksum` before and after running, and this should be enough to test that input also doesn't change

PR #208, opened 2020-02-23
[fix] passing wf.cache_locations and submitter.rerun

djarecka:
## Types of changes

## Summary

- tasks from a workflow didn't see the `cache_locations` from the workflow, now `wf.cache_locations` is added to each task
- `rerun` from `Submitter.__call__` is passed everywhere, so everything is rerun
- adding `task_rerun` to `Task` attribute; 
- for `Workflow` there is one more flag `propagate_rerun`: if set to `False`, the wf is triggered, but tasks are not automatically rerun (only if they have `task_rerun=True`) 

## Checklist

- [ ] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -800,6 +800,8 @@ class Workflow(TaskBase):
                 return result
         # creating connections that were defined after adding tasks to the wf
         for task in self.graph.nodes:
+            if self.task_rerun:
+                task.task_rerun = self.task_rerun
             task.cache_locations = task._cache_locations + self.cache_locations
             self.create_connections(task)
         # TODO add signal handler for processes killed after lock acquisition
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -811,7 +811,7 @@ class Workflow(TaskBase):
             self.hooks.pre_run_task(self)
             try:
                 self.audit.monitor()
-                await self._run_task(submitter)
+                await self._run_task(submitter, rerun=rerun)
                 result.output = self._collect_outputs()
             except Exception as e:
                 record_error(self.output_dir, e)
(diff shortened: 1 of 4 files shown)


djarecka:
@satra - I had to rethink the `rerun` option, interested what do you think about slightly different approach when `rerun` is in `Submitter` and when it is attribute of a `Task`.
(test are failing for slurm, but will work tomorrow)

satra:
I think we need to separate permanent reruns (the task uses some probabilistic state that someone wants to regenerate each time) vs temporary reruns (i want to create a new cache for whatever reason; e.g. i want a local cache of only a subset of Tasks).

so a task attribute should be used only for permanent and run parameter should be used for temporary.

in the case of a workflow, setting the workflow attribute should propagate to all tasks in the graph. (so effectively to all nested workflows/tasks as well). i don't know the exact syntax for rerunning a workflow, but presumably `sub.run(wf, rerun=True)`  could take care of it, as well as `wf.run(rerun=True)`

djarecka:
@satra - I agree, that this make sense, and I'm a assuming that the `rerun` in `Submitter` you still wannna keep for the temporarily solution. But what about situation (e.g. the one with jupyter example), that you want to restart wf, but doesn't mean that you want to rerun every single node. Do we want to include it?

I know we can always provide a "fake" argument to force wf to rerun, so this might be a solution, if we don't want to cover.

satra:
i think the jupyter situation could be addressed with directory hashing or making the directory read a node outside the workflow. 

(ps. i can't remember if lazy inputs take on the type of the input to which they are connected. if not they should).

satra:
you could also add another flag, which says `propagate_rerun=True/False` if true, it will set it to all sub tasks, if false, it will rerun the workflow task, but subtasks would rerun as needed.

djarecka:
@satra - sure we can do this. 
and yes, we should have `Directory` as a special type, will add an issue

see my new changes:
- when wf has `rerun=True`, everything will be rerun: https://github.com/nipype/pydra/pull/208/files#diff-61004b9f88d892cca950a805bfa61526R1992
- when wf doesn't have `rerun`, than even if the task has `rerun=True`, we don't care: 
https://github.com/nipype/pydra/pull/208/files#diff-61004b9f88d892cca950a805bfa61526R2051
- here is one way of forcing to rerun one task, adding a fake input, just to start wf rerun: https://github.com/nipype/pydra/pull/208/files#diff-61004b9f88d892cca950a805bfa61526R2107

djarecka:
@satra -added the `propagate_rerun` flag for wf

djarecka:
@satra - i'd merge it if you don't mind

PR #231, opened 2020-04-09
Adding BoshTask

djarecka:
## Types of changes

## Summary

adding Shell Command Task that uses boutiques descriptor and runs bosh command to execute the task
(see discussion #210)

## Checklist

- [ ] All tests passing

## Acknowledgment


djarecka:
@ValHayot @glatard, @satra - I've started creating a draft of `BoshTask` that uses `bosh` cli, it's not finished, but I've got stuck with a problem that `bosh` can't find the input files: I believe that this is related to the way how boutiques creates a docker command. It seems that it assume that the input file is in the working directory, is that right?
 I was also testing without `pydra` and it looks like the same command can work in the home directory, but doesn't work in any directory (even if I provide a full path to the input file).

glatard:
Hi @djarecka, it looks like the input files aren't mounted in the tool container. By default, `bosh` only mounts the current working directory, which might explain why it worked when executed from your home directory. The way to fix this would be to pass a list of mounts, through option `-v`, for instance: `bosh exec launch -v <image_dirname>:<image_dirname> [other params]` for each input file.

djarecka:
@glatard -oh, ok, thanks! somehow I assumed that everything has to be mounted automatically (as in cwl).

djarecka:
@glatard - could you please tell me what is the easiest way to get information about all the outputs that should be produced for the specific zenodo and and inputs (the invocation file)

djarecka:
@satra - please check the tests for `BoshTask`: https://github.com/nipype/pydra/blob/0073e150f649cdf9ee070bd80420f3f4ab88ae03/pydra/engine/tests/test_boutiques.py

as we discussed yesterday I used only zenodo files to create `input_spec` and `output_spec`, so at the end I have all names/paths of the output that could be created, but user should be asking only about the output that should be created given set of inputs

satra:
@djarecka - looks reasonable for a start - for the second workflow it would be nice to use the shelltask functionality where stdout is read into a variable so you could actually compare the output of the task.

djarecka:
@satra - let me know if this is what you wanted me to add

djarecka:
i'm randomly(?) getting error from `json.decoder.JSONDecodeError` to some boutiques tests, don't really understand why..

djarecka:
any idea why sometimes `json.load` could give an error when reading zenodo files? It complains that it should be `"`, not `'`, but I think the file is fine, and the error only happens ~20% cases on travis. Not able to reproduce it locally. 

I've added [retry bloc](https://github.com/nipype/pydra/pull/231/files#diff-9ec905824ab21430c8b22ddf59d7d9afR75)...

pydra/engine/boutiques.py
```diff
@@ -0,0 +1,221 @@
...
+
+                input_json[f.name] = value
+
+        filename = self.cache_dir / f"{self.name}-{ind}.json"
+        with open(filename, "w") as jsonfile:
+            json.dump(input_json, jsonfile)
+
+        return str(filename)
+
+    def _run_task(self):
```
satra, line 208: is there a reason why this is being overwritten?

pydra/engine/boutiques.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def _command_args_single(self, state_ind, ind=None):
+        """Get command line arguments for a single state"""
+        input_filepath = self._input_file(state_ind=state_ind, ind=ind)
+        cmd_list = (
+            self.inputs.executable + [input_filepath] + self.inputs.args + self.bindings
+        )
+        return cmd_list
+
+    def _input_file(self, state_ind, ind=None):
```
satra, line 184: this and the previous function appears to interact with state. that seems odd to me. we should not be expecting people to understand state to write a task. thus far all tasks have been somewhat independent of state, and we should keep it that way.

pydra/engine/specs.py
```diff
@@ -360,9 +359,23 @@ def _field_metadata(self, fld, inputs, output_dir):
         if "value" in fld.metadata:
             return output_dir / fld.metadata["value"]
         elif "output_file_template" in fld.metadata:
-            return output_dir / fld.metadata["output_file_template"].format(
-                **inputs.__dict__
+            sfx_tmpl = (output_dir / fld.metadata["output_file_template"]).suffixes
```
satra, line 362: how do you detect suffixes. for imaging files or other modalities this may need to be quite custom? for example `nii.gz`, `BRIK/HEAD`, etc.,.

how will pydra generalize this across domains?

pydra/engine/boutiques.py
```diff
@@ -0,0 +1,221 @@
...
+    def _download_spec(self, zenodo):
+        """ usind bosh pull to download the zenodo file"""
+        spec_file = (
+            Path(os.environ["HOME"])
+            / ".cache/boutiques/production"
+            / (zenodo.replace(".", "-") + ".json")
+        )
+        for i in range(3):
+            if not spec_file.exists():
+                sp.run(["bosh", "pull", zenodo])
```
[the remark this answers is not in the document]
djarecka: perhaps i should use urlib

pydra/engine/boutiques.py
```diff
@@ -0,0 +1,221 @@
...
+                break
+            except json.decoder.JSONDecodeError:
+                tries += 1
+                if tries == tries_max:
+                    raise
+
+        if input_spec is None:
+            input_spec = self._prepare_input_spec()
+        self.input_spec = input_spec
+        if output_spec is None:
```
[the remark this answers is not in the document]
djarecka: you're right, haven't even tested it...

pydra/engine/boutiques.py
djarecka: at the beginning had something in mind, but now it's almost the same indeed. allowed only to use stdout to rise an exception - not getting anything in stderr

pydra/engine/boutiques.py
djarecka: don't understand your comment. pydra prepares the files, this is  a private function

pydra/engine/specs.py
djarecka: my main problem was that `input_1` could be `filename`, `filename.nii` or `filename.nii.gz` and output should be always `filename.nii.gz`, and the templaete is `[input_1].nii.gz`...
so I decided to remove all suffixes if template has it's own, happy to hear better suggestions

satra: is that specific to bosh or boutiques that output is always compressed nifti? also does bosh/boutiques enforce nifti as output? i don't think anything in boutiques enforces it.

pydra/engine/boutiques.py
satra: i did not understand why this function is needed and how to convey to another developer that if they were to write the boutiques task they would need this function. what is the general role of this function? and how would another developer (say a CWL developer) know that this is needed. this seems to depend on `state_ind`.

djarecka: this function creates `invocation` json file with inputs that is needed for `bosh exec launch`. If task has splitter and provides two files for example for `infile` input, I will create two `invocation` files. I'm assuming that user doesn't know that `bosh` requires this file, but just provides `infile`, etc.

pydra/engine/specs.py
djarecka: if you see example for [bet](https://zenodo.org/record/1482743/files/fsl_bet.json?download=1) - you will see that "output-files" has "path-template": "[MASK].nii.gz", but `maskfile` (that has "value-key": "[MASK]"), could be either `filename` or `filename.nii.gz`.  For both cases  `output-files` is `filename.nii.gz`

pydra/engine/boutiques.py
satra: shouldn't this be the kind of thing that happens inside `run_task`? i.e. the creation of this file is only required when you have the inputs already set. this is not an input to the task itself but to bosh.

pydra/engine/specs.py
satra: that may be true of bet, but boutiques can support other tools as well. so there needs to be a general way of deciding what the outputs would be.

this is the part where knowing what outputs should be created given the inputs plays a role, but i don't think that's captured in boutiques yet. so as a first pass, you can simply leave inputs/outputs as separate things.

djarecka: I understood that it's not always a case, but I thought this is a good start. Not sure how do you want me to keep it separately. From the zenodo file I have only a template that uses inputs values

pydra/engine/boutiques.py
djarecka: yes, but this is used by `command_args` property that is used in `run_task`

djarecka: should probably use more specific name

satra: i see that `ShellCommandTask` follows a similar pattern. conceptually does `command_args` need to work for tasks with `State`. is it used anywhere in that form?

wouldn't it be easier if these didn't have to think about state for `command_args` and would raise an Exception if called when the Task has state, but would return the appropriate thing when the Task is stateless.

djarecka: can't answer you right now. I didn't see a problem with it, but can rethink this in a different PR

djarecka: and I actually do think that it's nice to have, it helps for debugging

pydra/engine/boutiques.py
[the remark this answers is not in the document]
djarecka: decided to introduce `input/output_spec_names` instead of `input/output_spec`, so you can provide subset of names that should be used, but type, help_string, etc. is taken from the zenodo spec file. Otherwise I wasn't sure how to deal with conflicts in types, help_strings etc.

PR #237, opened 2020-04-19
FIX: test equality of sorted list of hash values

kaczmarj:
## Types of changes

## Summary

The `helpers.hash_value` function does not guarantee order when giving the hashes of files in a directory, so this commit takes that into account by sorting the output of the `hash_value` calls before comparing.

This test was previously failing on my machine:

```
Linux dash 4.19.0-8-amd64 #1 SMP Debian 4.19.98-1 (2020-01-26) x86_64 GNU/Linux
python 3.8.2, pytest 5.4.1
ext4 filesystem
```

## Checklist

- [x] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/helpers_file.py
+++ b/pydra/engine/helpers_file.py
@@ -87,27 +87,60 @@ def hash_file(afile, chunk_len=8192, crypto=sha256, raise_notfound=True):
     return crypto_obj.hexdigest()
 
 
-def hash_dir(dirpath, raise_notfound=True):
+def hash_dir(
+    dirpath, ignore_hidden_files=False, ignore_hidden_dirs=False, raise_notfound=True
+):
+    """Compute hash of directory contents.
+
+    This function computes the hash of every file in directory `dirpath` and then
+    computes the hash of that list of hashes to return a single hash value. The
+    directory is traversed recursively.
+
+    Parameters
+    ----------
+    dirpath : :obj:`str`
+        Path to directory.
+    ignore_hidden_files : :obj:`bool`
+        If `True`, ignore filenames that begin with `.`.
+    ignore_hidden_dirs : :obj:`bool`
+        If `True`, ignore files in directories that begin with `.`.
+    raise_notfound : :obj:`bool`
+        If `True` and `dirpath` does not exist, raise `FileNotFound` exception. If
+        `False` and `dirpath` does not exist, return `None`.
+
+    Returns
+    -------
+    hash : :obj:`str`
+        Hash of the directory contents.
+    """
     from .specs import LazyField
 
     if dirpath is None or isinstance(dirpath, LazyField) or isinstance(dirpath, list):
         return None
     if not Path(dirpath).is_dir():
         if raise_notfound:
-            raise RuntimeError(f"Directory {dirpath} not found.")
+            raise FileNotFoundError(f"Directory {dirpath} not found.")
         return None
 
-    def search_dir(path):
-        path = Path(path)
-        file_list = []
-        for el in path.iterdir():
-            if el.is_file():
-                file_list.append(hash_file(el))
-            else:
-                file_list.append(search_dir(path / el))
-        return file_list
+    file_hashes = []
+    for dpath, dirnames, filenames in os.walk(dirpath):
+        # Sort in-place to guarantee order.
+        dirnames.sort()
+        filenames.sort()
+        dpath = Path(dpath)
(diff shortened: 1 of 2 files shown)


pydra/engine/tests/test_helpers.py
```diff
@@ -135,7 +135,9 @@ def test_hash_value_dir(tmpdir):
     with open(file_2, "w") as f:
         f.write("hi")
 
-    assert hash_value(tmpdir, tp=Directory) == hash_value([file_1, file_2], tp=File)
+    assert sorted(hash_value(tmpdir, tp=Directory)) == sorted(
```
satra, line 138: @kaczmarj - i think it should be the other way around. i.e. the change should be in the function to always return a consistent hash, and the sort could be on the returned hashes, but more importantly a directory hash should be a singleton (i.e a str) not a list. 

so this might be better addressed by fixing the hash function.

kaczmarj: ok. at the moment, `hash_value(tmpdir, tp=Directory)` returns one hash per file in the directory. you would prefer that to be a single string, representing a hash of all the files in the directory, correct?

satra: yes, it should also support nested directories. this could be easily done by generating a list of hashes of a sorted directory tree (via os.walk and perhaps ignoring empty directories and hidden files), and then hashing that list.

djarecka: this function does support nested directory - see https://github.com/nipype/pydra/blob/master/pydra/engine/tests/test_helpers.py#L142

Indeed it doesn't return one value (but I thought that the order is always the same...), the `hash` function from the input spec returns one value.

This can be changed to one value, but not sure if it changes anything, the order has to be assured anyway

djarecka: @kaczmarj - do you want me to finish this?

kaczmarj:
I’ll work on this tonight

pydra/engine/tests/test_helpers.py
kaczmarj: @satra @djarecka - please see my updates. the `hash_dir` function now returns a single hash value and the directory tree is sorted

pydra/engine/helpers_file.py
```diff
@@ -87,27 +87,60 @@ def hash_file(afile, chunk_len=8192, crypto=sha256, raise_notfound=True):
...
-        path = Path(path)
-        file_list = []
-        for el in path.iterdir():
-            if el.is_file():
-                file_list.append(hash_file(el))
-            else:
-                file_list.append(search_dir(path / el))
-        return file_list
+    file_hashes = []
+    for dpath, dirnames, filenames in os.walk(dirpath):
```
djarecka, line 126: I understand that `os.walk` always returns ordered? you didn't have any issue?

pydra/engine/helpers_file.py
```diff
@@ -87,27 +87,60 @@ def hash_file(afile, chunk_len=8192, crypto=sha256, raise_notfound=True):
...
+                continue
+            this_hash = hash_file(dpath / filename)
+            file_hashes.append(this_hash)
+
+    sha = sha256()
+    for h in file_hashes:
+        sha.update(h.encode())
 
-    return search_dir(dirpath)
+    return sha.hexdigest()
```
djarecka, line 143: perhaps you can use `hash_function` from `helpers`, just for consistency, we might modify the function

pydra/engine/tests/test_helpers.py
```diff
@@ -135,25 +137,49 @@ def test_hash_value_dir(tmpdir):
     with open(file_2, "w") as f:
         f.write("hi")
 
-    assert sorted(hash_value(tmpdir, tp=Directory)) == sorted(
-        hash_value([file_1, file_2], tp=File)
-    )
-    assert hash_value(tmpdir, tp=Directory) == helpers_file.hash_dir(tmpdir)
+    test_sha = hashlib.sha256()
```
djarecka, line 140: `hash_function`, see above

djarecka:
@kaczmarj - thanks! I wonder if we should be also take into account the names of subdirectories (relative to the dir). I believe neither my implementation nor your modifications takes this into account.
@satra - any opinion?

djarecka:
open #243 (merged with master and small updates)

kaczmarj:
thank you @djarecka

PR #241, opened 2020-04-30
adding verbose as an optional argument to Task.result

djarecka:
## Types of changes

## Summary

the additional argument allows to return results values together with input fields that correspond to a specific value.
If `True` or `val` - values of input fields is provided, if `ind` - indices of input fields are provided

## Checklist

- [ ] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -532,29 +532,37 @@ class TaskBase:
                 return True
         return False
 
-    def _combined_output(self):
+    def _combined_output(self, verbose=False):
         combined_results = []
         for (gr, ind_l) in self.state.final_combined_ind_mapping.items():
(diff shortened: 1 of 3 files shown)


djarecka:
@nicolocin @satra  - I've added an option to print the values or indices of the inputs together with the results. This is only a simple form that just uses state properties.

It also works for workflow if splitter/combiner is set on a workflow level. If it's on the node level, than wf doesn't have any state. This would have to be changed when results is being saved in files, we can think about it at some point, but not in this pr.

djarecka:
@nicolocin - you are right, the notebooks should be updated, but this is not the only thing that should be updated, I would do it in a separate PR, will work on this this week

Issue #226, opened 2020-04-02
optimize memory usage when splitting


satra:
right now pydra uses up a lot of memory in preparing for submission. we should profile memory usage relative to the number of elements in the splitter. 

we want to minimize the memory footprint over splits. ideally only have copies of inputs as necessary, but not always.

one could also consider optionally caching inputs to disk and reading from disk instead of sending over memory..

djarecka:
`to_job` is mostly to blame. also asyncio method `run_in_executor` leads to significant memory raise, before starting running any task

djarecka:
closed with #246

PR #246, opened 2020-05-08
moving from deepcopy to pickle files in to_job method

djarecka:
## Types of changes

## Summary

this is a part of an effort to decrease memory load (#226)
 
`to_job` creates pickle files for a task (without input) and the input (right now the input is saved together with the original task) instead of creating deep copies. The files are load and the specific input is set before running the task/wf.

## Checklist

- [x] All tests passing

## Acknowledgment


satra:
just some thoughts here:

`deepcopy` in `to_job` may be essential to ensure that tasks inside a `Workflow` are copied instead of referenced since `to_job` applies to all `Tasks`. a shallow copy will not make copies of nested objects. depending when a copy happens, this may require deepcopy, but i haven't though through the entire workflow.

i don't think getting input `ind` should happen in the worker task `load func`. workers should just process a single task that they are given. so architecturally i would like that function not to evaluate inputs corresponding to indices, but just run a specific job.

finally it will be important to understand how workflow splits are being handled as all of that runs on the main thread and not in workers.  so 

```python
wf = Workflow("foo")
wf.add(FunctionTask(...))
wf.splits(...)
```
then `wf` will make copies in the main thread, and any nested workflow with splits will also make copies in the main thread.
 
so this will require creating a concept diagram of how a workflow is executed and identifying ways to minimize memory usage on the main thread. perhaps understanding the aysnc workflow execution is more important first than fixing the memory issue. it will help then to think about how best to handle memory issue.

djarecka:
@satra - you are right that tasks within wf are referenced, and this was the reason why `deepcopy` had to be there, but now I'm saving the task in the pickle file, so the pickling library does it (as we were discussing this on wed).

I'm not asking for the input in the worker part anymore, could you point me to the line?

I will think about your comments, but the first thing I have to fix is the `slurm`, made some changes (not commited), but doesn't work...

pydra/engine/workers.py
```diff
@@ -177,7 +178,13 @@ def run_el(self, runnable, rerun=False, **kwargs):
 
     async def exec_as_coro(self, runnable, rerun=False):
         """Run a task (coroutine wrapper)."""
-        res = await self.loop.run_in_executor(self.pool, runnable._run, rerun)
+        if isinstance(runnable, TaskBase):
+            res = await self.loop.run_in_executor(self.pool, runnable._run, rerun)
+        else:  # it could be tuple that includes pickle files with tasks and inputs
+            ind, task_pkl, input_pkl, task_orig = runnable
+            res = await self.loop.run_in_executor(
+                self.pool, task_orig._load_and_run, ind, task_pkl, input_pkl, rerun
```
satra, line 186: this line uses `ind`

djarecka: but loads input just before running

pydra/engine/core.py
```diff
@@ -504,15 +504,24 @@ def get_input_el(self, ind):
             return None, inputs_dict
 
     def to_job(self, ind):
-        """Run interface one element generated from node_state."""
+        """ Pickling tasks and inputs, so don't have to load the input to teh memory """
```
satra, line 507:
```suggestion
        """Creates a standalone job by pickling to the task and inputs to disk. 
        
        This reduces memory consumption locally as a deepcopy is avoided. In future 
        work this will be optimized between storage vs memory consumption.
        """
```

pydra/engine/core.py
```diff
@@ -504,15 +504,24 @@ def get_input_el(self, ind):
...
-        el.inputs = attr.evolve(el.inputs, **inputs_dict)
-        return el
+        # copy the task and setting input/state to None
+
+        pkl_files = self.cache_dir / "pkl_files"
+        pkl_files.mkdir(exist_ok=True)
+        task_main_path = pkl_files / f"task_main_{self.name}.pklz"
+
+        # the pickle files should be independent on index, so could be saved once only
+        if ind == 0:
```
satra, line 516: is `ind` always 0 in call to this function when a job doesn't have state? as that job also needs to be pickled.

pydra/engine/core.py
satra: Seems like this function can be simply reduced to pickling the task and input without any consideration `ind` and the function that calls this can deal with indices. `ind` appears mostly as a passthrough.

pydra/engine/core.py
```diff
@@ -504,15 +504,24 @@ def get_input_el(self, ind):
...
-        # dj might be needed
-        # el._checksum = None
-        _, inputs_dict = self.get_input_el(ind)
-        el.inputs = attr.evolve(el.inputs, **inputs_dict)
-        return el
+        # copy the task and setting input/state to None
+
+        pkl_files = self.cache_dir / "pkl_files"
+        pkl_files.mkdir(exist_ok=True)
+        task_main_path = pkl_files / f"task_main_{self.name}.pklz"
```
satra, line 513: how will this deal with name clashes across workflow hierarchy? one of the reasons for suggesting doing this at the individual split level instead of the whole job is so that you can reuse the file if it already exists. 

should also use filelock to write the file to avoid clashes.

pydra/engine/core.py
djarecka: it's not, but I've realized that I could simply save it ones

pydra/engine/helpers.py
```diff
@@ -397,7 +397,7 @@ def get_open_loop():
     return loop
 
 
-def create_pyscript(script_path, checksum, rerun=False):
+def create_pyscript(script_path, checksum, rerun=False, ind=None):
```
satra, line 400: any reason why this wasn't removed as per slack conversation?

i.e instead of creating a pyscript, the slurm job simply does:

```
# SBATCH prefix

python -c "from pydra.engine.utils import run_task; run_task("/path/to/pkf_file", index, True)"
```

pydra/engine/core.py
```diff
@@ -503,16 +503,13 @@ def get_input_el(self, ind):
...
-        el = deepcopy(self)
-        el.state = None
-        # dj might be needed
-        # el._checksum = None
-        _, inputs_dict = self.get_input_el(ind)
-        el.inputs = attr.evolve(el.inputs, **inputs_dict)
-        return el
+    def pickle_task(self):
+        """ Pickling the tasks with full inputs"""
+        pkl_files = self.cache_dir / self.checksum / "pkl_files"
```
satra, line 508:
```suggestion
        pkl_files = self.cache_dir /  "pkl_files" / self.checksum
```
so you can wipe out `pkl_files` after the workflow is run or task is completed.

pydra/engine/core.py
```diff
@@ -503,16 +503,13 @@ def get_input_el(self, ind):
...
+        pkl_files.mkdir(exist_ok=True, parents=True)
+        task_main_path = pkl_files / "_task.pklz"
```
satra, line 510:
```suggestion
        pkl_files = self.cache_dir / "pkl_files"
        pkl_files.mkdir(exist_ok=True, parents=True)
        task_main_path = pkl_files / f"{self.name}_{self.checksum}_task.pklz"
```

pydra/engine/helpers.py
```diff
@@ -94,7 +95,7 @@ def load_result(checksum, cache_locations):
     return None
 
 
-def save(task_path: Path, result=None, task=None):
+def save(task_path: Path, result=None, task=None, use_locfile=False):
```
satra, line 98: any particular reason to make this an argument instead of always using it? then you don't need to make the recursive call.

pydra/engine/helpers.py
```diff
@@ -544,3 +508,37 @@ def get_available_cpus():
...
     return os.cpu_count()
+
+
+def load_and_run(task_pkl, ind=None, rerun=False, **kwargs):
+    """
+     loading a task from a pickle file, settings proper input
+     and running the task
+     """
+    task = load_task(task_pkl=task_pkl, ind=ind)
+    task._run(rerun=rerun, **kwargs)
```
satra, line 519: i would put this in a try - except clause, since things could fail for various reasons and then update the result file appropriately whether things crashed or not.

pydra/engine/helpers.py
```diff
@@ -544,3 +508,37 @@ def get_available_cpus():
...
+
+    if not task.result():
+        raise Exception("Something went wrong")
+    return task
```
satra, line 523: shouldn't this return a result pickle file, not task?

pydra/engine/helpers.py
djarecka: i just noticed that in `Task.run` the `SoftFileLock` is already used, but not sure if it's a good enough reason, can probably change

pydra/engine/core.py
djarecka: but would have to also change in `save` and this will work for all calls of `save`

pydra/engine/helpers.py
djarecka: ok

pydra/engine/helpers.py
djarecka: it returns `task` after running it. it's likely taht doesn't have to return anything

pydra/engine/helpers.py
djarecka: which result file you have in mind?

satra: the result file created by the task. in case the task crashes then a result file created by the code that traps that exception.

djarecka: so you want to write an error into `_result.pklz`? not sure if I follow

djarecka: i've added a simple crashfile

pydra/engine/helpers.py
satra: still returning task though, not result. i still think this should return a pointer to a pickled result file. and the result file should have a result object that indicates that the task has crashed.

djarecka: i was just double checking this, returned object is not used anywhere, so I'm removing this for now. it really could be anything here...

djarecka: actually I could put the file name, so could be used when function is used outside worker or submitter

pydra/engine/helpers.py
```diff
@@ -544,3 +515,96 @@ def get_available_cpus():
...
+
+def load_and_run(
+    task_pkl, ind=None, rerun=False, submitter=None, plugin=None, **kwargs
+):
+    """
+     loading a task from a pickle file, settings proper input
+     and running the task
+     """
+    task = load_task(task_pkl=task_pkl, ind=ind)
+    try:
```
satra, line 528: move to capture `load_task` as well.

pydra/engine/helpers.py
```diff
@@ -544,3 +515,96 @@ def get_available_cpus():
...
+        task(rerun=rerun, plugin=plugin, submitter=submitter, **kwargs)
+    except Exception as excinfo:
+        etype, eval, etr = sys.exc_info()
+        traceback = format_exception(etype, eval, etr)
+        crashfile = report_crash(task, traceback=traceback)
+        raise type(excinfo)(
+            str(excinfo.with_traceback(None)),
+            f" full crash report is here: {crashfile}",
+        )
```
satra, line 537: there could be 3 scenarios that could error:
1. load_task
2. preparation before a task runs in the __call__ function of the task
3. the task itself.

a part of 3 saves records an error (see the `_run` function in TaskBase. it also creates a result file with `errored=True`

this except should handle the other scenarios and not overwrite the result file. it should create one with `errored=True` if one doesn't exist and change the existing one if necessary.

djarecka: how to you want to get to point 2??

pydra/engine/helpers.py
satra: i think it's ok to return the result file name. however see the previous comment about improving what load and run traps.

pydra/engine/helpers.py
djarecka: this error will be raised before even task is saved in a pickle file

satra: it's part of `__call__` of TaskBase and all the way till the try-except clause in `_run`. many things can fail along the way till that try in `_run`

satra: >this error will be raised before even task is saved in a pickle file
i understand crashfile is generated later, but this is not about generating the crashfile, it's about trapping all possible errors in the code prior.

djarecka: so if `_result.pklz` exists you don't want me to overwrite? and what should be there if it doesn't exist? do you wanta traceback, or just `Result(output=None, runtime=None, errored=True)`?

satra: > so if _result.pklz exists you don't want me to overwrite?
correct (it also means the exception was raised by `_run `

> what should be there if it doesn't exist? do you wanta traceback, 
yes, you can record_error in the same way run does

> or just Result(output=None, runtime=None, errored=True)?
yes, return just an error.

djarecka:
@satra - let me know if I can merge this when tests pass

PR #258, opened 2020-05-15
[ fix] adding wf._connection to the checksum (closes #253, closes #252)

### LINKED ISSUE #252: only final set output is kept
this works
```python
wf.set_output([("output", wf.fit_clf.lzout.output),
                ("auc", wf.fit_clf.lzout.auc)])
```
but this doesn't
```python
wf.set_output([("output", wf.fit_clf.lzout.output)])
wf.set_output([("auc", wf.fit_clf.lzout.auc)])
```

### LINKED ISSUE #253: when set_output is changed, old workflow output is returned
changing set_output does not change the output of a previously cached workflow.

djarecka:
## Types of changes

## Summary

connection were not included in the checksum, so when workflow had the same input and different values in `set_output`, the cached values were loaded.

updated `set_output`, so it can be run twice in order to update workflow's output

## Checklist

- [x] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -927,18 +927,29 @@ class Workflow(TaskBase):
             TODO
 
         """
+        if self._connections is None:
+            self._connections = []
         if isinstance(connections, tuple) and len(connections) == 2:
-            self._connections = [connections]
+            new_connections = [connections]
         elif isinstance(connections, list) and all(
             [len(el) == 2 for el in connections]
         ):
-            self._connections = connections
+            new_connections = connections
         elif isinstance(connections, dict):
-            self._connections = list(connections.items())
+            new_connections = list(connections.items())
         else:
             raise Exception(
                 "Connections can be a 2-elements tuple, a list of these tuples, or dictionary"
             )
+        # checking if a new output name is already in the connections
+        connection_names = [name for name, _ in self._connections]
+        new_names = [name for name, _ in new_connections]
+        if set(connection_names).intersection(new_names):
+            raise Exception(
+                f"output name {set(connection_names).intersection(new_names)} is already set"
+            )
+
+        self._connections += new_connections
         fields = [(name, ty.Any) for name, _ in self._connections]
         self.output_spec = SpecInfo(name="Output", fields=fields, bases=(BaseSpec,))
         logger.info("Added %s to %s", self.output_spec, self)
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -219,34 +219,26 @@ class TaskBase:
 
     @property
     def checksum(self):
-        """Calculate a unique checksum of this task."""
-        # if checksum is called before run the _graph_checksums is not ready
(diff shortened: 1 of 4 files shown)


pydra/engine/core.py
```diff
@@ -225,13 +225,19 @@ def checksum(self):
...
 
         input_hash = self.inputs.hash
-        if self.state is None:
+        if self.state is None and not is_workflow(self):
             self._checksum = create_checksum(self.__class__.__name__, input_hash)
         else:
-            # including splitter in the hash
-            splitter_hash = hash_function(self.state.splitter)
+            hash_list = [input_hash]
+            if self.state:
```
satra, line 232: i'm not sure if state should matter? we don't save any working directories for workflows with state right, a workflow with state first gets converted to without state and then run.

djarecka: i don't have a good answer... was also wondering why i have this if statement here, and not in `checksum_sates` and even created #259 to not forget that i have to review these two functions. 
let me just do it within this pr, I will review the issue, I know that removing this part leads to at least one failure in our tests

djarecka: it has to be, since you can have two workflows that have the same inputs but different splitters  (e.g., scalar vs outer)

satra: but a workflow with a state is never run or cached. however, i think i know why this is needed, to compute the checksum of a nested workflow with state. so yes to compute the workflow graph checksum this is required.

djarecka: correct, this is used for checking the nodes within wf, not to create directory

djarecka:
@satra @nicolocin - any idea why windows test can fail due to longer time of runs. The time still shows that it doesn't try to rerun (t2<3), but it's really long comparing what i have on osx or linux
https://dev.azure.com/nipype/pydra/_build/results?buildId=46&view=logs&j=556971f2-d2d8-5a4a-af89-e8ac3ff70591&t=e8932013-9dbf-5aa6-729e-77ab8aa0caa2&l=1306

satra:
no ideas - but it seems that instead of comparing absolute values you could compare relative values, i.e. `original time/cached time > 1` this would also be better across different environments and cpu/library speeds.

djarecka:
@satra - i could possible add this kind of check, but would try to keep the absolute check as well, just perhaps not as strict. The reason why I like the absolute check is that I know that I ask python to wait for 3s.

satra:
the absolute check requires that you can guarantee that the start up time to get to the "sleep" function is minimal (i.e. in ms or less). but it seems from this particular error that this is not the case. perhaps turn on audit mode to see the duration of the task.

djarecka:
the  intriguing part  is that for some reason it happens only for my new tests, but I had the same asserts in other tests, and there were/are fine

djarecka:
ok, I've found a possible explanation - these are the only tests that should not rerun tasks, but have to build workflow, create the directory etc., so apparently these parts takes much londer on windows

satra:
looks good - merging.

Issue #250, opened 2020-05-13
don't allow names of functions to match attributes/names of worklow/task classes


satra:
for example this task will cause an issue downstream when trying to get outputs as `wf.split.lzout ...` since `split` is an actual function

```python
    wf.add(gen_splits(name="split",
                      n_splits=wf.lzin.n_splits,
                      test_size=wf.lzin.test_size,
                      X=wf.readcsv.lzout.X, Y=wf.readcsv.lzout.Y,
                      groups=wf.readcsv.lzout.groups))
```

nicolocin:
Close #274

PR #274, opened 2020-06-14
Raise error when task or workflow name conflicts with names of attributes/methods/other tasks added to workflow

nicolocin:
## Types of changes

## Summary

Fix for #250 

## Checklist

- [x] All tests passing
- [] I have updated documentation (if necessary)

## Acknowledgment

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -34,6 +34,11 @@
       "affiliation": "MIT, HMS",
       "name": "Ghosh, Satrajit",
       "orcid": "0000-0002-5312-6729"
+    },
+    {
+      "affiliation": "MIT",
+      "name": "Lo, Nicol",
+      "orcid": "0000-0002-7522-686X"
     }
   ],
   "keywords": [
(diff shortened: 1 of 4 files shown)


djarecka:
@satra - I would say, that even if we want to add the same function to a workflow it should have a different name

satra:
>  if we want to add the same function to a workflow it should have a different name

that's what this PR would achieve, but i think the overwriting of prior task values would still hold. i.e. adding the same task with a new name can change the settings of a previously added task.

satra:
let me check something about init and TaskBase

satra:
nevermind, when initializing a task it creates new objects, so this will not be an issue:

```
In [6]: fftt = task(annotate({"return": {"out": ty.Any}})(np.fft.fft)) 

In [11]: t1 = fftt(name="1", a=[1,4])                                                     
In [13]: t2 = fftt(name="2", a=[1,1])                                                     
In [19]: t1.inputs                                                                        
Out[19]: Inputs(a=[1, 4], n=None, axis=-1, norm=None, _func=b'\x80\x04\x95\x15\x00\x00\x00\x00\x00\x00\x00\x8c\tnumpy.fft\x94\x8c\x03fft\x94\x93\x94.')
```

Issue #279, opened 2020-06-14
Using ShellCommandTask for DenoiseImage


hjmjohnson:
Given a program with a slightly more complicated user interface, how would one create a `ShellCommandTask` to handle flags?

The simple examples only show how to handle positional arguments. 

For example:
```
COMMAND: 
     DenoiseImage
          Denoise an image using a spatially adaptive filter originally described in J. V. 
          Manjon, P. Coupe, Luis Marti-Bonmati, D. L. Collins, and M. Robles. Adaptive 
          Non-Local Means Denoising of MR Images With Spatially Varying Noise Levels, 
          Journal of Magnetic Resonance Imaging, 31:192-203, June 2010. 

OPTIONS: 
     -d, --image-dimensionality 2/3/4
          This option forces the image to be treated as a specified-dimensional image. If 
          not specified, the program tries to infer the dimensionality from the input 
          image. 

     -i, --input-image inputImageFilename
          A scalar image is expected as input for noise correction. 

     -n, --noise-model Rician/(Gaussian)
          Employ a Rician or Gaussian noise model. 

     -x, --mask-image maskImageFilename
          If a mask image is specified, denoising is only performed in the mask region. 

     -s, --shrink-factor (1)/2/3/...
          Running noise correction on large images can be time consuming. To lessen 
          computation time, the input image can be resampled. The shrink factor, specified 
          as a single integer, describes this resampling. Shrink factor = 1 is the 
          default. 

     -p, --patch-radius 1
                        1x1x1
          Patch radius. Default = 1x1x1 

     -r, --search-radius 2
                         2x2x2
          Search radius. Default = 2x2x2. 

     -o, --output correctedImage
                  [correctedImage,<noiseImage>]
          The output consists of the noise corrected version of the input image. 
          Optionally, one can also output the estimated noise image. 

     --version 
          Get Version Information. 

     -v, --verbose (0)/1
          Verbose output. 

     -h 
          Print the help menu (short version). 

     --help 
          Print the help menu. 
          <VALUES>: 1

```

============

I have 1000 images to denoise, 

```
inputs_dir = '/data/noisy'
outputs_dir ='/data/denoised'
input_files = glob.glob(f"{inputs_dir}/*.nii.gz")

## Pydra stuff


## Some data sink to put the resulting images in the output directory with the input names repaced with new extensions '_denoised.nii.gz'

```

hjmjohnson:
Will investigate at a later time.

PR #289, opened 2020-06-18
extending command_args to include more flexible argstr, and sep

djarecka:
## Types of changes

## Summary

including `argstr` and `sep` supported by nipype 

## Checklist

- [ ] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -461,6 +461,8 @@ class ShellCommandTask(TaskBase):
         """
         if is_lazy(self.inputs):
             raise Exception("can't return cmdline, self.inputs has LazyFields")
+        # checking the inputs fields before returning the command line
+        self.inputs.check_fields_input_spec()
         orig_inputs = attr.asdict(self.inputs)
         modified_inputs = template_update(self.inputs)
         if modified_inputs is not None:
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -448,9 +448,14 @@ class ShellCommandTask(TaskBase):
                 if argstr.endswith("..."):
                     argstr = argstr.replace("...", "")
                     if "sep" in field.metadata:
-                        cmd_el_str = field.metadata["sep"].join(
-                            [argstr.format(**{field.name: val}) for val in value]
-                        )
+                        if "{" + field.name + "}" in argstr:
+                            cmd_el_str = field.metadata["sep"].join(
+                                [argstr.format(**{field.name: val}) for val in value]
+                            )
+                        else:
+                            cmd_el_str = field.metadata["sep"].join(
+                                [f"{argstr} {val}" for val in value]
+                            )
                     else:
                         raise Exception("should we have ... without sep??")
 
@@ -460,7 +465,10 @@ class ShellCommandTask(TaskBase):
                             [str(val) for val in value]
                         )
                     else:
-                        cmd_el_str = argstr.format(**{field.name: value})
+                        if "{" + field.name + "}" in argstr:
+                            cmd_el_str = argstr.format(**{field.name: value})
+                        else:
+                            cmd_el_str = f"{argstr} {value}"
                 cmd_add += cmd_el_str.split(" ")
             # TODO: argstr in nipype1 is always required, should change
             else:
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -315,7 +315,15 @@ class ShellCommandTask(TaskBase):
         """Get command line arguments for a single state"""
         pos_args = []  # list for (position, command arg)
         for f in attr_fields(self.inputs):
-            if f.name == "executable":
+            # these inputs will eb used in container_args
+            if isinstance(self, ContainerTask) and f.name in [
+                "container",
+                "image",
+                "container_xargs",
+                "bindings",
+            ]:
+                continue
+            elif f.name == "executable":
                 pos = 0  # executable should be the first el. of the command
             elif f.name == "args":
                 pos = -1  # assuming that args is the last el. of the command
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -418,6 +418,10 @@ class ShellCommandTask(TaskBase):
         Creating a list with additional parts of the command that comes from
         the specific field.
         """
+        argstr = field.metadata.get("argstr", None)
+        if argstr is None and "output_file_template" not in field.metadata:
+            raise Exception(f"{field.name} doesn't have argstr field in the metadata")
+
         pos = field.metadata.get("position", None)
         if pos is None:
             if "output_file_template" in field.metadata:
@@ -437,46 +441,32 @@ class ShellCommandTask(TaskBase):
 
         cmd_add = []
         if field.type is bool:
-            if "argstr" in field.metadata:
-                if value is True:
-                    cmd_add.append(field.metadata["argstr"])
-            else:
-                raise Exception("if f.type is bool argst should be provided")
+            if value is True:
+                cmd_add.append(argstr)
         else:
-            if "argstr" in field.metadata:
-                argstr = field.metadata["argstr"]
-                if argstr.endswith("..."):
-                    argstr = argstr.replace("...", "")
-                    if "sep" in field.metadata:
-                        if "{" + field.name + "}" in argstr:
-                            cmd_el_str = field.metadata["sep"].join(
-                                [argstr.format(**{field.name: val}) for val in value]
-                            )
-                        else:
-                            cmd_el_str = field.metadata["sep"].join(
-                                [f"{argstr} {val}" for val in value]
-                            )
-                    else:
-                        raise Exception("should we have ... without sep??")
-
-                else:
-                    if "sep" in field.metadata and isinstance(value, list):
+            if argstr.endswith("..."):
+                argstr = argstr.replace("...", "")
+                if "sep" in field.metadata:
+                    if "{" + field.name + "}" in argstr:
                         cmd_el_str = field.metadata["sep"].join(
-                            [str(val) for val in value]
+                            [argstr.format(**{field.name: val}) for val in value]
                         )
                     else:
-                        if "{" + field.name + "}" in argstr:
-                            cmd_el_str = argstr.format(**{field.name: value})
-                        else:
-                            cmd_el_str = f"{argstr} {value}"
-                cmd_add += cmd_el_str.split(" ")
-            # TODO: argstr in nipype1 is always required, should change
+                        cmd_el_str = field.metadata["sep"].join(
+                            [f"{argstr} {val}" for val in value]
+                        )
+                else:
+                    raise Exception("should we have ... without sep??")
             else:
                 if "sep" in field.metadata and isinstance(value, list):
                     cmd_el_str = field.metadata["sep"].join([str(val) for val in value])
-                    cmd_add += cmd_el_str.split(" ")
                 else:
-                    cmd_add += ensure_list(value, tuple2list=True)
+                    if "{" + field.name + "}" in argstr:
+                        cmd_el_str = argstr.format(**{field.name: value})
+                    else:
+                        cmd_el_str = f"{argstr} {value}"
+            cmd_el_str = cmd_el_str.strip().replace("  ", " ")
+            cmd_add += cmd_el_str.split(" ")
         return pos, cmd_add
 
     @property
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -445,13 +445,17 @@ class ShellCommandTask(TaskBase):
                 if isinstance(value, list):
                     cmd_el_str = sep.join([str(val) for val in value])
                     value = cmd_el_str
-                if True:
-                    if "{" + field.name + "}" in argstr:
-                        cmd_el_str = argstr.format(**{field.name: value})
-                    else:
+
+                if "{" + field.name + "}" in argstr:
+                    cmd_el_str = argstr.format(**{field.name: value})
+                else:
+                    if value:
                         cmd_el_str = f"{argstr} {value}"
+                    else:
+                        cmd_el_str = ""
             cmd_el_str = cmd_el_str.strip().replace("  ", " ")
-            cmd_add += cmd_el_str.split(" ")
+            if cmd_el_str:
+                cmd_add += cmd_el_str.split(" ")
         return pos, cmd_add
 
     @property
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -403,16 +403,19 @@ class ShellCommandTask(TaskBase):
         the specific field.
         """
         argstr = field.metadata.get("argstr", None)
-        if argstr is None and "output_file_template" not in field.metadata:
-            raise Exception(f"{field.name} doesn't have argstr field in the metadata")
-
+        if argstr is None:
+            if "output_file_template" in field.metadata:
+                # assuming that input that has output_file_template and no arstr
+                # are not used in the command
+                return None
+            else:
+                raise Exception(
+                    f"{field.name} doesn't have argstr field in the metadata"
+                )
         pos = field.metadata.get("position", None)
         if pos is None:
-            if "output_file_template" in field.metadata:
-                # assuming that input that has output_file_template and no position
-                # are not used in the command
-                # for others, pos will be calculated at the end
-                return None
+            # position will be set at the end
+            pass
         elif not isinstance(pos, int):
             raise Exception(f"position should be an integer, but {pos} given")
         elif pos == 0:
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -314,6 +314,7 @@ class ShellCommandTask(TaskBase):
     def _command_args_single(self, state_ind, ind=None):
         """Get command line arguments for a single state"""
         pos_args = []  # list for (position, command arg)
+        positions_provided = []
         for f in attr_fields(self.inputs):
             # these inputs will eb used in container_args
             if isinstance(self, ContainerTask) and f.name in [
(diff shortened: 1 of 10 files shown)


pydra/engine/task.py
```diff
@@ -348,10 +346,37 @@ def _command_args_single(self, state_ind, ind=None):
                 cpath = cdir.joinpath(lpath.name)
                 value = str(cpath)
             if f.type is bool:
+                if "argstr" in f.metadata:
+                    cmd_add.append(f.metadata["argstr"])
```
satra, line 350: shouldn't this happen only if value is True

pydra/engine/task.py
```diff
@@ -348,10 +346,37 @@ def _command_args_single(self, state_ind, ind=None):
...
                 if value is not True:
                     break
             else:
-                cmd_add += ensure_list(value, tuple2list=True)
+                if "argstr" in f.metadata:
+                    argstr = f.metadata["argstr"]
+
+                    if argstr.endswith("..."):
+                        argstr = argstr.replace("...", "")
+                        if "sep" in f.metadata and isinstance(value, list):
```
satra, line 359: this is the kind of place where InputMultiObject would be used and therefore value should always be a list.

pydra/engine/task.py
```diff
@@ -348,10 +346,37 @@ def _command_args_single(self, state_ind, ind=None):
...
+
+                    else:
+                        if "sep" in f.metadata and isinstance(value, list):
+                            cmd_el_str = f.metadata["sep"].join(
+                                [str(val) for val in value]
+                            )
+                        else:
+                            cmd_el_str = argstr.format(**{f.name: value})
+                    cmd_add += cmd_el_str.split(" ")
+                else:
```
satra, line 374: does this mean argstr is not mandatory?

pydra/engine/task.py
djarecka: but I should be convert to list only if type is ` InputMultiObject`, or just convert everyone?

pydra/engine/task.py
djarecka: it is not

pydra/engine/task.py
djarecka: should change, will be cleaner

pydra/engine/task.py
satra: in nipype1 if argstr is absent it won't get added to the command line. this allows including other inputs that may be required for building the command but will not be used to generate the command line. (e.g. terminal_output).

pydra/engine/task.py
satra: if input type is InputMultiObject in nipype1 this value is always a list, so you don't have to check. this is the kind of check that can happen during annotation. when the instance sees an input with argstr.endswith('...') it can check if the input type is a list or an inputmultiobject.

pydra/engine/task.py
djarecka: ok, I'll make a not. I will rewrite it after hackathon, so can think about it. right now in my examples `argstr` is never required

djarecka:
I know this PR is a bit messy, I will try to clean the code tomorrow, but if test pass, I'll just merge it for now, so we can use converters for sem

djarecka:
@satra - please let me know if you have other comments

pydra/engine/tests/test_dockertask.py
```diff
@@ -757,15 +758,24 @@ def test_docker_inputspec_2(plugin, tmpdir):
             (
                 "file1",
                 attr.ib(
-                    type=File, metadata={"position": 1, "help_string": "input file 1"}
+                    type=File,
+                    metadata={
+                        "position": 1,
+                        "argstr": "",
```
satra, line 764: what does this do? is this equivalent to `"%s"` in nipype 1.   

you can imagine that position and/or argstr as necessary. 
- if position is given without argstr, it can default to `argstr = "{variablename}"`
- if no position and no argstr it doesn't go on the command line
- if no position then argstr must be `argstr = "{variablename}"` to be used. i think this makes it explicit.

satra:
shouldn't there be a set of tests to test the various permutations of command line generation keywords (argstr, sep, other variables) in terms of the command line generated, without necessarily executing the command?

and can the output form in #279 be generated and tested?

djarecka:
@satra - will add more tests

satra:
@djarecka - what would be useful is to take this opportunity to write the docs for the metadata requirements for fields in nipype. perhaps add a page in the docs for now.

djarecka:
@satra - i'm fixing some issues. more tests, more issues...

djarecka:
@satra - I've added tests and included the example: https://github.com/nipype/pydra/pull/289/files#diff-da79acdbbe28e33c4bcb432a61e00458R513

djarecka:
@satra - I'm not sure where should I add it to the docs now, but I was planning to have a special notebook for it

satra:
you could add a new rst page here: https://github.com/nipype/pydra/tree/master/docs

pydra/engine/tests/test_dockertask.py
djarecka: `argstr` can be `-f`, `-f...`, `-f {variable}`, or `-f {variable}...`, if nothing is provided, than only `variable` is used, so yes you can treat  `argstr = "{variablename}"`

I don't treat `position` and `argstr` together. There are completely independent for me. Position can be provided or not. If not provided than I assigned the position.

pydra/engine/tests/test_shelltask_inputspec.py
```diff
@@ -537,6 +537,217 @@ def test_shell_cmd_inputs_mandatory_1():
...
+                        "output_file_template": "{inpB}_out",
+                    },
+                ),
+            ),
+        ],
+        bases=(ShellSpec,),
+    )
+
+    shelly = ShellCommandTask(executable="executable", input_spec=my_input_spec)
+    # outA has position, so is a part of the command line
```
satra, line 648: only `argstr` should indicate whether should indicate if something is added to the command line not `position`

pydra/engine/tests/test_shelltask_inputspec.py
```diff
@@ -537,6 +537,217 @@ def test_shell_cmd_inputs_mandatory_1():
...
+                ),
+            ),
+        ],
+        bases=(ShellSpec,),
+    )
+
+    shelly = ShellCommandTask(
+        executable="executable", input_spec=my_input_spec, inpA="inpA"
+    )
+    # outA has no position, so is not a part of the command line
```
satra, line 615: only `argst`r should indicate whether should indicate if something is added to the command line not `position`

pydra/engine/tests/test_shelltask_inputspec.py
satra, line 648: this comment is a bit weird for this use case, since it actually doesn't output anything given lack of input.

pydra/engine/tests/test_shelltask_inputspec.py
```diff
@@ -705,6 +921,8 @@ def test_shell_cmd_inputs_di(tmpdir):
 
     my_input_file = tmpdir.join("a_file.txt")
     my_input_file.write("content")
+    my_mask_file = tmpdir.join("a_mask_file.txt")
```
satra, line 924: mask file is an output of the command not an input, so this should not exist.

pydra/engine/tests/test_shelltask_inputspec.py
```diff
@@ -718,7 +936,22 @@ def test_shell_cmd_inputs_di(tmpdir):
...
+    # input file name and mask file
+    shelly = ShellCommandTask(
+        executable="DenoiseImage",
+        inputImageFilename=my_input_file,
+        maskImageFilename=my_mask_file,
+        input_spec=my_input_spec,
+    )
+    assert (
+        shelly.cmdline
+        == f"DenoiseImage -i {my_input_file} -x {my_mask_file} -s 1 -p 1 -r 2 -o {my_input_file} {my_mask_file}"
```
satra, line 953: output should look like (note the brackets):

`f"DenoiseImage -i {my_input_file} -x {my_mask_file} -s 1 -p 1 -r 2 -o [{my_input_file}, {my_mask_file}]"

pydra/engine/tests/test_shelltask_inputspec.py
djarecka: so you mean `argstr` could be not provided, is that right?

satra: yes - if you don't want it in the command line don't provide argstr (it will still be used for hashing, etc.,.).

djarecka:
@satra - I thought that we agreed that for now we want to have the explicit output generation. If not, I have to come up, with some `if/elif/else` logic and add probably an extra field to `metadata` dictionary.

djarecka:
@satra - not sure what you mean that the logic is off. If you are talking about [this line](https://github.com/nipype/pydra/pull/289/files#diff-c6a774753d90549572ab684cb09d9c45R444), it is needed, since I'm assuming that even if `...` is used, someone can provide a single value

satra:
> for now we want to have the explicit output generation

it's a starting point and may be we don't need the boolean option. i'm fine with not including it for immediate release, but think about it for the next one.

> If you are talking about this line, it is needed, 

it replaces argstr and immediately checks for list, which is equivalent, but i understand the need to check for both in that section.

so for both for the moment it's ok.

djarecka:
@satra - for now i've resolved the issue with no pos vs no argstr. I should have some time later today to work on this, so will try to work on the output options you suggested. If not, I'd just merge it today anyway.

Just clarifying your idea - you want the user to provide `True` to use the template, and `str` to use the string, right?
But the template should still "make decision" if can generate entire template based on the set of the input, right?

For the `DenoiseImage` example I could introduce another fields that says that output  should be provided as a list, sounds reasonable to me to have this special case.

satra:
> Just clarifying your idea - you want the user to provide True to use the template, and str to use the string, right?
But the template should still "make decision" if can generate entire template based on the set of the input, right?

yes

> For the DenoiseImage example I could introduce another fields that says that output should be provided as a list, sounds reasonable to me to have this special case.

there are two things here. 
- one is the command line generated. for this i think the current implementation would work with `argstr="-o [{outfile}]"`
- the second is how you get the outputs to become two fields (outfile, noisefile), and for that i think introducing two inputs fields, one for outfile and one for noisefile (both with no argstr, but both with output_file_template) and then a third input that integrates the two with `argstr="-o [{outfile},{noisefile}]`

i think the above can also be done with just the current implementation without any change.

```python
                "correctedImage",
                attr.ib(
                    type=str,
                    metadata={
                        "help_string": """
                        The output consists of the noise corrected version of the input image.
                        Optionally, one can also output the estimated noise image.
                        """,
                        "output_file_template": "{inputImageFilename}_out",
                    },
                ),
            ),
              "noiseImage",
                attr.ib(
                    type=str,
                    metadata={
                        "help_string": """
                        The output consists of the noise corrected version of the input image.
                        Optionally, one can also output the estimated noise image.
                        """,
                        "output_file_template": "{inputImageFilename}_noise",
                    },
                ),
            ),
             "output",
                attr.ib(
                    type=str,
                    metadata={
                        "help_string": """
                        The output consists of the noise corrected version of the input image.
                        Optionally, one can also output the estimated noise image.
                        """,
                        "argstr": "-o [{correctedImage},{noiseImage}]",
                        "position": -1,
                        "mandatory": True,
                        "readonly": True,
                    },
                ),
            ),
```
the ideal thing would be that the type for output is something that doesn't list it in the inputs. i've used the metadata `readonly` (does not exist) for this.

djarecka:
@satra - this will not work for now well, since if you do not provide `noiseImage`, you will get `"-o correctedImage, NOTHING]"`, or with [my last changes](https://github.com/nipype/pydra/pull/289/commits/0a94c1c676552c05861f1cc230dbfe5e4663f42a#diff-86bd65c1273fb10b7c5f794cc3e311abR524), you won't get anything... But could be providing perhaps an empty strings for all inputs that are not set by the user

satra:
> this will not work for now well, since if you do not provide noiseImage, you will get `"-o correctedImage, NOTHING]"`

perhaps not right now, but it seems this could be handled by iteratively updating a directed graph of inputs for naming (ensuring there is no circular dependency) and excluding any inputs that are not set. i think this would be a few short functions to make work. 
1. create directed graph of inputs for output file names
2. evaluate the dependencies in a sorted order ignoring attr.NOTHING

djarecka:
@satra - and what about this: https://github.com/nipype/pydra/pull/289/files#diff-da79acdbbe28e33c4bcb432a61e00458R860

satra:
@djarecka - the cmdline property should return the proper command line, not an f-string. so it should be fully formatted at that time. it looks like the right direction.

however, this will only solve the inputspec -> command line generation bit. to generate two separate output fields will require additional changes, more related to the previous comment.

djarecka:
@satra - I'm giving up. I don't understand what you mean that it solves only  command line generation bit. Perhaps we can chat about it during the wed meeting

djarecka:
@satra - ok, I've just got what you mean

djarecka:
> the cmdline property should return the proper command line, not an f-string. so it should be fully formatted at that time. it looks like the right direction.

@satra - btw. the command line is fully formatted

satra:
@djarecka - see line 948 in  pydra/engine/tests/test_shelltask_inputspec.py

that's an f-string, not a fully formatted shell command. i.e. one i can copy and paste into the terminal to execute.

djarecka:
@satra - in the line 948 I'm using f-string in the `assert` statement, but python formats this before comparing (I'm using `f"..."` syntax)

djarecka:
@satra - I have a new version of the DI example. Sorry, not sure how I forgot about the output fields...

Anyway, I still have  a question about the code you suggested - both `correctedImage` and `noiseImage` use in the templates only `inputImageFilename`. So in theory, they could be always created whenever `inputImageFilename` is set. Do you still want user to be able to say `noiseImage=True` or `noiseImage=False`. Should we require this for every single field that has `output_file_template`? Right now, I always create whatever is possible (depends on the other inputs provided by the user, see [test_shell_cmd_inputs_template_1](https://github.com/nipype/pydra/pull/289/files#diff-da79acdbbe28e33c4bcb432a61e00458R540) vs [test_shell_cmd_inputs_template_2](https://github.com/nipype/pydra/pull/289/files#diff-da79acdbbe28e33c4bcb432a61e00458R620))

satra:
> Do you still want user to be able to say noiseImage=True or noiseImage=False.

In general, this option would be nice, since that additional output may require extra computation, which a user may want to avoid in some scenarios. for this particular interface i think is less critical.

i would perhaps use explicit indicator such as `type` being a `Union[str, bool]`, rather than making something by default accept both if `output_file_template` exists.  having the union would also allow setting a `default=False` option for example.

pydra/engine/tests/test_shelltask_inputspec.py
```diff
@@ -0,0 +1,1137 @@
...
+                    },
+                ),
+            ),
+            (
+                "output",
+                attr.ib(
+                    type=str,
+                    metadata={
+                        "help_string": "Combined output",
+                        "argstr": "-o [{correctedImage} {noiseImage}]",
```
satra, line 1049:
```suggestion
                        "argstr": "-o [{correctedImage}, {noiseImage}]",
```
just to be technically correct.

pydra/engine/tests/test_shelltask_inputspec.py
```diff
@@ -0,0 +1,1137 @@
...
+
+    # input file name
+    shelly = ShellCommandTask(
+        executable="DenoiseImage",
+        inputImageFilename=my_input_file,
+        input_spec=my_input_spec,
+    )
+    assert (
+        shelly.cmdline
+        == f"DenoiseImage -i {my_input_file} -s 1 -p 1 -r 2 -o [{my_input_file}_out {my_input_file}_noise]"
```
satra, line 1116:
```suggestion
        == f"DenoiseImage -i {my_input_file} -s 1 -p 1 -r 2 -o [{my_input_file}_out, {my_input_file}_noise]"
```

pydra/engine/tests/test_shelltask_inputspec.py
```diff
@@ -0,0 +1,1137 @@
...
+    # input file name and help_short
+    shelly = ShellCommandTask(
+        executable="DenoiseImage",
+        inputImageFilename=my_input_file,
+        help_short=True,
+        input_spec=my_input_spec,
+    )
+    assert (
+        shelly.cmdline
+        == f"DenoiseImage -i {my_input_file} -s 1 -p 1 -r 2 -h -o [{my_input_file}_out {my_input_file}_noise]"
```
satra, line 1128:
```suggestion
        == f"DenoiseImage -i {my_input_file} -s 1 -p 1 -r 2 -h -o [{my_input_file}_out, {my_input_file}_noise]"
```

pydra/engine/task.py
```diff
@@ -314,56 +314,159 @@ def command_args(self):
...
+        # checking if the position is not already used
+        elif pos in self._positions_provided:
+            raise Exception(
+                f"{field.name} can't have provided position, {pos} is already used"
+            )
+        self._positions_provided.append(pos)
+        value = self._field_value(
+            field=field, state_ind=state_ind, ind=ind, check_file=True
+        )
+        if field.metadata.get("readonly", False) and value is not None:
```
satra, line 427: just a note that this may be the kind of metadata we check for on input assignment rather than just on commandline formation.

djarecka: perhaps it can be a part of input validation, once we build one

satra: at least basic input validation would be a good target for the august release (0.8).

djarecka:
@satra - ok, so fields with `output_file_template` can be now a bool. You can see the new tests.

satra:
nice work @djarecka

Issue #290, opened 2020-06-18
Raise error when result.output is none


nicolocin:
`output` can be `None` when output is not set in `set_output` or if one of the tasks doesn't return anything.

The error currently looks like something like which isn't very descriptive.

```
<ipython-input-1-a0fe8ea3a54b> in run_workflow(wf, plugin, plugin_args, specfile)
    339     cwd = os.getcwd()
    340     with pydra.Submitter(plugin=plugin, **plugin_args) as sub:
--> 341         sub(runnable=wf)
    342     results = wf.result(return_inputs=True)
    343     os.chdir(cwd)

/usr/local/anaconda3/lib/python3.7/site-packages/pydra/engine/submitter.py in __call__(self, runnable, cache_locations, rerun)
     55             ]
     56         if is_workflow(runnable) and runnable.state is None:
---> 57             self.loop.run_until_complete(self.submit_workflow(runnable, rerun=rerun))
     58         else:
     59             self.loop.run_until_complete(self.submit(runnable, wait=True, rerun=rerun))

/usr/local/anaconda3/lib/python3.7/asyncio/base_events.py in run_until_complete(self, future)
    582             raise RuntimeError('Event loop stopped before Future completed.')
    583 
--> 584         return future.result()
    585 
    586     def stop(self):

/usr/local/anaconda3/lib/python3.7/site-packages/pydra/engine/submitter.py in submit_workflow(self, workflow, rerun)
     70                 await self.worker.run_el(workflow, rerun=rerun)
     71             else:
---> 72                 await workflow._run(self, rerun=rerun)
     73         else:  # could be a tuple with paths to pickle files wiith tasks and inputs
     74             ind, wf_main_pkl, wf_orig = workflow

/usr/local/anaconda3/lib/python3.7/site-packages/pydra/engine/core.py in _run(self, submitter, rerun, **kwargs)
    924                 self.hooks.post_run_task(self, result)
    925                 self.audit.finalize_audit(result=result)
--> 926                 save(odir, result=result, task=self)
    927                 os.chdir(cwd)
    928         self.hooks.post_run(self, result)

/usr/local/anaconda3/lib/python3.7/site-packages/pydra/engine/helpers.py in save(task_path, result, task, name_prefix)
    129             if task_path.name.startswith("Workflow"):
    130                 # copy files to the workflow directory
--> 131                 result = copyfile_workflow(wf_path=task_path, result=result)
[...]

PR #292, opened 2020-06-18
raise error when workflow returns None

nicolocin:
## Types of changes

## Summary

Fix for #290.   

Edit: wait a better way is to actually check if `set_output` is used as @djarecka suggested.  Will update 

## Checklist

- [x] All tests passing

## Acknowledgment


satra:
would we not allow a workflow that doesn't generate an output?

djarecka:
@satra - i think that's what we decided that wf has to have input and output

effigies:
If a task can return None, I don't see why a workflow shouldn't be able to.

satra:
> If a task can return None, I don't see why a workflow shouldn't be able to.

yes, that is true, but i think it still needs a named output port for that None, just like the task needs a named output for that None.

effigies:
Ah. Yes, I can accept a requirement for an output port.

djarecka:
yes, `result.output.out` can be `None`, but whenever `Result` is `None`, something is wrong

pydra/engine/core.py
```diff
@@ -882,6 +878,11 @@ def create_connections(self, task):
 
     async def _run(self, submitter=None, rerun=False, **kwargs):
         # self.inputs = dc.replace(self.inputs, **kwargs) don't need it?
+        # output_spec needs to be set using set_output or at workflow initialization
```
satra, line 881: i believe the only place to set an output_spec is via the init, so if any check should happen it should go to init, rather than in _run.

pydra/engine/helpers.py
```diff
@@ -128,10 +128,10 @@ def save(task_path: Path, result=None, task=None, name_prefix=None):
         if result:
             if task_path.name.startswith("Workflow"):
                 # copy files to the workflow directory
-                if result.output is None:
-                    raise ValueError(
-                        "Workflow output cannot be None, define output(s) with set_output"
-                    )
+                # if result.output is None:
```
satra, line 131: remove any code that is not required instead of commenting it out. you can always go back to history to retrieve it.

pydra/engine/task.py
```diff
@@ -70,11 +70,11 @@ def __init__(
         audit_flags: AuditFlag = AuditFlag.NONE,
         cache_dir=None,
         cache_locations=None,
-        input_spec: ty.Optional[SpecInfo] = None,
+        input_spec: ty.Union[SpecInfo, BaseSpec, None] = None,
```
satra, line 73:
```suggestion
        input_spec: ty.Optional[ty.Union[SpecInfo, BaseSpec]] = None,
```

pydra/engine/task.py
```diff
@@ -70,11 +70,11 @@ def __init__(
...
         messenger_args=None,
         messengers=None,
         name=None,
-        output_spec: ty.Optional[BaseSpec] = None,
+        output_spec: ty.Union[SpecInfo, BaseSpec, None] = None,
```
satra, line 77:
```suggestion
        output_spec: ty.Optional[ty.Union[SpecInfo, BaseSpec]] = None,
```

pydra/engine/core.py
nicolocin: output_spec is set at initialization, but is overwritten if you run `set_output` after adding all the tasks.  

```python
def set_output(self, connections):
...
self.output_spec = SpecInfo(name="Output", fields=fields, bases=(BaseSpec,))
logger.info("Added %s to %s", self.output_spec, self)
```

nicolocin: Unless you passed in all the tasks to a workflow at workflow initialization already it doesn't actually make a lot of sense to specify output_spec...?   Let me know if I misunderstood anything

pydra/engine/helpers.py
nicolocin: Sorry forgot to delete!  Will be more careful next time

satra:
we should still check for result.output is not None before calling copyfile. it may be possible that result.output is None because the underlying code crashed for some reason (take a look at the run function in task - near line 402).

```
/usr/local/anaconda3/lib/python3.7/site-packages/pydra/engine/helpers.py in save(task_path, result, task, name_prefix)
    129             if task_path.name.startswith("Workflow"):
    130                 # copy files to the workflow directory
--> 131                 result = copyfile_workflow(wf_path=task_path, result=result)
    132             with (task_path / f"{name_prefix}_result.pklz").open("wb") as fp:
    133                 cp.dump(result, fp)
```

so: `if task_path.name.startswith("Workflow") and result.output is not None:`

satra:
@nicolocin and @djarecka - it would be nice if this issue is fixed - running into too many of these :) please see my comment above though.

PR #301, opened 2020-06-24
Revert "raise error when workflow returns None"

satra:
Reverts nipype/pydra#292

@nicolocin and @djarecka - i'm reverting this change for now. it seems to create a state where complex workflows don't finish. should have looked at more testing before merge.

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -81,7 +81,7 @@ class TaskBase:
         audit_flags: AuditFlag = AuditFlag.NONE,
         cache_dir=None,
         cache_locations=None,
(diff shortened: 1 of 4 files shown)


satra:
@nicolocin - after this is reverted, please re-open the PR, and let's make sure to consider all places where output is used and make sure we test the consequences of that.

satra:
i won't do this revert right away if we can fix this, but please don't merge anything else and release a version of pydra.

nicolocin:
Ahh I see.  Didn't close the issue so it's still here #290

djarecka:
@satra - should we add a specific test? Do you have one?

djarecka:
@satra - could we close it?

PR #303, opened 2020-06-28
better checks if tasks are errored, removing nodes from graph

djarecka:
## Types of changes

## Summary

instead of #302 

## Checklist

- [ ] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -699,7 +699,7 @@ class Workflow(TaskBase):
 
         """
         if input_spec:
-            if isinstance(input_spec, BaseSpec):  # TODO: test this block
+            if isinstance(input_spec, BaseSpec):
                 self.input_spec = input_spec
             else:
                 self.input_spec = SpecInfo(
@@ -754,24 +754,6 @@ class Workflow(TaskBase):
         if name in self.name2obj:
             return self.name2obj[name]
         return self.__getattribute__(name)
-
-    # TODO: not sure if we will need this, if not, should be removed
-    # @property
-    # def done_all_tasks(self):
-    #     """
-    #     Check if all tasks from the graph are done.
-    #
-    #     .. important ::
-    #         The fact that all tasks are reported as done
-    #         doesn't mean that results of the workflow
-    #         are available.
-    #         That can be checked with :py:meth:`~Workflow.done`.
-    #
-    #     """
-    #     for task in self.graph.nodes:
-    #         if not task.done:
-    #             return False
-    #     return True
 
     @property
     def nodes(self):
@@ -1010,7 +992,13 @@ class Workflow(TaskBase):
                 output_wf[name] = val_out
             except ValueError:
                 output_wf[name] = None
-                raise ValueError(f"Task {val.name} raised an error")
+                # checking if the tasks has predecessors that raises error
+                if isinstance(getattr(self, val.name)._errored, list):
+                    raise ValueError(
+                        f"Tasks {getattr(self, val.name)._errored} raised an error"
+                    )
+                else:
+                    raise ValueError(f"Task {val.name} raised an error")
         return attr.evolve(output, **output_wf)
 
 
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -182,7 +182,7 @@ class TaskBase:
 
         self.plugin = None
         self.hooks = TaskHook()
-        self.errored = False
+        self._errored = False
 
     def __str__(self):
         return self.name
@@ -220,6 +220,11 @@ class TaskBase:
     def version(self):
         """Get version of this task structure."""
         return self._version
+
+    @property
+    def errored(self):
+        """Check if the task has raised an error"""
+        return self._errored
 
     @property
     def checksum(self):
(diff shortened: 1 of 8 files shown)


pydra/engine/core.py
```diff
@@ -182,6 +182,7 @@ def __init__(
 
         self.plugin = None
         self.hooks = TaskHook()
+        self.errored = False
```
satra, line 185: should this be a property. i don't think you want a user to set this.

pydra/engine/core.py
```diff
@@ -677,7 +686,7 @@ def __init__(
 
         """
         if input_spec:
-            if isinstance(input_spec, BaseSpec):
+            if isinstance(input_spec, BaseSpec):  # TODO: test this block
```
satra, line 689: is this todo taken care of?

pydra/engine/core.py
```diff
@@ -733,22 +742,23 @@ def __getattr__(self, name):
...
-            doesn't mean that results of the workflow
-            are available.
-            That can be checked with :py:meth:`~Workflow.done`.
-
-        """
-        for task in self.graph.nodes:
-            if not task.done:
-                return False
-        return True
+    # TODO: not sure if we will need this, if not, should be removed
```
satra, line 745: remove if not needed.

pydra/engine/core.py
```diff
@@ -916,7 +926,10 @@ async def _run(self, submitter=None, rerun=False, **kwargs):
             try:
                 self.audit.monitor()
                 await self._run_task(submitter, rerun=rerun)
-                result.output = self._collect_outputs()
+                result.output, err = self._collect_outputs()
```
satra, line 929: why would/should collect_outputs return an error? and not _run_task raise an exception?

satra: or is this specifically being done for workflows? if so can't _collect_outputs also raise an exception? the exception is already being trapped.

pydra/engine/core.py
```diff
@@ -973,15 +988,21 @@ def set_output(self, connections):
...
         # collecting outputs from tasks
         output_wf = {}
         for name, val in self._connections:
             if not isinstance(val, LazyField):
                 raise ValueError("all connections must be lazy")
-            output_wf[name] = val.get_value(self)
-        return attr.evolve(output, **output_wf)
+            val_out = val.get_value(self)
+            if val_out == "error":
+                error = True
```
satra, line 1001: why not just raise a `ValueError` here instead of using an `error` variable?

pydra/engine/graph.py
```diff
@@ -186,7 +186,7 @@ def _sorting(self, notsorted_list, predecessors):
                 remaining_nodes.append(nd)
         return sorted_part, remaining_nodes
 
-    def remove_nodes(self, nodes):
+    def remove_nodes(self, nodes, check_ready=True):
```
satra, line 189: add docstrings to functions that are being modified.

pydra/engine/specs.py
```diff
@@ -121,6 +122,20 @@ def __setstate__(self, state):
             state["output"] = klass(**state["output"])
         self.__dict__.update(state)
 
+    def get_output_field(self, field_name):
+        """
+        used in get_values in Workflow
```
satra, line 127:
```suggestion
        """Used in get_values in Workflow
```

pydra/engine/specs.py
```diff
@@ -121,6 +122,20 @@ def __setstate__(self, state):
...
+
+        Parameters
+        ----------
+        field_name ::`str`
```
satra, line 131:
```suggestion
        field_name : `str`
```

pydra/engine/specs.py
```diff
@@ -492,24 +507,26 @@ def get_value(self, wf, state_index=None):
...
-                        if self.field == "all_":
-                            res_l_new = [attr.asdict(res.output) for res in res_l]
-                        else:
-                            res_l_new = [
-                                getattr(res.output, self.field) for res in res_l
-                            ]
+                        res_l_new = []
+                        for res in res_l:
+                            if res.errored:
+                                return "error"
```
satra, line 513: why not use exceptions, here and elsewhere in this function?

pydra/engine/tests/test_specs.py
```diff
@@ -101,6 +101,10 @@ def __init__(self):
         class Result:
             def __init__(self):
                 self.output = Output()
+                self.errored = False
+
+            def get_output_field(self, field):
+                return getattr(self.output, field)
```
satra, line 107: since this is one line replacing another line, is there a need for this function?

pydra/engine/core.py
djarecka: you're probably right, but I'm using this in the Submitter

pydra/engine/core.py
djarecka: CC: @nicolocin

pydra/engine/core.py
djarecka: I will, just not completely sure if everything will work, but looks like

pydra/engine/core.py
djarecka: yes, I should change

pydra/engine/tests/test_specs.py
djarecka: these are "fake" classes for testing, so need the same methods. I'm not completely sure if I like these classes/tests, but left for now

pydra/engine/core.py
djarecka: oh, you mean I should create a property?

pydra/engine/core.py
djarecka: assuming that @nicolocin is dealing with it, removing todo

djarecka:
@satra - can you please take a look

pydra/engine/core.py
```diff
@@ -534,7 +548,11 @@ def done(self):
                 return True
         else:
             if self.result():
```
satra, line 550: since result is a function, shouldn't it be assigned to a variable so that it is not recomputed?

PR #305, opened 2020-07-01
[rfc]changing Task initialization syntax closes #295 

### LINKED ISSUE #295: RFC: Task initialization syntax
Let's break things!

Right now, when you instantiate a Task, you can give it a `name`:

```Python
@pydra.mark.task
def greet(name : str) -> str:
    return f"Hi, {name}"

greeter = greet(name="greeter")
print(greeter(name="@effigies").output.out)
```

```
Hi, @effigies
```

So this works, but it might be worth considering whether this is going to cause confusion in a workflow context:

```Python
wf = pydra.Workflow(name="myworkflow", input_spec=["name"])
wf.add(greet(name="greeter"))
wf.greeter.inputs.name = wf.lzin.name
wf.set_output([("out", wf.greeger.lzout.out")])
wf.inputs.name = "@effigies"

with pydra.Submitter() as sub: 
    print(sub(wf).output.out)
```

```
Hi, myworkflow
```

So this is both confusing and actually doesn't work.

In any event, I wanted to open discussion about possibly rethinking the syntax, or possibly just some names.

djarecka:
## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Summary

changing syntax for adding input fields during initialization (see discussion in #295)
- all inputs to the python function has to be included in `inputs` dictionary (and not as `kwargs`)
- for `ShellCommandTask` and `ContainerTask` everything but the default input fields (`executable`, `args`, `image`, `bindings`) also has to be provided in the `inputs` dictionary

## Checklist

- [x] All tests passing

## Acknowledgment


satra:
this is an api breaking change and let's take a decision on this and update the few repos that we have that use pydra

djarecka:
@satra - so you don't want to have it today/tomorrow release.

satra:
i mean this is as good a time as any, but perhaps we should ask a few people. because once changed, we would unlikely change it back.

djarecka:
@satra - you're right, should give more time. I wasn't planning to change tutorial before accepting this (pretty sure that won't discover anything new after changes in tests), but let me know if you want me to do it

satra:
i think we can have people look at this and if they agree we make another release soon after, but let's leave this out of the immediate release.

djarecka:
@satra - i have to say that I liked out current syntax, after all these changes I started thinking that perhaps it could be keep as an option, i.e. if `inputs` is used, we expect everything to be in the `inputs` dictionary, but if inputs is not specified than we take kwargs.

satra:
@effigies and @oesteban - would love your input on this. i agree with dorota's last comment. that it would make it much easier in most cases to take kwargs as input fields if there is no clash. but allow for inputs keyword if a user choses.

@djarecka - since the issue here is about clashes with potential attributes of the classes. perhaps we can check if the input spec has a field that clashes and in those cases force people to use inputs.

effigies:
I think that seems fine. I'm very likely to develop patterns where I can be very consistent, whether that's avoiding parameters that clash or always using inputs, but we don't have to make it impossible to do otherwise.

djarecka:
@satra - but perhaps we should not allow for mixing, i.e. some inputs in `inputs` some in the `kwargs`

satra:
> we should not allow for mixing

yes no mixing.

effigies:
We really let this one fall by the wayside. Going to need to be redone post-#662.

satra:
this can perhaps be closed given the syntax changes.

PR #307, opened 2020-07-04
fix: new location for prov context

satra:
## Types of changes

## Summary
original prov context is not available.

## Checklist

- [ ] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/schema/context.jsonld
+++ b/pydra/schema/context.jsonld
@@ -8,7 +8,7 @@
         "@container": "@type"
       }
     },
-    "https://openprovenance.org/prov.jsonld",
+    "https://raw.githubusercontent.com/openprov/prov-jsonld/69964ed16818f78dc5f71bdf97add026288291d4/context.json",
     {
       "pydra": "https://s.pydra.org/",
       "uid": "pydra:id/",


effigies:
Do we need to consider versioning our context file? And is there any consequence to this change for a consumer? Would it recognize the new version as the same as the old, or can we add some kind of assertion saying the object referenced by the new URI is the same as the old URI?

satra:
the consumer shouldn't need to change since we are changing this in the master for now. the other thing we can do is to include the contents of the versioned url into this context. 

in terms of versioning the context, we can consider it but not for the moment. it's fairly lightweight and till we formalize the prov pieces (including BEP028 and nidm-workflows), i would keep this in draft state. we should be careful not to remove things (which would cause some backward compatibility issues), but adding things should not be a problem.

effigies:
Okay. The new location looks like it's as stable as ours, so I wouldn't insist on including its contents.

LGTM.

djarecka:
since the original link works now, maybe we shouldn't merge it

satra:
this is still a more persistent link - so moving to this is better relative to the original link.

djarecka:
@satra - I'm still afraid that something else was also causing this error yesterday, but we can worry if it returns

PR #319, opened 2020-07-17
upgrade syntax to Python 3.7

basnijholt:
## Types of changes

## Checklist

- [x] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/helpers_state.py
+++ b/pydra/engine/helpers_state.py
@@ -575,7 +575,7 @@ def splits_groups(splitter_rpn, combiner=None, inner_inputs=None):
         combiner = []
     if inner_inputs:
         previous_states_ind = {
-            f"_{v.name}": v.keys_final for _, v in inner_inputs.items()
+            f"_{v.name}": v.keys_final for v in inner_inputs.values()
         }
         inner_inputs = {k: v for k, v in inner_inputs.items() if k in splitter_rpn}
     else:
--- a/pydra/engine/helpers_state.py
+++ b/pydra/engine/helpers_state.py
@@ -17,7 +17,7 @@ class PydraStateError(Exception):
         self.value = value
 
     def __str__(self):
-        return "{}".format(self.value)
+        return f"{self.value}"
 
 
 def splitter2rpn(splitter, other_states=None, state_fields=True):
@@ -311,7 +311,7 @@ def add_name_combiner(combiner, name):
     combiner_changed = []
     for comb in combiner:
         if "." not in comb:
-            combiner_changed.append("{}.{}".format(name, comb))
+            combiner_changed.append(f"{name}.{comb}")
         else:
             combiner_changed.append(comb)
     return combiner_changed
@@ -425,8 +425,7 @@ def splits(splitter_rpn, inputs, inner_inputs=None, cont_dim=None):
     # analysing states from connected tasks if inner_inputs
     if inner_inputs:
         previous_states_ind = {
-            "_{}".format(v.name): (v.ind_l_final, v.keys_final)
-            for _, v in inner_inputs.items()
+            f"_{v.name}": (v.ind_l_final, v.keys_final) for _, v in inner_inputs.items()
         }
         inner_inputs = {k: v for k, v in inner_inputs.items() if k in splitter_rpn}
     else:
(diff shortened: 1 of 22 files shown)


pydra/engine/helpers_state.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, value):
         self.value = value
 
     def __str__(self):
-        return "{}".format(self.value)
+        return f"{self.value}"
```
effigies, line 20: This seems like a weird one to use an f-string for.

```suggestion
        return str(self.value)
```

pydra/engine/helpers_state.py
```diff
@@ -425,8 +425,7 @@ def splits(splitter_rpn, inputs, inner_inputs=None, cont_dim=None):
     # analysing states from connected tasks if inner_inputs
     if inner_inputs:
         previous_states_ind = {
-            "_{}".format(v.name): (v.ind_l_final, v.keys_final)
-            for _, v in inner_inputs.items()
+            f"_{v.name}": (v.ind_l_final, v.keys_final) for _, v in inner_inputs.items()
```
effigies, line 428:
```suggestion
            f"_{v.name}": (v.ind_l_final, v.keys_final) for v in inner_inputs.values()
```

pydra/engine/helpers_state.py
```diff
@@ -576,7 +575,7 @@ def splits_groups(splitter_rpn, combiner=None, inner_inputs=None):
         combiner = []
     if inner_inputs:
         previous_states_ind = {
-            "_{}".format(v.name): v.keys_final for _, v in inner_inputs.items()
+            f"_{v.name}": v.keys_final for _, v in inner_inputs.items()
```
effigies, line 578:
```suggestion
            f"_{v.name}": v.keys_final for v in inner_inputs.values()
```

pydra/engine/task.py
```diff
@@ -625,7 +625,7 @@ def binds(self, opt, ind=None):
         """
         bargs = []
         for (key, val) in self.bind_paths(ind).items():
-            bargs.extend([opt, "{0}:{1}:{2}".format(key, val[0], val[1])])
+            bargs.extend([opt, "{}:{}:{}".format(key, val[0], val[1])])
```
effigies, line 628:
```suggestion
            bargs.extend([opt, f"{key}:{val[0]}:{val[1]}"])
```

pydra/engine/tests/data_tests/loading.py
```diff
@@ -2,7 +2,7 @@
 
 
 def loading(filename):
-    with open(filename, "r") as f:
+    with open(filename) as f:
```
effigies, line 5: Explicit is generally better than implicit. I would revert this one:

```suggestion
    with open(filename, "r") as f:
```

pydra/engine/tests/test_dockertask.py
```diff
@@ -1059,10 +1057,10 @@ def test_docker_cmd_inputspec_copyfile_1(plugin, tmpdir):
     assert res.output.out_file.exists()
     # the file is  copied, and than it is changed in place
     assert res.output.out_file.parent == docky.output_dir
-    with open(res.output.out_file, "r") as f:
+    with open(res.output.out_file) as f:
```
effigies, line 1060: Looks like there are a lot of these...

pydra/utils/profiler.py
```diff
@@ -86,7 +86,10 @@ def _sample(self, cpu_interval=None):
             except psutil.NoSuchProcess:
                 pass
 
-        print("%f,%f,%f,%f" % (time(), cpu, rss / _MB, vms / _MB), file=self._logfile)
+        print(
+            "{:f},{:f},{:f},{:f}".format(time(), cpu, rss / _MB, vms / _MB),
```
effigies, line 90:
```suggestion
            f"{time():f},{cpu:f},{rss / _MB:f},{vms / _MB:f}",
```

pydra/engine/tests/data_tests/loading.py
basnijholt: I feel this one is pretty obvious.

I also don't see `d.get(key, None)` everywhere for example 😄 :trollface:

djarecka: I use often `with open(filename, "r") as f`, but I agree that ` with open(filename) as f` is pretty standard and looks good to me

djarecka:
@basnijholt - thank you for the PR! The only tests that are failing are the ones that checks the style. This is because we are using `pre-commit` to run `black` with each commit

basnijholt:
I am just leaving for two weeks of vacation. So unfortunately I cannot run black now. I did give you guys push permission on this branch, if you want to fix it before merging. Otherwise you could just push one commit up on top of master after the merge.

djarecka:
If you want ton be included in the zenodo file, please edit the [file](https://github.com/nipype/pydra/blob/master/.zenodo.json) or provide the info, so I can add you

PR #320, opened 2020-07-18
adding a simple validator to the attrs class 

djarecka:
## Types of changes

## Summary
adding a type validator for attr classes  (not finished) - I had to write a wrapper around attr validators, since we are using `ty.Union` and `ty.List/Dict` and the validators can't check the types automatically. 
Decided to install `typing-inspect` to get the types from `ty.Union` etc., for py3.8 `typing` has its own method - `get_args`, but not for py3.7.
I'm not checking all combinations for now, i.e. I do not use recursion to check all depths of the type.

Checking values - for now only `allowed_values` ia added to the allowed metadata keys, but in the future it can be used for other validations.

## Checklist

- [ ] All tests passing

## Acknowledgment


pydra/engine/helpers.py
```diff
@@ -13,9 +13,13 @@
 import re
 from time import strftime
 from traceback import format_exception
+import typing as ty
+import typing_inspect as tyi
```
satra, line 17: can we drop this and code up these inspections ourselves? or use attrs?

satra: we do this in other parts of the code where we gather type information.

djarecka: if you have an example, happy to see

djarecka:
@satra - is this `on/off` flag should be on the tasks level, or you're thinking about `pydra` settings?

satra:
for on/off flag, it should be global (but with the potential to turn it on for a specific task if need be). but to start with at least global. see the last example here: https://www.attrs.org/en/stable/init.html#callables

for typing. attrs itself has direct support for this: https://www.attrs.org/en/stable/types.html and more specifically: https://www.attrs.org/en/stable/examples.html?highlight=validation#types

djarecka:
sure I can use typing in the classes, but their validators do not work well with types like `ty.List[int]`, try:
```
@attr.s
class OtherClass:
    a_number = attr.ib(validator=attr.validators.instance_of(int))
    list_of_numbers = attr.ib(validator=attr.validators.instance_of(ty.List[int]))
bb = OtherClass(a_number=3, list_of_numbers=[1, 2])
```

satra:
i'm simply saying that the information about typing can be retrieved through attrs (instead of the typing_inspect package), i'm not saying that you don't need to create custom validation routes for them.

djarecka:
if you know how to do it, could you please share it with me, because I did have a problem how to get `int` from `ty.List[int]`, or how to get `(str, int)` from `ty.Union[str, int]`.

satra:
```
In [22]: import typing 
    ...: @attr.s(auto_attribs=True) 
    ...: class AutoC: 
    ...:     cls_var: typing.ClassVar[int] = 5  # this one is ignored 
    ...:     l: typing.List[int] = attr.Factory(list) 
    ...:     x: int = 1 
    ...:     foo: str = attr.ib( 
    ...:          default="every attrib needs a type if auto_attribs=True" 
    ...:     ) 
    ...:     bar: typing.Any = None 
    ...:     zzz: typing.Optional[typing.Union[typing.List[str], int]] = ["a"] 
    ...:                                                                                        

In [23]: attr.fields(AutoC).zzz.type.__dict__                                                   
Out[23]: 
{'_inst': True,
 '_special': False,
 '_name': None,
 '__origin__': typing.Union,
 '__args__': (typing.List[str], int, NoneType),
 '__parameters__': (),
 '__slots__': None,
 '__module__': 'typing'}

In [24]: attr.fields(AutoC).l.type.__dict__                                                     
Out[24]: 
{'_inst': False,
 '_special': False,
 '_name': 'List',
 '__origin__': list,
 '__args__': (int,),
 '__parameters__': (),
 '__slots__': None}

```

djarecka:
ok, so i've removed `typing_inspect`, using `__args__`, `__origin__` etc. for python 3.7, but for python 3.8 we methods from `typing_inspect` are already inside `typing`

djarecka:
don't follow, so you wouldn't use the `validator` field in `attr.ib`?

Issue #298, opened 2020-06-22
refactor CIs


satra:
in reference to #293 

i think there is a broader refactoring of testing that i would like to see. and may be concentrating on one CI rather than 3 different ones (see the pydraml github workflow).

but before we switch CIs let's use this to discuss the different tests and actions

```
currently on travis (some on azure)
- tests
- slurm
- dask (currently not using a separately generated cluster)
- container tests
- pre-release check (allows failure)

- docs
  - build PR docs (currently on readthedocs)
  - push docs (currently on circle)
``` 

i think we can do a lot of this through github actions and perhaps use the other CIs for only the pieces that we cannot accomplish through github actions.

PR #328, opened 2020-08-06
CI: Switch to Github Actions

nicolocin:
## Types of changes

## Summary

Refactoring CI #298 
Tests pydra on ubuntu, macOS and windows on python 3.7 and 3.8

TODO:
 - [x] Run different ways of installing pydra
 - [x] Run docker tests  (merged [docker-ga](https://github.com/nicolocin/pydra/tree/docker-ga), allow fail before docker tests can run on windows.  Docker not available on MacOS in github actions)
 - [x] Run singularity tests  (merged [singularity-ga](https://github.com/nicolocin/pydra/tree/singularity-ga))
 - [x] Run dask tests  (add in separate PR)
 - [x] Run slurm tests  (merged [slurm-ga-ci](https://github.com/nicolocin/pydra/tree/slurm-ga-ci))
 - [x] Publish package to PyPI
 - [ ] Build docs 
 - [ ] Pre-release?
 - [ ] [Enable required status checks](https://docs.github.com/en/github/administering-a-repository/enabling-required-status-checks) for tests that are not allowed to fail 
 

## Checklist

- [x] All tests passing

## Acknowledgment


.github/workflows/testpydra.yml
```diff
@@ -0,0 +1,38 @@
...
+        python -m pip install --upgrade pip==20.2.1 setuptools==30.2.1 wheel
+    - name: Install dependencies
+      run: |
+        python -m pip install -r min-requirements.txt
+    - name: Install pydra
+      run: |
+        python -m pip install .[test]
+    - name: Pytest
+      env:
+        IGNORE_TESTS: '*dockertask.py'  # docker tests run in a seperate action for now
```
djarecka, line 33: any chances you forgot to add a new workflow to the git repository? I don't see the action with docker tests.

nicolocin: I have the docker one in a separate branch!  Not completely done yet - https://github.com/nicolocin/pydra/tree/docker-ga

djarecka:
@nicolocin - when you say that "seems like some errors are specific to certain os/python/installation combos." do you mean that this error consistently happens only in the specific combo? or they happen in one combo, but it's pretty random which one specifically? It's an important distinction IMO.
I'm asking since the specific error you pointed me was present in Travis, but I think it happened randomly from time to time, I believe I never had this locally

.github/workflows/testdockertask.yml
```diff
@@ -0,0 +1,39 @@
...
+
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }} on ${{ matrix.os }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Update build tools
+      run: python -m pip install --upgrade pip==20.2.1 setuptools==30.2.1 wheel
```
djarecka, line 31: I wonder if we should keep this hardcoded here. We probably should test (at least for one env) other pip/setuptools (as we did in travis). @satra - any opinion?

.github/workflows/testslurm.yml
```diff
@@ -0,0 +1,49 @@
...
+# https://steele.blue/tiny-github-actions/
+
+name: Run SLURM tests
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    continue-on-error: true
```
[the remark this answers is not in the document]
djarecka: I don't like `continue-on-error: true`. This basically say that whatever happens here, we will have green mark, am I right?

.github/workflows/testdockertask.yml
```diff
@@ -0,0 +1,39 @@
...
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    continue-on-error: true # until we fix tests on windows
```
djarecka, line 15: the same as for slurm - I'm not really convinced that there is any reason to use it. If I understand it correctly, whatever happens this will give us green light...

pydra/engine/tests/test_dockertask.py
```diff
@@ -50,7 +48,6 @@ def test_docker_1(plugin):
         assert "Unable to find image" in res.output.stderr
 
 
-@no_win
```
djarecka, line 53: I'm happy that we can remove `no_win` mark here, but we shouldn't be using `continue-on-error` instead... This marks are at least pretty explicit

djarecka:
Thanks @nicolocin ! I've left some comments in the code, but there are two main issue, overall the most important issues are:
- usage of `continue-on-error` - I'm not sure why we want to use it, I'd avoid it. 

- keeping docker test with everyone else: see my earlier comments

.github/workflows/testallowfail.yml
```diff
@@ -0,0 +1,37 @@
...
+    - name: Update build tools (pip 10.0.1)
+      run: python -m pip install pip==10.0.1 setuptools==30.3.0
+    - name: Install dependencies (min-requirements.txt)
+      run: pip install -r min-requirements.txt
+    - name: Install Pydra (test)
+      run: pip install -e ".[test]"
+    - name: Pytest
+      run: pytest -vs -n auto --cov pydra  --cov-config .coveragerc --cov-report xml:cov.xml --doctest-modules pydra
+    - name: Upload to codecov
+      run: codecov --root /pydra -f /pydra/cov.xml -F unittests
```
djarecka, line 37: does it work? can you please point me to a report?

.github/workflows/testdockertask.yml
```diff
@@ -0,0 +1,39 @@
...
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest] # docker not available on macOS
+        python-version: [3.7, 3.8]
+        install: [min-req, install, develop, sdist, wheel, style]
```
djarecka, line 20: `style` - only specific python and os

djarecka: we can  remove `sdist`

djarecka:
@nicolocin - I've thought that in `testallowfail.yml` we agreed to keep the tests that were failing with Travis, i.e. dask, but just realized that this is not a case. 
- Where do we test dask?
- looks like the tests from   `testallowfail.yml` fails because of the missing dependencies used by `etelemetry`: `ci`: https://github.com/nicolocin/pydra/runs/1096805344?check_suite_focus=true
That should be easy to fix (and perhaps `etelemetry` should also have it in dependencies, cc:@satra)

satra:
regarding etelemetry - i'm pretty positive `pip install etelemetry` installs its dependencies. so it makes me suspect something else in the stack (including perhaps older versions of pip). if etelemetry is indeed the issue, could someone post a clean example where pip doesn't install the dependencies?

djarecka:
@satra  - I think it is `pip==10.0.1`, not sure if we have to test for `pip==10.0.1`. But we should either remove it or fix it

setup.cfg
```diff
@@ -26,7 +26,7 @@ install_requires =
     attrs >= 19.1.0
     cloudpickle >= 1.2.2
     filelock >= 3.0.0
-    etelemetry >= 0.2.0
+    etelemetry >= 0.2.2
```
satra, line 29: you also have to regenerate min-requirements with python tools/update...py

djarecka:
@nicolocin - great job, thanks for finishing!

could you please address @satra comments from yesterday and make sure min-requirements are up to date

PR #329, opened 2020-08-08
fixing issues with numpy arrays provided as an input

djarecka:
## Types of changes

## Summary

The current splitter didn't work properly when numpy array were provided as an input. Modified `ensure_list` so it doesn't modify an array.

(might need some future work if we allow for "inner splitter" to also have numpy array)

Thanks to @Hoda1394 for reporting!

## Checklist

- [ ] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/helpers.py
+++ b/pydra/engine/helpers.py
@@ -44,10 +44,13 @@ def ensure_list(obj, tuple2list=False):
     """
     if obj is None:
         return []
-    if isinstance(obj, list):
+    # list or numpy.array (this might need some extra flag in case an array has to be converted)
+    elif isinstance(obj, list) or hasattr(obj, "__array__"):
         return obj
     elif tuple2list and isinstance(obj, tuple):
         return list(obj)
+    elif isinstance(obj, list):
+        return obj
     return [obj]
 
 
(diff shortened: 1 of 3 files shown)


Hoda1394:
> @Hoda1394 - would you have time to check if this branch works with your example?
@djarecka , Thanks for the fix. I will check it today.

Hoda1394:
@djarecka I tried it and it still gives me the same error despite the length of the two inputs are the same! it runs only when the second variable has the length of one! which is the case that the two variables definitely have not the same shape but there is no complaint from Pydra!
P.S. To elaborate more on the problem, I am using the scaler split over two inputs x and y in a pydra workflow. x is a list of NumPy arrays and y is a list of numbers. they have the same length but I receive this error: "ValueError: Operands wf1.y and wf1.x do not have same shape."

effigies:
Is there a minimal example that could be made into a test case?

Hoda1394:
@effigies I will create one.

Hoda1394:
@djarecka and @effigies, while preparing the test, I found out what was the problem and could fix it with a change in the input. now it works as expected.
Thanks

djarecka:
@satra - added a test for 2d arrays

satra:
thanks - LGTM.

PR #335, opened 2020-08-26
adding requires field to output_spec

djarecka:
## Types of changes

## Summary

- adding requires field to `output_spec`,
- simplifying functions in helpers (removing ` output_names_from_inputfields`)

## Checklist

- [ ] All tests passing

## Acknowledgment


pydra/engine/specs.py
```diff
@@ -449,6 +449,10 @@ def _field_defaultvalue(self, fld, output_dir):
 
     def _field_metadata(self, fld, inputs, output_dir):
         """Collect output file if metadata specified."""
+        if "requires" in fld.metadata:
+            for inp in fld.metadata["requires"]:
+                if getattr(inputs, inp) in [attr.NOTHING, None, False]:
```
satra, line 454: `False` could be a valid state for a field. 

```suggestion
                if getattr(inputs, inp) in [attr.NOTHING, None]:
```

pydra/engine/specs.py
```diff
@@ -461,7 +465,7 @@ def _field_metadata(self, fld, inputs, output_dir):
         elif "callable" in fld.metadata:
             return fld.metadata["callable"](fld.name, output_dir)
```
satra, line 466: do we have this documented somewhere?

djarecka: only tests probably...

pydra/engine/specs.py
djarecka: but usually you use flag that are bool

djarecka:
will open a new one

PR #336, opened 2020-08-28
conditional output: adding requires to output fields

djarecka:
## Types of changes

## Summary

* adding requires field to output_spec
* using fields from `output_file_template` as a default list of the required fileds (overwritten if `requires` present in the field metadata)
* simplifying functions in helpers (removing `output_names_from_inputfields`)

## Checklist

- [ ] All tests passing

## Acknowledgment


pydra/engine/tests/test_shelltask.py
```diff
@@ -2450,6 +2479,298 @@ def test_shell_cmd_outputspec_wf_1(plugin):
...
+        output_spec=my_output_spec,
+    )
+    shelly.inputs.file1 = "new_file_1.txt"
+    shelly.inputs.file2 = "new_file_2.txt"
+
+    res = shelly()
+    assert res.output.stdout == ""
+    assert res.output.newfile1.exists()
+    # additional input not provided so no newfile2 set (even if the file was created)
+    assert res.output.newfile2 is attr.NOTHING
```
satra, line 2771: should there be an additional check that when `addtional_inp` is provided that it is generated? 

also should there be a check/proper exception raised on the input/outputspec if someone uses a `output_file_template` which has an arg that is not defined in the inputs?  this should help users debug the specs better. perhaps a check_specs function that can be used in standalone mode by a user, but also during runtime.

pydra/engine/specs.py
```diff
@@ -461,7 +491,7 @@ def _field_metadata(self, fld, inputs, output_dir):
         elif "callable" in fld.metadata:
             return fld.metadata["callable"](fld.name, output_dir)
         else:
-            raise Exception("not implemented")
+            raise Exception("not implemented (_field_metadata)")
```
satra, line 494:
```suggestion
            raise Exception("(_field_metadata) is not a current valid metadata key.")
```

pydra/engine/specs.py
```diff
@@ -449,6 +450,35 @@ def _field_defaultvalue(self, fld, output_dir):
 
     def _field_metadata(self, fld, inputs, output_dir):
         """Collect output file if metadata specified."""
+        if "requires" in fld.metadata:
+            field_required = fld.metadata["requires"]
+        # if the output field doesn't have requires field, we use fields from output_file_template
+        elif "output_file_template" in fld.metadata:
```
satra, line 456: shouldn't this not be an elif and make sure that `requires` is a superset of the keys in `output_file_template` when both are provided? 

i.e. the intersection of keys from `output_file_template` and `requires` is equal to the keys from `output_file_template`.

pydra/engine/specs.py
```diff
@@ -449,6 +450,35 @@ def _field_defaultvalue(self, fld, output_dir):
...
+                    )
+                elif getattr(inputs, inp) in [attr.NOTHING, None]:
+                    return attr.NOTHING
+            elif isinstance(inp, tuple):  # (name, allowed values)
+                inp, allowed_val = inp
+                if not hasattr(inputs, inp):
+                    raise Exception(
+                        f"{inp} is not a valid input field, can't be used in requires"
+                    )
+                elif getattr(inputs, inp) not in allowed_val:
```
satra, line 475: i didn't see a test for this mode.

pydra/engine/specs.py
djarecka: yes, this makes sense, will change

pydra/engine/specs.py
djarecka: you're right, will add

pydra/engine/tests/test_shelltask.py
djarecka: will add a tests that shows that it works if `additional_inp` provided.

For the second question - i thought that we want to treat is as an "optional", so simply do not return the output when specific set of the input is not set. 
Or do you mean, that fields i the `requires` should be treated like this, but field that are not in the `requires` but only in the template should lead to the error if they are not set?

pydra/engine/specs.py
djarecka: but `output_file_template` can have names of outputs... I could select the ones that are in the `input_spec`, or just expect users to add all required inputs to the `requires` list

satra: consider whatever is necessary to validate and/or provide the user with a clear error message.

djarecka: but this is related to my other question I asked yesterday. Let me repeat and rephrase:

If a field from `requires` list (assuming for now that it also takes what is needed from the template) is not present:
a) we raise an exception (with clear message)
b) we set `NOTHING` for this specific output field

I was assuming that the answer is `b`, but from you questions I'm guessing that you're thinking about `a`... 
I really think that `b` makes more sense, because:
- we can always set an input field as a `mandatory` if you want to have exceptions when it's not set
- as we were discussing, many interfaces from Nipype has multiple output that is created only for a specific combination for the input, and I was hoping that `requires` can be used to solve this issue, and not to raise errors.

satra: yes, outputs should be set to ATTR.nothing when requirements aren't satisfied. 

my hope was to provide appropriate error messages when output_file_templates are used with/without requires:
1. without requires: is the output generated if some input is missing? i would say without requires an output is always created based on output_file_template. and should provide appropriate error message to user when an input to the template is missing.
2. with requires: the output is only generated when requirements are satisfied.`output_file_template` can still error when used and should again generate appropriate error message.

djarecka: doesn't this contradict what you wrote yesterday - " the intersection of keys from output_file_template and requires is equal to the keys from output_file_template"?

I've made changes today and added all fields from ` output_file_template` to `requires`..

djarecka: but as for the idea, sure, we can do it

satra: i think it's a bit different:

1. intersection, when `requires` exists, and is satisfied, should also have the `output_file_template` fields. i.e. if the requirements are satisfied then the `output_file_template` fields are also required. 

2. if it's just `output_file_template` without any `requires`, then i agree with your interpretation that the output should only be created if all `args` are available. to enforce generation when requirements are satisfied, the task creator should always add the `requires` metadata to output fields.

djarecka: the second point is very unclear to me:
- what are you calling `args`
- what you mean in " to enforce generation when requirements are satisfied, the task creator should always add the requires metadata to output fields."?  Do you meant that all fields from `output_file_template` should be added to `requires` (that is not provided, so can be treated as an empty list). Isn't this the same as in point 1?

What about the following strategy?
- If you want to have an input that has to be provided - use `mandatory=True` for the input field (if the input not provided the exception will be raised)
- If you want to have a specific combination of the input you can use `requires` and `xor` for the input fields (if the provided input doesn't match `requires/xor` an exception is raised)
- if the specific input, say `inp_notmand` is not mandatory, but without the input field the output field is not created, then the output field should have `requires=["inp_notmand"]`. If `inp_notmand` is not provided the output is not created (no exception). No `requires` is treated as `requires=[]`.
- the template should only include fields from the `requires` list and fields that are always provided (ideally they should have `mandatory=True`). If all fields that have `mandatory=True` are provide, pydra checks `requires`. If all fields from the list are set, then it tries to create the name. If there is still something missing, then we raise the error.

satra: i'm fine with the strategy. i don't think anything changes except for how `requires` is assigned for something that only has an `output_file_template` and no `requires`.  if we are to say that `output_file_template` should always have requires, then it goes back to simply adding `output_file_template` format args to `requires`.

djarecka: maybe I will keep for now as it is (i.e. no errors are raised, just output is not created). I really don't like the idea that fields from `output_file_template` are treated differently when `requires` is present. 
My strategy from yesterday has a problem, that I had to add to many `requires` to the tests, and I don't like it anymore...
Let me know if you have strong objections, I never fully understood why it is important to you that `output_file_template` can lead to an error

PR #348, opened 2020-09-19
[mnt, wip] ShellCommandTask updates (motivated my nipype interfaces)

djarecka:
## Types of changes

## Summary
- adding a new test file with tests related to the nipype's interfaces conversion
- changing ShellCommandTask, so the child classes can have a class attribute for input/output_specs and executable/args

## Checklist

- [ ] All tests passing

## Acknowledgment

### DIFF
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -43,6 +43,7 @@ import cloudpickle as cp
 import inspect
 import typing as ty
 from pathlib import Path
+import warnings
 
 from .core import TaskBase, is_lazy
 from ..utils.messenger import AuditFlag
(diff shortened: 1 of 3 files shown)


pydra/engine/task.py
```diff
@@ -275,15 +276,32 @@ def __init__(
...
-
-        self.output_spec = output_spec
+        if input_spec is not None:
+            self.input_spec = input_spec
+        elif not hasattr(self, "input_spec"):
+            self.input_spec = SpecInfo(name="Inputs", fields=[], bases=(ShellSpec,))
+        else:
+            # changing class attribute to instance attribute, so it is part of __dict__
+            # (used in TaskBase.__set/get-state__)
+            self.input_spec = self.input_spec
```
satra, line 286: this is weird. this should automatically happen in python. with the code that was deleted.

```
In [1]: class A:
   ...:     var1 = None
   ...:     def __init__(self, var1=None):
   ...:         self.var1 = var1
   ...: 

In [2]: a = A()

In [3]: a.__dict__
Out[3]: {'var1': None}
```

djarecka: but my comment was about the class attribute, your example set `var1` also in init

pydra/engine/task.py
```diff
@@ -275,15 +276,32 @@ def __init__(
             TODO
 
         """
-        if input_spec is None:
-            input_spec = SpecInfo(name="Inputs", fields=[], bases=(ShellSpec,))
-        self.input_spec = input_spec
```
satra, line 280: this is set in the init

djarecka: yes, the old code was working fine.

But what I did in this PR is allowing to use the class attribute (if no `input_spec` is provided during initialization), so it can use the default `input_spec` for the class, se [this](https://github.com/nipype/pydra/pull/348/files#diff-1a747d8c932de194d2dda44683c02558R22).

djarecka: so before I set `input_spec = SpecInfo(name="Inputs", fields=[], bases=(ShellSpec,))`, I also want to check for the class attribute

satra: you can do something like this:

```
In [12]: class A:
    ...:     var1 = None
    ...:     def __init__(self, var1=None):
    ...:         self.var1 = var1 or self.var1 or 'fixed'
    ...: 

In [13]: a = A()

In [14]: a.__dict__
Out[14]: {'var1': 'fixed'}

In [15]: class B(A):
    ...:     var1 = 'derived'
    ...: 

In [16]: b = B()

In [17]: b.__dict__
Out[17]: {'var1': 'derived'}

```

djarecka: sure, i believe this is exactly what i did, just used more explicit syntax, but can move to yours.

djarecka: it's definitely much shorter :)

PR #361, opened 2020-10-02
[mnt] slurm updates: sbatch args and requeuing

djarecka:
## Acknowledgment

## Types of changes

## Summary

- tests for adding `sbatch_args` to the Submitter
- better error handling for invalid slurm options
- options for requeuing if the job is cancelled 

## Checklist

- [ ] All tests passing

### DIFF
--- a/pydra/engine/workers.py
+++ b/pydra/engine/workers.py
@@ -1,6 +1,6 @@
 """Execution workers."""
 import asyncio
-import sys
+import sys, os
 import re
 from tempfile import gettempdir
 from pathlib import Path
@@ -300,7 +300,12 @@ class SlurmWorker(DistributedWorker):
             # Exception: Polling / job failure
             done = await self._poll_job(jobid)
             if done:
-                return True
+                if done == "CANCELLED":
+                    os.remove(cache_dir / f"{checksum}.lock")
+                    cmd_re = ("scontrol", "requeue", jobid)
+                    await read_and_display_async(*cmd_re, hide_display=True)
+                else:
+                    return True
             await asyncio.sleep(self.poll_delay)
 
     async def _poll_job(self, jobid):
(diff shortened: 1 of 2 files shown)


pydra/engine/workers.py
```diff
@@ -296,7 +300,12 @@ async def _submit_job(self, batchscript, name, checksum, cache_dir):
             # Exception: Polling / job failure
             done = await self._poll_job(jobid)
             if done:
-                return True
+                if done == "CANCELLED":
+                    os.remove(cache_dir / f"{checksum}.lock")
```
mgxd, line 304:
```suggestion
                    (cache_dir / f"{checksum}.lock").unlink(missing_ok=True)
```

mgxd (commented):
Hi Dorota! Does this constantly requeue if a job is cancelled, even if by the user themselves? If so, it might be worth setting this switch as a parameter when initializing a `SlurmWorker` and/or adding a limit to the number of requeues, WDYT?

satra:
also on submitter exit, which could happen if someone hits ctrl+c we may want to actively cancel all running jobs.

@mgxd -  in most cases the auto requeue is going to be the default. but, i agree that canceling it would be good option to add to allow for users to let a job fail without disrupting the rest of the workflow. in such a case the exception associated with the job should note that the task was killed by an external process.

more generally we currently don't have a config setup in pydra, but before introducing config, i think this could be set as a worker parameter for now.

djarecka:
@mgxd @satra - yes, right now it will go forever, so setting some max number would be good.
I was wondering also if there is a slurm command that could stop the proces or rerunning (i.e. if I could stop regardless what I set in pydra).

Also, if you have an idea how to prepare a good automatic test for it, please let me know!

djarecka:
I was planning to add the job array in this pr, but will move it to a new one.

I hope the tests are now enough robust (this version seems to work fine on openmind and in the slurm container)

satra:
thanks @djarecka - looks reasonable.

PR #375, opened 2020-11-07
fixing the memory issue for workflows with a big splitters (closes #371)

### LINKED ISSUE #371: big memory use for workflows
The problem was originally reported by @Hoda1394 and I'm currently testing suing this kind of workflows:

```
@mark.task
def show_var(a):
    return a

def generate_list(l):
    my_list = np.arange(l).tolist()
    return my_list

wf= Workflow(name="wf", input_spec=["x"])
wf.split("x", x=generate_list(size))
wf.add(show_var(name="show", a=wf.lzin.x))
wf.set_output([("out", wf.show.lzout.out)])
wf()
```

So I do see a non-linear (quadratic?) relation between memory and the size of the list. I decided to rely mostly on the RSS memory measure from GNU time, and here is a plot:
rss_wf

This is a workflow related issue, I've tested tasks with a splitter and the relation is linear.

I've been struggling to debug it, but I got to a point that I think I know where the problem is generated, but not sure why. It could be also that I'm simply missing something else. I've modified `submit` method to add a simple print statement (wrapped in a coroutine) to `futures` instead of `worker.run_el` (see [here](https://github.com/djarecka/pydra/blob/tmp_testmem_2.1/pydra/engine/submitter.py#L137)) and this single modified line still leads to increase the memory. Any thoughts? (cc:@satra, @mgxd )

djarecka:
## Acknowledgment

## Types of changes

## Summary

- removing state_inputs from  Task class (was not used anymore), 
- fixing issues for workflow with a big splitter 
- adding tests to check the memory object together with pympler library 

## Checklist

- [x] All tests passing

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -163,7 +163,6 @@ class TaskBase:
 
         # checking if metadata is set properly
         self.inputs.check_metadata()
-        self.state_inputs = inputs
         # dictionary to save the connections with lazy fields
         self.inp_lf = {}
         self.state = None
@@ -481,7 +480,6 @@ class TaskBase:
             )
         if kwargs:
             self.inputs = attr.evolve(self.inputs, **kwargs)
-            self.state_inputs = kwargs
         if not self.state or splitter != self.state.splitter:
             self.set_state(splitter)
         return self
--- a/pydra/engine/tests/test_profiles.py
+++ b/pydra/engine/tests/test_profiles.py
@@ -4,6 +4,45 @@ from ... import mark
 
 import numpy as np
 from pympler import asizeof
+from pytest import approx
+
+
+def generate_list(l):
+    return np.arange(l).tolist()
+
+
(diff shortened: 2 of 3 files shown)


djarecka:
@satra - not sure how I missed this earlier, but it was just about the task that are loaded, we were resetting state and inputs, but not `state_inputs` that was just a remainder of one of the old versions

satra:
if it's that simple, that's great.

djarecka:
@satra - it came from the `futures` set, but python/asyncio are responsible for the linear relation, the quadratic part was in the task part itself...

djarecka:
@satra - are you ok with adding additional library to the test installation? I think this simple test could be pretty useful

satra:
yes a testing library is fine for now. we need to refactor test dependencies anyway.

djarecka:
ok, added an additional test, once it passes I will merge it

PR #382, opened 2020-11-21
Fix/input hash

djarecka:
## Acknowledgment

## Types of changes

## Summary

- removing all hash function calls before run function call
- using `uuid` instead of hah function for slurm scripts etc.
- adding a `changed` flag to `BaseSpec` to track if the class has changed before recalculating the hash value again

## Checklist

- [ ] All tests passing

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -6,7 +6,7 @@ import logging
 import os
 from pathlib import Path
 import typing as ty
-from copy import deepcopy, copy
+from copy import deepcopy
 from uuid import uuid4
 
 import cloudpickle as cp
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -252,29 +252,6 @@ class TaskBase:
             )
         return self._checksum
 
-    @property
-    def uid(self):
-        """ setting the unique id number for the task
-            It will be used to create unique names for slurm scripts etc.
-            without a need to run checksum
-        """
-        if not self._uid:
-            self._uid = str(uuid4())
-        return self._uid
-
-    def uid_states(self):
-        """ setting a list of the unique id numbers for the task with a splitter
-            It will be used to create unique names for slurm scripts etc.
-            without a need to run checksum
-        """
-        if self._uid_states is None:
-            self._uid_states = []
-            if not hasattr(self.state, "inputs_ind"):
-                self.state.prepare_states(self.inputs)
-            for ind in range(len(self.state.inputs_ind)):
-                self._uid_states.append(str(uuid4()))
-        return self._uid_states
-
     def checksum_states(self, state_index=None):
         """
         Calculate a checksum for the specific state or all of the states of the task.
@@ -287,6 +264,9 @@ class TaskBase:
             TODO
 
         """
+        if is_workflow(self) and self.inputs._graph_checksums is attr.NOTHING:
+            self.inputs._graph_checksums = [nd.checksum for nd in self.graph_sorted]
+
         if state_index is not None:
             inputs_copy = deepcopy(self.inputs)
             for key, ind in self.state.inputs_ind[state_index].items():
@@ -313,6 +293,29 @@ class TaskBase:
             for ind in range(len(self.state.inputs_ind)):
                 checksum_list.append(self.checksum_states(state_index=ind))
             return checksum_list
+
+    @property
+    def uid(self):
+        """ setting the unique id number for the task
+            It will be used to create unique names for slurm scripts etc.
+            without a need to run checksum
+        """
+        if not self._uid:
+            self._uid = str(uuid4())
+        return self._uid
+
+    def uid_states(self):
+        """ setting a list of the unique id numbers for the task with a splitter
+            It will be used to create unique names for slurm scripts etc.
+            without a need to run checksum
+        """
+        if self._uid_states is None:
+            self._uid_states = []
+            if not hasattr(self.state, "inputs_ind"):
+                self.state.prepare_states(self.inputs)
+            for ind in range(len(self.state.inputs_ind)):
+                self._uid_states.append(str(uuid4()))
+        return self._uid_states
 
     def set_state(self, splitter, combiner=None):
         """
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -7,6 +7,7 @@ import os
 from pathlib import Path
 import typing as ty
 from copy import deepcopy, copy
+from uuid import uuid4
 
 import cloudpickle as cp
 from filelock import SoftFileLock
@@ -183,6 +184,8 @@ class TaskBase:
         self.cache_locations = cache_locations
         self.allow_cache_override = True
         self._checksum = None
+        self._uid = None
+        self._uid_states = None
         # if True the results are not checked (does not propagate to nodes)
         self.task_rerun = rerun
 
@@ -248,6 +251,29 @@ class TaskBase:
                 self.__class__.__name__, hash_function([input_hash, splitter_hash])
             )
         return self._checksum
+
+    @property
+    def uid(self):
+        """ setting the unique id number for the task
+            It will be used to create unique names for slurm scripts etc.
+            without a need to run checksum
+        """
+        if not self._uid:
+            self._uid = str(uuid4())
+        return self._uid
+
+    def uid_states(self):
+        """ setting a list of the unique id numbers for the task with a splitter
+            It will be used to create unique names for slurm scripts etc.
+            without a need to run checksum
+        """
+        if self._uid_states is None:
+            self._uid_states = []
+            if not hasattr(self.state, "inputs_ind"):
+                self.state.prepare_states(self.inputs)
+            for ind in range(len(self.state.inputs_ind)):
+                self._uid_states.append(str(uuid4()))
+        return self._uid_states
 
     def checksum_states(self, state_index=None):
         """
@@ -546,8 +572,8 @@ class TaskBase:
         """ Pickling the tasks with full inputs"""
         pkl_files = self.cache_dir / "pkl_files"
         pkl_files.mkdir(exist_ok=True, parents=True)
-        task_main_path = pkl_files / f"{self.name}_{self.checksum}_task.pklz"
-        save(task_path=pkl_files, task=self, name_prefix=f"{self.name}_{self.checksum}")
+        task_main_path = pkl_files / f"{self.name}_{self.uid}_task.pklz"
+        save(task_path=pkl_files, task=self, name_prefix=f"{self.name}_{self.uid}")
         return task_main_path
 
     @property
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -184,7 +184,7 @@ class TaskBase:
         self.cache_locations = cache_locations
         self.allow_cache_override = True
         self._checksum = None
-        self._uid = str(uuid4())
+        self._uid = uuid4().hex
         # if True the results are not checked (does not propagate to nodes)
         self.task_rerun = rerun
 
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -438,7 +438,10 @@ class TaskBase:
         lockfile = self.cache_dir / (checksum + ".lock")
         # Eagerly retrieve cached - see scenarios in __init__()
         self.hooks.pre_run(self)
-        # TODO add signal handler for processes killed after lock acquisition
+        # adding info file with the checksum in case the task was cancelled
+        # and the lockfile has to be removed
+        with open(self.cache_dir / f"{self.uid}_info.json", "w") as jsonfile:
+            json.dump({"checksum": self.checksum}, jsonfile)
         with SoftFileLock(lockfile):
             if not (rerun or self.task_rerun):
                 result = self.result()
@@ -962,6 +965,10 @@ class Workflow(TaskBase):
                 "Workflow output cannot be None, use set_output to define output(s)"
             )
         checksum = self.checksum
+        # adding info file with the checksum in case the task was cancelled
+        # and the lockfile has to be removed
+        with open(self.cache_dir / f"{self.uid}_info.json", "w") as jsonfile:
+            json.dump({"checksum": checksum}, jsonfile)
         lockfile = self.cache_dir / (checksum + ".lock")
         # Eagerly retrieve cached
         if not (rerun or self.task_rerun):
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -184,7 +184,7 @@ class TaskBase:
         self.cache_locations = cache_locations
         self.allow_cache_override = True
         self._checksum = None
-        self._uid = None
+        self._uid = str(uuid4())
         # if True the results are not checked (does not propagate to nodes)
         self.task_rerun = rerun
 
@@ -295,12 +295,10 @@ class TaskBase:
 
     @property
     def uid(self):
-        """ setting the unique id number for the task
+        """ the unique id number for the task
             It will be used to create unique names for slurm scripts etc.
             without a need to run checksum
         """
-        if not self._uid:
-            self._uid = str(uuid4())
         return self._uid
 
     def set_state(self, splitter, combiner=None):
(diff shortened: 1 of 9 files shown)


djarecka:
@satra @nicolocin - please check the changes. The task submission should be really quick right now (and this is my observation from running some tests on openmind, but @nicolocin can double check for her script).

As for the hash recompute - I originally was changing the `TaskBase` class and was experimenting with input as property, but at the end I decided to change `BaseSpec` class and created a new flag `changed`. 

I'm still thinking that for files I could create an extra dictionary when we can save the hashed values. Right now all hash values will be recomputed if anything changes.

pydra/engine/workers.py
```diff
@@ -246,27 +209,60 @@ def run_el(self, runnable, rerun=False):
...
-    async def _submit_job(self, batchscript, name, checksum, cache_dir):
+    async def _submit_job(self, batchscript, name, uid, cache_dir):
         """Coroutine that submits task runscript and polls job until completion or error."""
-        script_dir = cache_dir / f"{self.__class__.__name__}_scripts" / checksum
+        script_dir = cache_dir / f"{self.__class__.__name__}_scripts" / uid
         sargs = self.sbatch_args.split()
         jobname = re.search(r"(?<=-J )\S+|(?<=--job-name=)\S+", self.sbatch_args)
         if not jobname:
-            jobname = ".".join((name, checksum))
+            jobname = ".".join((name, uid))
```
satra, line 265: this is actually better. otherwise two of the same jobs could be submitted.

pydra/engine/workers.py
```diff
@@ -304,15 +300,51 @@ async def _submit_job(self, batchscript, name, checksum, cache_dir):
                     done in ["CANCELLED", "TIMEOUT", "PREEMPTED"]
                     and "--no-requeue" not in self.sbatch_args
                 ):
-                    if (cache_dir / f"{checksum}.lock").exists():
+                    if (cache_dir / f"{uid}.lock").exists():
```
satra, line 303: this part i don't understand. the task lock file should use the checksum, since we want the same exact jobs to be locked when run.

pydra/engine/workers.py
```diff
@@ -304,15 +300,51 @@ async def _submit_job(self, batchscript, name, checksum, cache_dir):
...
                         # for pyt3.8 we could you missing_ok=True
-                        (cache_dir / f"{checksum}.lock").unlink()
+                        (cache_dir / f"{uid}.lock").unlink()
```
satra, line 305: same here.

pydra/engine/specs.py
```diff
@@ -104,9 +114,11 @@ def hash(self):
             )
         inp_hash = hash_function(inp_dict)
         if hasattr(self, "_graph_checksums"):
-            return hash_function((inp_hash, self._graph_checksums))
-        else:
-            return inp_hash
+            inp_hash = hash_function((inp_hash, self._graph_checksums))
+        # setting inp_hash and changing the flag to False
+        super().__setattr__("inp_hash", inp_hash)
+        super().__setattr__("changed", False)
```
satra, line 120: not sure why these two lines use super.

pydra/engine/core.py
```diff
@@ -546,8 +575,8 @@ def pickle_task(self):
         """ Pickling the tasks with full inputs"""
         pkl_files = self.cache_dir / "pkl_files"
         pkl_files.mkdir(exist_ok=True, parents=True)
-        task_main_path = pkl_files / f"{self.name}_{self.checksum}_task.pklz"
-        save(task_path=pkl_files, task=self, name_prefix=f"{self.name}_{self.checksum}")
+        task_main_path = pkl_files / f"{self.name}_{self.uid}_task.pklz"
+        save(task_path=pkl_files, task=self, name_prefix=f"{self.name}_{self.uid}")
```
satra, line 579: are you creating separate task files for each split. that may not be optimal in the previous implementation you had one file from which you read all the states right?

satra (changes requested):
i think the slurm rerun/cancel job check needs to be revised. this may require computing the task checksum, but it should not need to be computed more than once.

satra:
@djarecka  - you may want to check the slurm locally in the docker container. it's been running for 3+ hours. may be worthwhile adding a timeout to the slurm test so that it fails if it takes more then 30 mins.

but i think the reason it's running may be related to checking uid instead of checksum.

djarecka:
@satra - I was testing with the slurm container, but I guess I didn't run all the tests in it... Cancelled the slurm workflow for now

pydra/engine/specs.py
djarecka: i wanted to avoid situation  when setting these attributes will set `changed` to `True`, but perhaps I should just use an `if ` statement in `self.__setattr__`

pydra/engine/core.py
djarecka: yes, one file. what is your concern?

satra: if it's one file, i don't understand (at this point) why each state split has a different uid? where is that being used?

pydra/engine/specs.py
```diff
@@ -90,6 +97,9 @@ def hash(self):
         """Compute a basic hash for any given set of fields."""
         from .helpers import hash_value, hash_function
 
+        # if inp_hash already calculated and nothing has changed, the old value can be used
+        if self.changed is False and self.inp_hash:
+            return self.inp_hash
         inp_dict = {}
         for field in attr_fields(self):
             if field.name in ["_graph_checksums", "bindings"] or field.metadata.get(
```
satra, line 105: shouldn't this if skip hashing `inp_hash` and `changed`. also attrs has a function called `on_setter` that you can use as well.

pydra/engine/core.py
djarecka: let me think... i keep braking things today, so perhaps should double check this

pydra/engine/specs.py
djarecka: it should

pydra/engine/workers.py
djarecka: I somehow missed this commend yesterday. I've tried to remove all checsksums from worker and submitter. I thought that this is what we discussed. 
But indeed checksum probably make more sense here (and likely solves other problem I have)

pydra/engine/core.py
djarecka: this is called for the task with full input, not for every state separately

pydra/engine/specs.py
djarecka: i couldn't find `on_setter`, could you please point me to it

djarecka:
@satra - i've reverted the changes for the names of lockfiles and created extra file to read value of checksums.

I've also created another attribute for `BaseSpec` - `files_hash` to save hash values for files, please let me know what do you think

djarecka:
@satra - I made the changes we discussed, but I'm not sure about the `Audit` class. First I thought that perhaps I could use `task.uid` [here](https://github.com/nipype/pydra/blob/master/pydra/engine/audit.py#L50), but than I realized that this is unique for every single message.

pydra/engine/specs.py
```diff
@@ -66,11 +66,20 @@ class SpecInfo:
 class BaseSpec:
     """The base dataclass specs for all inputs and outputs."""
 
+    inp_hash = None
+    # a flag to check if anything has changed since the last hash was calculated
+    changed = None
```
satra, line 71:
```suggestion
    changed = True
```

pydra/engine/specs.py
```diff
@@ -90,23 +101,66 @@ def hash(self):
         """Compute a basic hash for any given set of fields."""
         from .helpers import hash_value, hash_function
 
+        # if inp_hash already calculated and nothing has changed, the old value can be used
+        if self.changed is False and self.inp_hash:
```
satra, line 105: is this `if` necessary. shouldn't you always run this check since you don't have control over external files.

pydra/engine/specs.py
```diff
@@ -90,23 +101,66 @@ def hash(self):
...
+            # if there are files in self.files_hash, checking if they didn't changed
+            check_files = [
+                Path(val[0]).stat().st_mtime == val[1]
+                for val in self.files_hash.values()
+            ]
+            if all(check_files):
+                return self.inp_hash
+        if self.files_hash is None:
+            self.files_hash = {}
```
satra, line 114: shouldn't this be in `init`?

djarecka: so, should i create an init method? it could be

pydra/engine/specs.py
```diff
@@ -90,23 +101,66 @@ def hash(self):
...
+                "changed",
+                "files_hash",
+            ] or field.metadata.get("output_file_template"):
                 continue
             # removing values that are notset from hash calculation
             if getattr(self, field.name) is attr.NOTHING:
                 continue
+            value = getattr(self, field.name)
+            # checking if the field name is in the dictionary with pre-computed hash values for files
+            if field.name in self.files_hash:
```
satra, line 130: instead of this if, couldn't you do this in the check for files below?

pydra/engine/specs.py
```diff
@@ -90,23 +101,66 @@ def hash(self):
...
             inp_dict[field.name] = hash_value(
-                value=getattr(self, field.name), tp=field.type, metadata=field.metadata
+                value=value, tp=field.type, metadata=field.metadata
             )
+            # if file/dir, the hash value will be saved to prevent recomputation
+            if (
+                field.type in [File, Directory]
+                or "pydra.engine.specs.File" in str(field.type)
+                or "pydra.engine.specs.File" in str(field.type)
+            ) and is_existing_file(value):
```
satra, line 149: what if value is a list of files or some other structure with File as an inner type. this was taken care of in hash_value

pydra/engine/workers.py
```diff
@@ -246,27 +209,60 @@ def run_el(self, runnable, rerun=False):
...
         if (script_dir / script_dir.parts[1]) == gettempdir():
             logger.warning("Temporary directories may not be shared across computers")
-
         if isinstance(runnable, TaskBase):
-            checksum = runnable.checksum
             cache_dir = runnable.cache_dir
             name = runnable.name
+            uid = runnable.uid
         else:
-            checksum = runnable[-1].checksum_states()[runnable[0]]
```
satra, line 255: i'm not following this logic here, perhaps useful to add to docstring. how long can runnable be? what is the 0 index and the -1 index?

pydra/engine/workers.py
```diff
@@ -246,27 +209,60 @@ def run_el(self, runnable, rerun=False):
...
+        return self._submit_job(batch_script, name=name, uid=uid, cache_dir=cache_dir)
+
+    def _prepare_runscripts(self, task, interpreter="/bin/sh", rerun=False):
+        if isinstance(task, TaskBase):
+            cache_dir = task.cache_dir
+            ind = None
+            uid = task.uid
+        else:
+            ind = task[0]
+            cache_dir = task[-1].cache_dir
```
satra, line 230: same comment here with task. what is `task[0]` and `task[-1]` and how long can task be?

pydra/engine/core.py
```diff
@@ -288,6 +294,29 @@ def checksum_states(self, state_index=None):
...
             return checksum_list
 
+    @property
+    def uid(self):
+        """ setting the unique id number for the task
+            It will be used to create unique names for slurm scripts etc.
+            without a need to run checksum
+        """
+        if not self._uid:
+            self._uid = str(uuid4())
```
satra, line 304: shouldn't this be in the `init`?

pydra/engine/core.py
```diff
@@ -183,6 +184,8 @@ def __init__(
         self.cache_locations = cache_locations
         self.allow_cache_override = True
         self._checksum = None
+        self._uid = None
+        self._uid_states = None
```
satra, line 188: still worried about this as this now generates a uuid for every split and stores it in the task. perhaps this can be dynamically generated at the point of need.

pydra/engine/specs.py
djarecka: you mean, without using `self.changed` at all? i think you're right

pydra/engine/workers.py
djarecka: i've added a comment, this is for situation when `runnable` is in fact (ind, pkl file, runnable)

pydra/engine/core.py
djarecka: you're right, this is pretty useless - right now it's used only in `run_el`, removed

pydra/engine/specs.py
djarecka: in the new version i've changed this completely - `hash_file` compares the modification times and hashes

pydra/engine/core.py
```diff
@@ -183,6 +184,7 @@ def __init__(
         self.cache_locations = cache_locations
         self.allow_cache_override = True
         self._checksum = None
+        self._uid = str(uuid4())
```
satra, line 187: perhaps this form:

```suggestion
        self._uid = uuid4().hex
```

pydra/engine/helpers.py
```diff
@@ -803,6 +806,8 @@ def load_task(task_pkl, ind=None):
         _, inputs_dict = task.get_input_el(ind)
         task.inputs = attr.evolve(task.inputs, **inputs_dict)
         task.state = None
+        # resetting uid for task
+        task._uid = str(uuid4())
```
satra, line 810:
```suggestion
        task._uid = uuid4().hex
```

pydra/engine/helpers_file.py
```diff
@@ -77,14 +79,26 @@ def hash_file(afile, chunk_len=8192, crypto=sha256, raise_notfound=True):
...
         while True:
             data = fp.read(chunk_len)
             if not data:
                 break
             crypto_obj.update(data)
-    return crypto_obj.hexdigest()
+
+    cont_hash = crypto_obj.hexdigest()
+    if precalculated is not None:
+        precalculated[str(Path(afile))] = (Path(afile).stat().st_mtime, cont_hash)
```
satra, line 100: does this key get deleted somewhere when the path is removed? or `precalculated` is reset?

djarecka: I've decided to reset in `BaseSpec` for each name of the input name, see https://github.com/nipype/pydra/pull/382/files#diff-e76a5941f8f208ec57e47ec8c8233030ced110ae66222044ae5287f52bf65ed2R93

pydra/engine/specs.py
```diff
@@ -66,17 +66,31 @@ class SpecInfo:
...
+            "changed",
+            "files_hash",
+        ]:
             super().__setattr__(name, value)
         else:
             tp = attr.fields_dict(self.__class__)[name].type
             # if the type has a converter, e.g., MultiInputObj
             if hasattr(tp, "converter"):
                 value = tp.converter(value)
+            self.files_hash[name] = {}
```
satra, line 93: i see this is where `precalculated` is reset. just a note that in the future we may want to optimize this further. for example if you have a list of files and i change one or two, it would be nice to not have to recompute the hash of all files. we may want to consider a more global filehash LRU cache that various tasks can dip into.

pydra/engine/helpers_file.py
djarecka: perhaps the test could also help when reading the pr: https://github.com/nipype/pydra/pull/382/files#diff-65ddfb82f62778d7637f5040b6d46040a9ee599cd5e42fa78d00dfc333413851R245

pydra/engine/submitter.py
```diff
@@ -158,6 +156,9 @@ async def _run_workflow(self, wf, rerun=False):
         # creating a copy of the graph that will be modified
         # the copy contains new lists with original runnable objects
         graph_copy = wf.graph.copy()
+        # resetting uid for nodes in the copied workflows
+        for nd in graph_copy.nodes:
+            nd._uid = str(uuid4())
```
satra, line 161:
```suggestion
            nd._uid = uuid4().hex
```

pydra/engine/specs.py
djarecka: that was my first approach - I had a global dictionary, but than I recalled that you worried that this dictionary would grow too much and I changed to nested dictionary. we can think more later

PR #388, opened 2020-12-04
fixing __call__ function

djarecka:
## Acknowledgment

## Types of changes

## Summary

fixing `__call__` function, so the `submitter` is not overwritten when state

## Checklist

- [ ] All tests passing

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -401,7 +401,7 @@ class TaskBase:
         plugin = plugin or self.plugin
         if plugin:
             submitter = Submitter(plugin=plugin)
-        elif self.state:
+        elif submitter is None and self.state:
             submitter = Submitter()
 
         if submitter:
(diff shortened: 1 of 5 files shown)


satra:
i know this fixes the current issue, but perhaps we could extend the plugin support a bit. we could add `plugin_kwargs` to the call signature and to wherever we set `self.plugin`

```
def __cal__(..., plugin_kwargs={}, **kwargs):
    ...
    plugin_kwargs = plugin_kwargs or self.plugin_kwargs
    if submitter is None:
        submitter = Submitter(plugin=plugin or 'cf', **plugin_kwargs)
    ...
```

djarecka:
for some reason, we didn't always want to create a new submitter, if nothing is provided... but will check it

satra:
i didn't pay attention to the full logic, so this should be better.

```
if submitter is None and self.state or plugin:
```

djarecka:
I actually want to revisit why we raise an error [here](https://github.com/nipype/pydra/pull/388/files#diff-3b85fb775d1f906f1a06277170afbaa573e7fc2c4701d138e79efe7fef229bc5L412), can't really remember why we don't handle this

djarecka:
@satra - are you ok with merging this?

Issue #387, opened 2020-12-03
output_spec field types not implemented


chasejohnson3:
What version of Pydra are you using?
0.12

What were you trying to do?
Pass a parameter of type Directory (or anything other than File) into the output_fields of an output_spec for a pydra task.

What did you expect will happen?
I expect to make fields in an output_spec with varying types. In particular: Int, Float, File, Bool, Str, Directory, and List.

What actually happened?
When all output_spec fields are of the type File, the task runs as expected. When any other type is introduced, the task stops and the following error message is shown.
`Error: “Exception: not implemented (collect_additional_output)”`

Can you replicate the behavior? If yes, how?
Yes, the following task is an adaptation to the pydra tutorial 5_intro_shelltask and throws the error
 `Error: “Exception: not implemented (collect_additional_output)”`

```
import nest_asyncio
nest_asyncio.apply()
import pydra
import attr

cmd = "echo"
args = ["newfile_1.txt", "newfile_2.txt"]

my_output_spec = pydra.specs.SpecInfo(
    name="Output",
    fields=[
        (
            "out1",
            attr.ib(
                type=pydra.specs.File,
                metadata={
                    "output_file_template": "{args}",
                    "help_string": "output file",
                },
            ),
        ),
        (
            "outDir",
            attr.ib(
                type=pydra.specs.Directory, # Causes error!
                metadata={
                    "output_file_template": "output_directory",
                    "help_string": "The directory for output to be written",
                },
            ),
        ),
    ],
    bases=(pydra.specs.ShellOutSpec,),
)

shelly = pydra.ShellCommandTask(name="shelly", executable=cmd, args=args, output_spec=my_output_spec).split("args")

print("cmndline = ", shelly.cmdline)

with pydra.Submitter(plugin="cf") as sub:
    sub(shelly)
shelly.result()
```

How I found the error:
I was running a pydra task generated from a Slicer Execution Model for [BRAINSToolsBrainsConstellationDetector](https://github.com/BRAINSia/BRAINSTools/tree/master/BRAINSConstellationDetector). One of the output fields had type pydra.specs.Directory and caused the error. The task ran to completion when this field was removed from the output_spec (all other output fields had type pydra.specs.File).

Location of the error:
In pydra/pydra/engine/specs.py in class ShellOutSpec under collect_additional_outputs. 
there is a conditional statement where only types File and MultiOutputFile are currently supported as output types.
```
if fld.type in [File, MultiOutputFile]:
        ...
else:
[...]

PR #389, opened 2020-12-08
Added allowed types for an output_spec field

chasejohnson3:
## Acknowledgment

## Types of changes

## Summary
The only types allowed for an output_spec field before this PR were File and MultiOutputFile. This PR adds Directory, Int, Float, Bool, Str, and List to the types allowed for an output_spec field. If the type of the field is Int, Float, Bool, or Str, the field "callable" is required in the metadata to process stdout or an output file and return an object of one of these types.
This PR addresses issue #387

## Checklist
- [ ] All tests passing


pydra/engine/specs.py
```diff
@@ -444,7 +464,16 @@ def collect_additional_outputs(self, inputs, output_dir):
...
-                if fld.type in [File, MultiOutputFile]:
+                if fld.type in [
+                    File,
+                    MultiOutputFile,
+                    Directory,
+                    Int,
+                    Float,
+                    Bool,
+                    Str,
+                    List,
```
satra, line 475: i think you should be able to use python types for this. i don't think you need special classes. did you try the Python types?

chasejohnson3: Good point - I was following the existing pattern in which a class for File was defined, but I do think it would be better to use Python types. I will make the change and update the PR

chasejohnson3:
@satra The types have been converted to python types, where applicable.

djarecka:
@chasejohnson3 -thanks! I think it would be nice to add some tests with `callable` to see how it works.

also, I've restarted slurm tests, they sometimes can randomly fail :(

chasejohnson3:
@djarecka I am working on adding adding tests with callable and editing specs.py to make stdout available to a callable. Is there an existing test or code sample in which stdout is accessed before the output for a task is returned?

chasejohnson3:
@djarecka @satra I added a test that uses both stderr and stdout as inputs to a callable. I accessed stderr and stdout by adding the parameter outputs to collect_additional_outputs() in specs.py. All tests and checks are passing, but take a look to make sure this implementation seems robust.

pydra/engine/tests/test_shelltask.py
```diff
@@ -2837,6 +2837,96 @@ def test_shell_cmd_outputspec_6a(tmpdir, plugin, results_function):
...
+                "out_file_index",
+                attr.ib(
+                    type=int,
+                    metadata={"help_string": "output file", "callable": get_file_index},
+                ),
+            ),
+            (
+                "stderr_field",
+                attr.ib(
+                    type=int,
```
satra, line 2881:
```suggestion
                    type=str,
```

djarecka:
@chasejohnson3 - thank you so much for your contribution! Please consider adding your name to the [zenodo file](https://github.com/nipype/pydra/blob/master/.zenodo.json). You could also just send me your info and I'll add it

chasejohnson3:
@djarecka Thank you for the feedback to my pull request successful. It doesn't look like I have direct write access to the zenodo file, so here is my info:
{
    "affiliation": "University of Iowa",
    "name": "Johnson, Charles E.",
    "orcid": "0000-0001-7814-3501"
}

djarecka:
@chasejohnson3 - Thanks!

PR #390, opened 2020-12-09
checksum_states improvements

djarecka:
## Acknowledgment

## Types of changes

## Summary

- improving `checksum_states` so it doesn't have to calculate the content hash of big files for every single element of the task state (function used in `result`), using `files_hash` that keeps track of all the files for the task (combining files from different state elements)

## Checklist

- [ ] All tests passing

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -274,7 +274,16 @@ class TaskBase:
                     key.split(".")[1],
                     getattr(inputs_copy, key.split(".")[1])[ind],
                 )
+            # setting files_hash again in case it was cleaned by setting specific element
+            # that might be important for outer splitter of input variable with big files
+            # the file can be changed with every single index even if there are only two files
+            inputs_copy.files_hash = self.inputs.files_hash
             input_hash = inputs_copy.hash
+            # updating self.inputs.files_hash, so big files hashes
+            # doesn't have to be recompute for the next element
+            for key, val in inputs_copy.files_hash.items():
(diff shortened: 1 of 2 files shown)


satra:
i do think this is where you would use a separate LRU cache, so you can reuse hashes across tasks. so i would recommend implementing a more general solution to this.

djarecka:
so you want to save hashes in files?

satra:
easiest may indeed be a single "database" file in the cache directory that can be protected against concurrent writes by softlock. but would be harder to maintain an LRU cache since items would need to be shuffled.

the alternative would be to create a BaseSpec class attribute that is set before anything is initialized. i don't know the implications for pickling, how to achieve concurrent/async updates etc.,. but i suspect this must have been looked at by people in the community.

```
class LRUwrapper:
    cache = None
    def __repr__(self):
         return str(cache)

class BaseSpec:
     filecache = LRUwrapper()
```

djarecka:
it depends how many things we want to fix with this PR. The part I was addressing here is run by the main node.

satra:
@djarecka - it's fine for this PR to fix the present issue, but i would suggest at least filing an issue for improving hashing efficiency

satra (approved):

djarecka:
@satra - oh, ok. I was double checking this on openmind first and was planning to try to follow your suggestion, but it might be better in a new PR that modifies all places that use hash. will open an issue

PR #401, opened 2021-01-12
using asyncio.SelectorEventLoop() to get a loop

djarecka:
## Acknowledgment

## Types of changes

## Summary

I was getting error from `asyncio.get_event_loop()` when pydra was run after `datalad` (also uses `asyncio`).

## Checklist

- [x] All tests passing

### DIFF
--- a/pydra/engine/helpers.py
+++ b/pydra/engine/helpers.py
@@ -649,8 +649,16 @@ def get_open_loop():
     if os.name == "nt":
         loop = asyncio.ProactorEventLoop()  # for subprocess' pipes on Windows
     else:
-        loop = asyncio.SelectorEventLoop()
-    asyncio.set_event_loop(loop)
+        try:
+            loop = asyncio.get_event_loop()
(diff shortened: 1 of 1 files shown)


mgxd:
Hey @djarecka, what specific error were you getting? Do you have a minimal reproducible example?

djarecka:
@mgxd - here you have description: https://github.com/datalad/datalad/issues/5313

haven't checked yet if python 3.8 solves it

Do you remember why you used `Selectors` only for windows?

mgxd:
I believe at the time the default loop was only `SelectorEventLoop`, so wanted to ensure Windows was getting the correct implementation. Though it seems asyncio's defaults should handle this fine in at least 3.9

https://docs.python.org/3/library/asyncio-eventloop.html#event-loop-implementations

djarecka:
I haven't removed the check for the OS, still using `ProactorEventLoop` for windows

djarecka:
at the end I decided to bring back `get_event_loop`, but just added `try/except` block and in case of error I just use `new_event_loop` instead of using selector directly.

PR #422, opened 2021-02-20
RF: Simplify positioning

effigies:
## Acknowledgment

## Types of changes

- [x] Refactor

## Summary
Started looking through #421 but didn't have time to comment before merge. As I tried to understand what was going on, felt moved to simplify some things.

The `position_adjustment` function worked but was not obviously correct from a quick read, so I rewrote to be very explicit. I've checked and it's slightly faster than the old if `None` is present, slightly slower if not (~2% runtime change in either direction). Also, sorting in-place is not generally expected of a sorting function, so `position_sort` does not modify the input. One change to functionality is that `position_sort` does not concatenate the lists of inputs, so it now works for arbitrary objects, not just lists of objects, and we concatenate at the end.

## Checklist

- [x] All tests passing

### DIFF
--- a/docs/input_spec.rst
+++ b/docs/input_spec.rst
@@ -129,8 +129,8 @@ In the example we used multiple keys in the metadata dictionary including `help_
    If no `argstr` is used the field is not part of the command.
 
 `position` (`int`):
-   Position of the field in the command, could be positive or negative integer.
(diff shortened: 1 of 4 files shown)


djarecka:
looks great, I've actually never used `bisect`... ready to merge?

effigies:
Merge away!

PR #424, opened 2021-02-20
STY: Simplify output_spec definition and population

effigies:
## Acknowledgment

## Types of changes

- [x] Style

## Summary
The goal is readability. Rather than repeatedly create `SpecInfo` in the branches, set some default parameters, then see if we should update any. Finally, create the `SpecInfo` from the selected parameters.

Also some cases where `if/else` branches could be turned trivially into `if/elif/else` chains.

## Checklist

- [x] All tests passing

### DIFF
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -149,47 +149,31 @@ class FunctionTask(TaskBase):
             rerun=rerun,
         )
         if output_spec is None:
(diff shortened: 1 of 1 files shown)


djarecka:
@effigies lgtm!

slurm has some tests that are failing from time to time, so I rerun it.

PR #426, opened 2021-02-22
ENH: Allow attr_fields to exclude fields

effigies:
## Acknowledgment

## Types of changes

- [x] Refactor

## Summary

It is fairly common for calls to `attr_fields` to be followed by a filter for specific names that get ignored. This adds an exclusion list to the call to simplify the loops.

## Checklist

- [ ] All tests passing

### DIFF
--- a/pydra/engine/specs.py
+++ b/pydra/engine/specs.py
@@ -8,8 +8,8 @@ import re
 from .helpers_file import template_update_single
 
 
-def attr_fields(x):
-    return x.__attrs_attrs__
+def attr_fields(spec, exclude_names=()):
+    return [field for field in spec.__attrs_attrs__ if field.name not in exclude_names]
 
 
 class File:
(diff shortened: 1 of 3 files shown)


pydra/engine/specs.py
```diff
@@ -445,47 +435,36 @@ class ShellOutSpec:
...
-                                fld, inputs, output_dir, outputs
-                            )
-                #                        else:
-                #                            additional_out[fld.name] = self._field_metadata(
-                #                                fld, inputs, output_dir, outputs
-                #                            )
-                else:
-                    raise Exception("not implemented (collect_additional_output)")
+        for fld in attr_fields(self, exclude_names=("return_code", "stdout", "stderr")):
+            if fld.type not in [
```
djarecka, line 439: i'm actually not sure if we should be checking this here.. but can think about this in a different pr

djarecka:
lgtm

PR #432, opened 2021-02-26
[TMP, do not merge] adding example for task with FuctionImage

djarecka:
## Acknowledgment

## Types of changes

## Summary

showing example with `FunctionalImage` as an input - print statements show when the hash value changes 

`test_5` is the one that failing

## Checklist

- [ ] All tests passing

### DIFF
--- a/pydra/engine/helpers.py
+++ b/pydra/engine/helpers.py
@@ -696,8 +696,19 @@ def hash_value(value, tp=None, metadata=None, precalculated=None):
             return hash_dir(value, precalculated=precalculated)
         elif type(value).__module__ == "numpy":  # numpy objects
             return sha256(value.tostring()).hexdigest()
-        else:
+        elif (
+            isinstance(
+                value, (int, float, complex, bool, str, bytes, LazyField, os.PathLike)
+            )
+            or value is None
+        ):
             return value
+        else:
+            warnings.warn(
+                f"pydra doesn't fully support hashing for {type(value)}, "
+                f"cp.dumps is used in hash functions, so it could depend on the system"
+            )
+            return sha256(cp.dumps(value)).hexdigest()
 
 
 def output_from_inputfields(output_spec, input_spec):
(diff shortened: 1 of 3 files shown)


satra:
this may help track down the issue. while the functionality of an object is preserved through a roundtrip through cloudpickle, the bytestring is not preserved.  so if the test that is failing is saving the task to file and loading it, it would likely get a different hash. 

```python
In [15]: def func1(a): 
    ...:     return a + 1 
    ...: class foo: 
    ...:     def test(self, b): 
    ...:         return func1(b) 
    ...:                                                                                                

In [16]: ff = foo()                                                                                     

In [17]: sha256(cp.dumps(ff)).hexdigest()                                                               
Out[17]: '9dcd775519701de26beab29199ef7259657c406caf5af36186192a0fa47967d4'

In [18]: sha256(cp.dumps(deepcopy(ff))).hexdigest()                                                     
Out[18]: '9dcd775519701de26beab29199ef7259657c406caf5af36186192a0fa47967d4'

In [19]: sha256(cp.dumps(cp.loads(cp.dumps(ff)))).hexdigest()                                           
Out[19]: 'b85dbb27044c14651596da8eaba9818cfa6351162cc296b2bc1df571891b35c2'

In [20]: ff.test(1)                                                                                     
Out[20]: 2

In [21]: cp.loads(cp.dumps(ff)).test(1)                                                                 
Out[21]: 2
```

satra:
indeed, we should not rely on pickle. 

you could consider using the `__dict__` or `__getstate__` of the object if available. or something else about the object. as far as i know there is no generic solution in python for this. 

in the short term we could issue exceptions and/or force those tasks to always run.

satra:
here is a related post (but old - some things in python have changed with 3.x): https://stackoverflow.com/questions/16157835/create-hash-for-arbitrary-objects/

djarecka:
@satra - I actually did test what happens when I create a `deepcopy` of this specific object and I was able to get the same hash value.

PR #434, opened 2021-03-01
WIP: Add pydra.tasks.core tasks

effigies:
## Acknowledgment

## Types of changes

## Summary

Starting off with `Merge`.

## Checklist

- [ ] All tests passing


pydra/tasks/core/sequences.py
```diff
@@ -0,0 +1,117 @@
...
+        self._numinputs = max(numinputs, 0)
+        self.input_spec = SpecInfo(
+            name="Inputs",
+            fields=[(f"in{i + 1}", ty.List) for i in range(self._numinputs)],
+            bases=(MergeInputSpec,),
+        )
+        # Can't set as class attribute
+        self.output_spec = SpecInfo(
+            name="Outputs", fields=[("out", ty.List)], bases=(BaseSpec,)
+        )
```
effigies, line 91: I was surprised by this constraint. Setting it as a class attribute caused this to fail with a `KeyError`:

https://github.com/nipype/pydra/blob/b934e4b549fa4e16cdffff844dcf655cdf0a4c13/pydra/engine/core.py#L199-L202

pydra/tasks/core/sequences.py
```diff
@@ -0,0 +1,115 @@
+import attr
+import typing as ty
+from pydra.engine.specs import BaseSpec, SpecInfo
+from pydra.engine.core import TaskBase
+from pydra.engine.helpers import ensure_list
+
+
+@attr.s(kw_only=True)
+class MergeInputSpec(BaseSpec):
+    axis: ty.Literal["vstack", "hstack"] = attr.ib(
```
effigies, line 10: Fails on Python 3.7, since `Literal` was added in 3.8. Not sure if we have another way of doing this or want to monkey-patch it into 3.7, as in https://github.com/python/typing/blob/b5c0b6d66b3adba4dde3f4f4145ce83a70fd03f7/typing_extensions/src_py3/typing_extensions.py#L543-L676.

satra: in some projects we use `sys.version_info` to pull `Literal` in for 3.7

https://github.com/dandi/dandi-cli/blob/master/setup.cfg#L56
https://github.com/dandi/dandi-cli/blob/master/dandi/models.py#L22

effigies: Ah, `typing_extensions`. I tried it with just `typing` and broke pip.

pydra/tasks/core/sequences.py
```diff
@@ -0,0 +1,197 @@
...
+    input_spec = SplitInputSpec
+
+    def __init__(self, splits, *args, **kwargs):
+        self.output_spec = SpecInfo(
+            name="Outputs",
+            fields=[(f"out{i + 1}", list) for i in range(len(splits))],
+            bases=(BaseSpec,),
+        )
+        super().__init__(*args, **kwargs)
+        self.inputs.splits = splits
```
effigies, line 55: Right now no way to update `output_spec` if `splits` is computed and passed via a workflow. I'm not sure we could determine the output names in order to connect the outputs of this task if the number of splits is not known.

We could instead do something like `max_splits` and create all `out{i}` fields with the understanding that anything depending on an output that is not generated will receive a `None` (or `attr.NOTHING`...)

effigies:
Before moving on to IO tasks, which I suspect will take longer, I think it would be worth reviewing this for whether this is how we want to model writing tasks, as well as how we want to expose these. For instance, do we want to import all into `pydra.tasks.core` or keep them in their more topical subpackages.

Also, no attachment to `sequence`. Just avoiding keyword `list`.

pydra/tasks/core/sequences.py
satra: i'm not following this i think. a task writer should not concern themselves with splits or combines. that's outside of the task's responsibility. it should simply take whatever input is given and work with it.

satra:
i have another question. this PR does mimic nipype, but does pydra need to fully mimic nipype for function tasks. 

is there a reason why these are not functions with annotations? for pure python i would like to be as close to python as possible. in pydra-ml i took this approach (completely open to discussions).

python functions that can be tested as normal (should be annotated, which i didn't do when i wrote that and forgot).  https://github.com/nipype/pydra-ml/blob/master/pydra_ml/tasks.py

and converted to pydra tasks: https://github.com/nipype/pydra-ml/blob/master/pydra_ml/classifier.py#L32

pydra/tasks/core/sequences.py
effigies: The task splits a list. It has nothing to do with splitting/combining.

effigies:
> is there a reason why these are not functions with annotations?

`Merge()` has a variable set of inputs. `Split()` has a variable set of outputs.

pydra/tasks/core/sequences.py
satra: semantics are hard! :)

satra:
i have forgotten a lot of things :)

pydra/tasks/core/sequences.py
djarecka: > Right now no way to update `output_spec` if `splits` is computed and passed via a workflow. I'm not sure we could determine the output names in order to connect the outputs of this task if the number of splits is not known.
This is not exactly what you want, but we have also option to pass all outputs, by using `lzout.all_`: https://github.com/nipype/pydra/blob/master/pydra/engine/tests/test_workflow.py#L3662

djarecka: But I have to say that I'm not convinced to this specific task, I believe users could always write an easy python function

djarecka:
I left a comment in the review section, but I'm also not convince why we want to have a special task `Split`...

pydra/tasks/core/sequences.py
satra: sorry to derail the thread, but hopefully we can get back on track. i do think the notion of dynamic named outputs and inputs cannot be solved with python functions. i had indeed forgotten about the details of split and merge.

@djarecka - can you give an example of how you can create N arbitrary splits and feed the different components of the splits to different downstream tasks with a pure python function?

djarecka: so there are different subsequent functions and arbitrary number of splits? Is it really a valid use case? How do you create a workflow? Does the `Split` answer this?

satra: in nipype splits creates a named output for the different numbers of splits. an example is the following:

let's say a subject has M T1s and N T2s and i want them to be processed together. i can create a Merge node to aggregate them (merge also produces the relevant output to split them apart), do the processing and then split them back into 2 components (a list of processed T1s and a list of processed T2s. as a workflow designer i know that there are two categories here, what i don't ahead of time is the number. hence being able to set split would be required.

another thought here is that the outputs would come from `lzout`, if we can't find an attribute at construction time we can check that attribute at runtime. it may still error out, but we should be able to evaluate that connection. @effigies - would that address your question, without having to define max_splits?

effigies: Yes, I think if we could have static and dynamic output specs, then that could work. In the static case we would do name checking, and in the dynamic case we leave it to post-run.

In the specific case of `Split()`, you need to know how many splits you have, even if you don't know the specific number in each split. We could either do that dynamically by connections or we could do it statically by declaring a number of splits.

djarecka: so if you know the number of splits in `Split()` you could do it with normal python function. Could someone give me a simple workflow when this could be used. I'm probably missing something important, since I still don't see where I'd like to use it.

effigies: I don't specifically care about `Split()`. I've literally never used it. Nonetheless the point is that there are numerous Nipype interfaces where the full set of inputs or outputs isn't known until the interface is instantiated. The `Merge()` example in this PR is one such that I have used and it's easier to write:

```
merge3 = Merge(n=3)
merge4 = Merge(n=4)
```

Than:

```
@pydra.mark.task
def merge3(in1, in2, in3):
    return in1 + in2 + in3

m3 = merge3()

...
```

Similarly, to do the same thing with `Split()` would involve repeatedly writing:

```
@pydra.mark.task
def custom_split(inlist, splits) -> {"out1": list, "out2": list, "out3": list}:
    # ...
```

djarecka: ok, so it is design to mimic these simple python function, but we just want to allow people to use it instead of creating annotation. I'm fine with this, but I'm a bit afraid of using the name `Split`. Perhaps `SplitTask` and `MergeTask`? I know it's longer, but having `split` and `Split` - one used as a method, the other as a class can confuse people IMO

tclose:
these could be added to `pydra.tasks.common`

effigies:
Needs to be redone in new syntax.

Issue #444, opened 2021-03-29
Possible deadlock when splitting at workflows


jw-96:
I found a way to create a deadlock in Pydra.

Pydra version: 0.14.1, 
Python version: 3.9.2
In a jupyter notebook.

I was working on translating this [tutorial](https://miykael.github.io/nipype_tutorial/notebooks/example_preprocessing.html) into Pydra. For which I created two workflows an inner and an outer one. The outer workflow has a splitter on three arguments. When trying to run the workflow, I noticed how it starts out with all 8 virtual cores of my machine being utilized 100%, but after a few minutes, they were all at idle again without the cell having finished executing. Upon closer observation, I noticed that they seem to go idle in pairs, which made me think of a deadlock. 

The output when stopping the cell showed a process getting interrupted when trying to acquire a lock, which confirmed my suspicion. 

I replicated the deadlock in a smaller example program because the tutorial consists of many nodes and needs a dataset. 
Together with the output, it is available [here]( https://github.com/neurorepro/pydra-glm-example/blob/main/other/Deadlock.ipynb ).

```Python
import nest_asyncio
nest_asyncio.apply()

import pydra

task_list = ["First", "Second"]
start_list = [1,2,3,4,5,6,7,8]

@pydra.mark.task 
def one_arg(start_number):
    print(start_number)
    for k in range(100000000):
        start_number += 1
    return start_number

@pydra.mark.task 
def two_num(a, b):
    print(a, b)
    a = a + b
    for k in range(100000000):
        a += 1
    return a


@pydra.mark.task 
def fin_node(a, b, task_name):
    print(a, b, task_name)
    return a+b

# Outer workflow
test_outer = pydra.Workflow(name="test_outer", input_spec=["start_number", "task_name"])
# Splitting on both arguments
test_outer.split(["start_number", "task_name"], start_number = start_list, task_name = task_list)

# Inner Workflow
test_inner = pydra.Workflow(name="test_inner", input_spec=["start_number1"])
test_inner.add(one_arg(name = "Ilevel1", start_number = test_inner.lzin.start_number1))
test_inner.set_output([("res", test_inner.Ilevel1.lzout.out)])



# Outer workflow has two nodes plus the inner workflow
test_outer.add(one_arg(name = "level1", start_number = test_outer.lzin.start_number))
test_outer.add(test_inner)
test_inner.inputs.start_number1 = test_outer.level1.lzout.out




# The second node takes as input the output of the first node and the inner workflow
[...]

PR #460, opened 2021-04-21
SoftFileLock asyncio workaround (#444)

jw-96:
## Acknowledgment

## Types of changes

## Summary
The SoftFileLock does not work with asyncio. This implements the workaround suggested at [#78](https://github.com/benediktschmitt/py-filelock/issues/78).

## Checklist

- [x] All tests passing

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -10,7 +10,7 @@ from copy import deepcopy
 from uuid import uuid4
 
 import cloudpickle as cp
-from filelock import SoftFileLock, Timeout
+from filelock import SoftFileLock
 import asyncio
 import shutil
 from tempfile import mkdtemp
@@ -37,6 +37,7 @@ from .helpers import (
     ensure_list,
     record_error,
     hash_function,
+    PydraFileLock,
 )
 from .helpers_file import copyfile_input, template_update
 from .graph import DiGraph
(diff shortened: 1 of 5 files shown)


satra:
@jw-96 thanks for tracking this down. to keep things similar could you instead add a contextual option

```
class PydraFileLock:

    def __init__(self, lockfile):
        self.lockfile = lockfile

    def __enter__(self):
        lock = SoftFileLock(self.lockfile)
        acquired_lock = False
        while not acquired_lock:
                lock.acquire(timeout=0)
                acquired_lock = True
            except Timeout:
                await asyncio.sleep(1)
        self.lock = lock   
        return self

    def __exit__(self, exc_type, exc_value, traceback):
        self.lock.release()
        return None
```

you could add this to helpers, and then this becomes a drop in replacement for SoftFileLock wherever it is used.

jw-96:
@satra I tried implementing your proposed change. However, I got this error:
```
SyntaxError: 'await' outside async function
```
I also tried the following: 
```
class PydraFileLock:
    """Wrapper for filelock's SoftFileLock that makes it work with asyncio."""

    def __init__(self, lockfile):
        self.lockfile = lockfile

    async def __enter__(self):
        lock = SoftFileLock(self.lockfile)
        acquired_lock = False
        while not acquired_lock:
            try:
                lock.acquire(timeout=0)
                acquired_lock = True
            except Timeout:
                await asyncio.sleep(1)
        self.lock = lock
        return self

    def __exit__(self, exc_type, exc_value, traceback):
        self.lock.release()
        return None
```
However, it resulted in both tasks entering the critical section.
I also added mine and @djarecka test.

effigies:
Is it easy to identify when we're in an async context? If so, maybe what we want is a factory:

```Python
class AsyncSoftFileLock:
    ...
    async def __enter__(...):
        ...

def get_file_lock(filename):
    if async:
        return AsyncSoftFileLock(filename)
    return SoftFileLock(filename)
```

jw-96:
Should we use pytest-timeout to automatically stop the test after 5 seconds? Currently I am only checking if the results are correct, but if the deadlock would reoccur in the future, then the testing could get stuck.

satra:
if this works, let's go with this. i realized there are more complications (we use softlock in both an async and a non async setting. 

i'm now curious why the change was not needed in the workflow run class which also has a softlock. perhaps we should test this out where the subtask is a workflow as well.

mgxd:
it's been awhile since I took a look at the code, so as a refresher for myself:
- We are sync within `Task._run` (and a lock is being created there)
- We are async within `Workflow._run` (also creates a lock)

I think the idea (at the time) behind this was that `Task`s were singular commands whereas `Workflow`s needed to be expanded and waited on.

djarecka:
`Task._run` is run in [exec_as_coro](https://github.com/nipype/pydra/blob/master/pydra/engine/workers.py#L160)

jw-96:
@satra I was able to get it working with [__aenter__](https://www.python.org/dev/peps/pep-0492/#asynchronous-context-managers-and-async-with). I also added timeouts to the tests.

satra:
@jw-96 - thank you. closed my PRs and just triggered this to run. the tests are taking much longer. so we may want to consider reducing the sleep time. let's see how long the tests are relative to the other branches.

jw-96:
I noticed the same. I reduced the sleep time to 0.1 seconds and added exponential back-off capped at 4 seconds.

satra:
>  I reduced the sleep time to 0.1 seconds and added exponential back-off capped at 4 seconds.

did these help? you haven't pushed these yet right?

jw-96:
I only run ```pytest test_workflow.py::test_duplicate_input_on_split_wf```: 
with 1 second sleep: 11.55s
with 0.1 seconds + exp. backoff: 10.66s
I run it thrice and the run times only varied by ± 0.05s

satra:
i'm going to merge this. let's keep the exponential backup in tow in case that comes in handy.

satra:
actually before i merge. could you please add yourself to the .zenodo.json file?

pydra/engine/core.py
```diff
@@ -11,6 +11,7 @@
 
 import cloudpickle as cp
 from filelock import SoftFileLock
+import asyncio
```
satra, line 14: i believe this is not necessary since it is part of the default language specs now.

jw-96:
Is the orcid necessary? I do not have one.

pydra/engine/core.py
jw-96: Yes, it seems to be not necessary

satra:
the orcid is not necessary, but helps if you want tracking of contributions across different projects.

pydra/engine/helpers.py
```diff
@@ -893,3 +893,29 @@ def argstr_formatting(argstr, inputs, value_updates=None):
...
+    async def __aenter__(self):
+        lock = SoftFileLock(self.lockfile)
+        acquired_lock = False
+        while not acquired_lock:
+            try:
+                lock.acquire(timeout=0)
+                acquired_lock = True
+            except Timeout:
+                await asyncio.sleep(self.timeout)
+                if self.timeout <= 2:
```
djarecka, line 914: should we raise some meaningful exception here?

jw-96:
I removed the ```import asyncio``` and added myself to the .zenodo.json. I quickly made a orcid.

pydra/engine/helpers.py
satra: i think this is ok. it's just used to trap. this is within a polling loop.

djarecka: ok

jw-96:
Sorry forgot a comma in zenodo file. Is there a way to only have the latest commit running?

djarecka:
@jw-96 - don't worry. However, could I ask you for one small change? Could you please move your name before Satra. Thanks

jw-96:
Ok, like this?

djarecka:
yes, thanks a lot!

PR #467, opened 2021-05-21
[ENH] Changes to BoshTask

jw-96:
Translating more boutiques specifications to Pydra specs. Checking if output files exist.
## Acknowledgment

## Types of changes

## Summary
Adding support for some more bosh input specifications: support for default values, integer values, and mandatory inputs.
QoL: ```bosh_file``` input can now be a string, and ```zenodo_id``` can now also start with the "zenodo" prefix.

This pull request also changes how output files that do not exist are handled. Currently, the path to where the file is expected is returned regardless of the file existing. The change in this pull request will make it return ```attr.NOTHING``` if the file does not exist. If an output is marked as mandatory and the output file does not exist then an error will be raised. The motivation behind this change is that it is otherwise not immediately evident which files were created. For example, BET generates usually only a subset of the outputs, but the current behavior would show paths for all. 

The tests were updated to take this behaviour into account.
In the process I found that the [test](https://github.com/nipype/pydra/blob/f2106f62b5cd2273ef96a5aab69de94e413d9aed/pydra/engine/tests/test_shelltask.py#L3017) ```test_shell_cmd_outputspec_7d``` has an issue. It intends to test whether the directory named ```test``` has been created at ```resultsDir```. However, Pydra checks if the ```test``` directory was created in its current working directory, i.e., the cache directory. The test never failed because pydra returned the path to where it expected the directory to be and the test only checked if the name of the lowest directory matches.  The change of the pull request now has Pydra return ```attr.NOTHING```. I think this is a bug relating to how absolute paths in ```output_file_template``` are being interpreted. A possible solution could be to add a metadata key akin to boutiques' ```uses-absolute-path```.

Also, BoshTask already used mandatory outputs prior to this pull request. However, hey are not yet part of the [output specifications](https://pydra.readthedocs.io/en/latest/output_spec.html). Should it be added there?

## Checklist

- [X] All tests passing
- [X] I have added tests to cover my changes
- [X] My code follows the code style of this project

### DIFF
--- a/pydra/engine/boutiques.py
+++ b/pydra/engine/boutiques.py
@@ -19,6 +19,7 @@ class BoshTask(ShellCommandTask):
         zenodo_id=None,
         bosh_file=None,
         audit_flags: AuditFlag = AuditFlag.NONE,
+        bindings=[],
         cache_dir=None,
         input_spec_names: ty.Optional[ty.List] = None,
         messenger_args=None,
@@ -80,6 +81,7 @@ class BoshTask(ShellCommandTask):
         self.input_spec = self._prepare_input_spec(names_subset=input_spec_names)
         self.output_spec = self._prepare_output_spec(names_subset=output_spec_names)
         self.bindings = ["-v", f"{self.bosh_file.parent}:{self.bosh_file.parent}:ro"]
+        self.add_input_bindigs(bindings)
 
         super().__init__(
             name=name,
@@ -190,7 +192,10 @@ class BoshTask(ShellCommandTask):
         """creating output spec from the zenodo file
         if name_subset provided, only names from the subset will be used in the spec
         """
-        boutputs = self.bosh_spec["output-files"]
+
+        boutputs = self.bosh_spec.get("output-files", None)
+        if not boutputs:
+            return SpecInfo(name="Outputs", fields=[], bases=(ShellOutSpec,))
         fields = []
         for output in boutputs:
             name = output["id"]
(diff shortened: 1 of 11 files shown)


satra:
@jw-96 - do consider moving this into a different repo: `pydra-boutiques`. it would be great if pydra didn't have to be updated when boutiques changes. i think this was in pydra only because of ease, rather than the right place.

jw-96:
@satra I discussed the possibility of moving BoshTask to its own repo with @djarecka. We think it would be a good idea, especially since the Boutiques tests take quite a while. However, this PR also contains some changes that are not in ```BoshTask.py```:
- For non existent files the output is set to ```attr.NOTHING```.
  - This required some minor adjustments to some tests.
- Adding the ```absolute_path``` to the specifications. If set to true, the formatted template will not be added to the ```output_dir``` path. This change enables one to specify a path in ```output_file_template``` to where the output file will be found (This also enabled fixing ```test_shell_cmd_outputspec_7d``` ).

satra:
Sure. Anything that is relevant to pydra itself should stay.

djarecka:
@jw-96 - do you think it would be easy for you to separate the changes to pydra itself (and if possible, some test that do not requires boutiques should be added). In that case we could open a new repo right away, and add boutiques related changes there

jw-96:
@djarecka Yes, it should not be hard to separate them. Should I make a new PR for everything not related to boutiques?

djarecka:
that would be great, thank you!

djarecka:
I've also added you as a member to the organization if you want to start a new repository. If not, we can work on this on Thursday

jw-96:
Ok, thanks. I opened a new pull request. I probably won't have time today to open the new repo, but I can do it on Monday.

Issue #468, opened 2021-05-21
ANTS input lists


jw-96:
I was trying to implement the ANTS ```N4BiasFieldCorrection```. The command accepts the output names as a list. 
I tried to implement it the following way: 
```
from pydra.engine import specs
from pydra import ShellCommandTask
import typing as ty
from pydra.engine.core import TaskBase, is_lazy
from pydra.utils.messenger import AuditFlag

input_fields = [
    (
        "input_image",
        specs.File,
        {
            "help_string": "input for bias correction. Negative values or values close to zero should be processed prior to correction",
            "argstr": "--input-image {input_image}",
        },
    ),
    (
        "outputs",
        ty.List[str],
        {
         "help_string": "just here to gen the outpus",
         "argstr": "--output {outputs}",
        }
    ),
]
N4BiasFieldCorrection_input_spec = specs.SpecInfo(
    name="Input", fields=input_fields, bases=(specs.ShellSpec,)
)
class N4BiasFieldCorrection(ShellCommandTask):
    
    def __init__(
        self,
        name,
        audit_flags: AuditFlag = AuditFlag.NONE,
        cache_dir=None,
        messenger_args=None,
        messengers=None,
        output_cpath="/output_pydra",
        rerun=False,
        strip=False,
        **kwargs,
    ):
        input_spec = N4BiasFieldCorrection_input_spec
        output_spec = None
        executable = "N4BiasFieldCorrection"
        super().__init__(
                    name=name,
                    input_spec=input_spec,
                    output_spec=output_spec,
                    executable=executable,
                    audit_flags=audit_flags,
                    messengers=messengers,
                    messenger_args=messenger_args,
                    cache_dir=cache_dir,
                    strip=strip,
                    output_cpath=output_cpath,
                    rerun=rerun,
                    **kwargs,
                )
```

running the task
```
n4 = N4BiasFieldCorrection(name = "n4_pydra", input_image = "/home/jwigger/Documents/pydra-boutiques/data/test.nii.gz", outputs = ['test_corrected.nii.gz', 'test_bias.nii.gz'])
n4()
```
yielded the following output: 
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-9-cf2eeaf33ba7> in <module>
----> 1 n4()

~/miniconda3/lib/python3.9/site-packages/pydra/engine/core.py in __call__(self, submitter, plugin, plugin_kwargs, rerun, **kwargs)
[...]

PR #471, opened 2021-05-28
Changes to Input/Output Specs

jw-96:
This PR makes three changes: 
- For non existent files the output is set to attr.NOTHING.
  - This required some minor adjustments to some tests.
- Adding the absolute_path to the specifications. 
- Added an input callable named ```formatter```.

## Acknowledgment

## Types of changes

## Summary

This pull request also changes how output files that do not exist are handled. Currently, the path to where the file is expected is returned regardless of the file existing. The change in this pull request will make it return ```attr.NOTHING``` if the file does not exist. If an output is marked as mandatory and the output file does not exist then an error will be raised. The motivation behind this change is that it is otherwise not immediately evident which files were created. For example, BET generates usually only a subset of the outputs, but the current behavior would show paths for all. 

The tests were updated to take this behaviour into account.
In the process I found that the [test](https://github.com/nipype/pydra/blob/f2106f62b5cd2273ef96a5aab69de94e413d9aed/pydra/engine/tests/test_shelltask.py#L3017) ```test_shell_cmd_outputspec_7d``` has an issue. It intends to test whether the directory named ```test``` has been created at ```resultsDir```. However, Pydra checks if the ```test``` directory was created in its current working directory, i.e., the cache directory. The test never failed because pydra returned the path to where it expected the directory to be and the test only checked if the name of the lowest directory matches.  The change of the pull request now has Pydra return ```attr.NOTHING```. 

The second feature adds a new input/output spec called ```absolute_path```. If set to true, the formatted template will not be added to the output_dir path. This change enables one to specify a path in output_file_template to where the output file will be found (This also enabled fixing test_shell_cmd_outputspec_7d).

The final feature is an input callable ```formatter```. It is used to provide a custom argstr for an input. It works similar to the output callable. It can take as arguments all inputs. There are several use cases for this feature. For example it can be used to compose and command line argument out of several inputs (this is for example used in nipype for N4BiasFieldCorrection).

## TODO:
- [x] Test currently fail because the changes to the boutique tests were not moved to this PR -> delete BoshTask and its test?
- [x] update documentation
- [x] write more tests
- [x] add input callable suggested in #468
## Checklist

- [x] All tests passing
- [X] I have added tests to cover my changes
- [X] My code follows the code style of this project


pydra/engine/tests/test_shelltask_inputspec.py
```diff
@@ -2073,3 +2073,412 @@ def test_shell_cmd_inputs_di(tmpdir, use_validator):
...
+        executable="touch",
+        input_spec=input_spec,
+        output_spec=out_spec,
+        out_name=[str(Path(tmpdir) / "test_1.nii"), "test_2.nii"],
+    )
+    shelly()
+    res = shelly.result()
+    # checking if the outputs are Nothing
+    assert res.output.out_list[0] == Path(tmpdir) / "test_1.nii"
+    assert res.output.out_list[1] == attr.NOTHING
```
djarecka, line 2297: I'm not sure if I wouldn't expect an error here

djarecka: i think it just might be very misleading, and I'm not sure if they are any good  advantages

jw-96: So, a MultiOutput should raise an error if one of the outputs is not existing, or only if it set to mandatory ?

djarecka: I'm not sure if this is about `MultiOutputPath`
perhaps I don't understand something more general. I thought that "absolute_path" should give you an option to save something outside the working directory if you provide an absolut path.
In this test, both files are created, but the second one doesn't have proper path (so it's treated as the relative path).
So perhaps we should inform the users that something is wrong with path instead of creating the file, but not sending it to the output

pydra/engine/tests/test_shelltask_inputspec.py
```diff
@@ -2073,3 +2073,412 @@ def test_shell_cmd_inputs_di(tmpdir, use_validator):
...
+
+
+def test_shellspec_formatter_1(tmpdir):
```
djarecka, line 2300: we could use a `pytest.mark.parametrize` to split this test

djarecka: btw. some of the test could be also represented by `readonly` option, as in `test_shell_cmd_inputs_di`, but we discussed about it. Your approach is more general

pydra/engine/tests/test_shelltask_inputspec.py
jw-96: With the second file I am testing that if ```absolute_path``` is specified that a file in the cache directory is not returned, i.e., I am double checking that the provided file path (here just 'test_2.nii') is not appended to the path of the nodes output directory. I am just creating the file because the name needs to be in the ```MultiInputObj``` such that it can be searched by the ```MultiOutputFile```.

pydra/engine/tests/test_shelltask_inputspec.py
jw-96: Yes, another use case to test could be the following. An application created either an output with extension '.nii.gz' or '.nii' depending on a flag. So, the ```formatter``` could be used to create the right template based on the flag.

pydra/engine/tests/test_shelltask_inputspec.py
djarecka: but the only reason why the second file is not appended to the output directory is because `test_2.nii'` is not the correct absolute path. Am i getting this right?

jw-96: Yes, no file exists at that location.

djarecka:
i'm merging it, but we can have more discussion later on when to raise error

djarecka:
I'm reading this PR again, since there are things that stopped working, i.e. [dcm2niix](https://github.com/nipype/pydra-dcm2niix/blob/main/pydra/tasks/dcm2niix/utils.py). 

You can see that `out_dir` is used in several output fields. Since `out_dir` is an input it has to be provided as an absolut path. Before `pydra` was able to find files and copy them to the `output_dir` (so you had the files in the user defined `out_dir` and in the task dir). Now it can't find it, unless each of the output field has `absolute_path==True`. Was this your intention?

jw-96:
Yes, that was the intention. If the output is outside the cache directory of that task, then it should have ```absolute_path==True```.

PR #477, opened 2021-06-14
[fix] fixing cmdline for shell tasks with templates and spliter (continuation of #475)

djarecka:
## Acknowledgment

## Types of changes

## Summary

This is continuation of #475: didn't notice that the cmdline was not correct (thanks @jw-96)

This is a quick fix, should be rewritten at some point....

## Checklist

- [ ] All tests passing

### DIFF
--- a/pydra/engine/helpers_file.py
+++ b/pydra/engine/helpers_file.py
@@ -525,46 +525,60 @@ def copyfile_input(inputs, output_dir):
 
 
 # not sure if this might be useful for Function Task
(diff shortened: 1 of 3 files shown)


djarecka:
@jw-96 - the cmdline was indeed incorrect, created this pull request to fix it. I also added a test that checks the cmdline form itself.

This is not the best fix, but more general changes I tried were failing, and I can't work on them today. Hopefully this will work for your example too.

jw-96:
Ok, thanks. I will use the same fix for the input callable.

PR #493, opened 2021-08-31
Ci/ci and pre-commit updates

### LINKED ISSUE #478: doc building failing on circle ci
doc building is failing. 

one thing to consider is to just build it in a github action workflow as well. this is not essential.

### LINKED ISSUE #484: cleaning up CI and release processes
- [ ] make the style test be not run for every python env. so create a separate github workflow
- [ ] update precommit to include other linting components (see dandischema repo for an example)
- [ ] move to auto for releases (also see dandischema for a simple release workflow - however we need to fix a bit there that the twine upload is only tried if auto creates a new release)

potentially consider connecting to pre-commit.ci (dandischema has implemented this, which auto-checks and commits even if a dev has forgotten to do so)

djarecka:
## Acknowledgment

## Types of changes

## Summary

fixes #484 fixes #478

- Updating GA workflows:
    - some changes for styles check
    - adding `release.yml`
- adding more checks to pre-commit
- fixing doc builds

**all changes to the code were due to flake/codespells complaints...**

## Checklist

- [ ] All tests passing


pydra/engine/workers.py
```diff
@@ -356,13 +356,7 @@ class DaskWorker(Worker):
     def __init__(self, **kwargs):
         """Initialize Worker."""
         super().__init__()
-        try:
-            from dask.distributed import Client
-        except ImportError:
-            logger.critical("Please instiall Dask distributed.")
-            raise
```
satra, line 361: this import is in init so it gets caught at the point of initialization of the worker, not when the worker is executed, which is an async function.

satra: this move seems to be done for styling rather than code. you can mark something to be ignored by styling if that is a problem.

djarecka: ok, will work on flask settings, not on the code

PR #499, opened 2021-09-24
Rf/state

djarecka:
## Acknowledgment

## Types of changes

## Summary

State class refactoring:
- moving some functions to the `State` class
- fixing/refactoring inner_splitter parts
- fixing/refactoring container_dimensions part  
- changes in State's properties (avoiding multiple rerunning)

## Checklist

- [ ] All tests passing


pydra/engine/tests/test_workflow.py
```diff
@@ -1403,52 +1454,198 @@ def test_wf_3nd_ndst_6(plugin, tmpdir):
...
+
+
+def test_wf_3nd_ndst_8(plugin, tmpdir):
+    """workflow with three tasks, third one connected to two previous tasks,
+    using default splitter in the third task
+    """
+    wf = Workflow(name="wf_ndst_9", input_spec=["x"])
+    wf.add(add2(name="add2x", x=wf.lzin.x).split("x"))
+    wf.add(add2(name="add2y", x=wf.lzin.x).split("x"))
+    wf.add(multiply(name="mult", x=wf.add2x.lzout.out, y=wf.add2y.lzout.out))
```
satra, line 1489: this uses the default outer product splitter for the third task, but what happens in a two node workflow where the output of the first task with a splitter goes to two inputs of the second task. does the same rule apply?

also is this now a nested state that you would need two combines to get out of?

pydra/engine/tests/test_workflow.py
```diff
@@ -1403,52 +1454,198 @@ def test_wf_3nd_ndst_6(plugin, tmpdir):
...
+    the third one uses outer splitter (so the first splitter is "repeated")
+    """
+    wf = Workflow(name="wf_ndst_9", input_spec=["x"])
+    wf.add(add2(name="add2x", x=wf.lzin.x).split("x"))
+    wf.add(add2(name="add2y", x=wf.add2x.lzout.out))
     wf.add(
-        identity_2flds(
-            name="iden2flds_2", x1=wf.identity.lzout.out, x2=wf.iden2flds_1.lzout.out2
+        multiply(name="mult", x=wf.add2x.lzout.out, y=wf.add2y.lzout.out).split(
+            ["_add2x", "_add2y"]
```
satra, line 1518: is this explicit split required and why? how would the user/workflow creator know to do this?

why is this different from: `test_wf_3nd_ndst_8`

satra: i see the next test uses default

satra (commented):
overall it looks good. mostly some comments on behavior in a two node case, and how the user is expected to know what happens?

pydra/engine/tests/test_workflow.py
djarecka: yes, was checking if gives the same when default

djarecka:
I reverted some changes. I'm not sure if I like them anymore, and it would need more work anyway. Notes from that changes are here: https://github.com/djarecka/pydra/tree/rf/state_repeated_notes

PR #501, opened 2021-10-04
added nest_asyncio to install requires

tclose:
## Acknowledgment

## Types of changes

## Summary

Adds nest-asyncio dependency to setup.cfg

## Checklist

N/A


djarecka:
since this is required only for working with jupyter notebook, I would not add it to the basic requirements. However, we could perhaps add additional mode of installation "jupyter". Or checking if pydra is run inside the jupyter notebook and raise an error if `nest_asyncio` is not detected.
What do you think?

tclose:
added "jupyter" install option to pyproject.toml as suggested

PR #516, opened 2022-03-22
NF: Add Environment class, with initial Native/Docker implementations

effigies:


djarecka:
for now `test_docker.py` and `test_singularity.py` might not work, examples for docker env are in `test_environments.py` 
(still more needed)

ghisvail:
I was able to test this PR successfully on [pydra-deface](https://github.com/ghisvail/pydra-deface) using FSL tasks and a [container image](https://github.com/ReproNim/neurodocker/pull/527) built from the latest version of FSL.

A few paths of improvement:

- Although `docker` is the most "popular", `podman` is also a suitable alternative. It would be nice to offer the possibility to use any of those somehow. I was able to test `podman` by directly editing the source code in my test venv.

- On the same vain, the Docker registry is certainly the most popular because it is the default, but it would also be worth being able to specify an alternative registry such as Quay or GHCR for instance.

More info can be found in the [containers transport](https://github.com/containers/image/blob/main/docs/containers-transports.5.md) documentation.

effigies:
I agree, we should definitely try to keep this general. I don't have enough experience with non-Docker/non-Singularity runtimes to be opinionated here, so happy to follow your lead.

ghisvail:
I did a bit of research:

- Regarding supporting `podman` in addition to `docker`, since the former provides a [compatible CLI](https://developers.redhat.com/blog/2020/11/19/transitioning-from-docker-to-podman#install_podman) with the latter, Linux distributions either advise [using an alias](https://nixos.wiki/wiki/Podman) or provide a [compatibility package](https://packages.ubuntu.com/lunar/podman-docker). So it sounds fine to assume the runtime will have a `docker` command available at first, before thinking of a means to override it if need be.

- Regarding support for other registries, I think the most comfortable is to leave it as part of the image name, i.e. `ghcr.io/$username/$image`, instead of introducing a `registry` parameter. Let `docker` or `podman` deal with the default registry
based on their respective settings.

To conclude, I think the current implementation is good enough as it is :+1:

effigies:
@djarecka This is passing locally, but there are a lot of skipped tests. We'll see how far this gets, but this is it from me for today.

djarecka:
@effigies - ok, thank you. I will review the tests that are failing

effigies (commented):
Looked at this with some fresh eyes and left a few notes. We should be careful to name the functions and attributes to make future contributors (such as ourselves) less confused.

djarecka:
i'm trying to understand why the slurm test fails now, since you didn't any new test here yet. any idea?

effigies:
Nope. I've never understood anything about the slurm tests. I'll rebase on master and push, if that might help.

djarecka:
@effigies - sorry, that was meant to be written in a different PR...too many tabs...;-)

effigies:
Yeah, slurm is still failing.

djarecka:
with the new image it was working last week, so not sure what happened, but that's the discussion for another place...

pydra/engine/environments.py
```diff
@@ -0,0 +1,84 @@
...
+        self, image, tag="latest", binds=None, output_cpath="/output_pydra", xargs=None
+    ):
+        self.image = image
+        self.tag = tag
+        self.xargs = xargs
+        self.output_cpath = output_cpath
+
+    @staticmethod
+    def bind(loc, mode="ro", root="/mnt/pydra"):  # TODO
+        # XXX Failure mode: {loc} overwrites a critical directory in image
```
djarecka, line 45: @effigies - not sure if I understand this comment. Do you remember what you had in mind? is it still relevant?

pydra/engine/environments.py
```diff
@@ -0,0 +1,84 @@
+from .helpers import execute
+
+from pathlib import Path
+
+
+class Environment:
```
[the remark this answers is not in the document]
ghisvail: Depends mostly whether there is a lot of code re-use coming from the base class, and whether you expect a lot of `isinstance` calls downstream.

So far, the `setup` and `teardown` naming convention screams `ContextManager` to me, but I am probably missing something.

ghisvail (commented):
I think the value proposition from the proposed `Environment` abstraction is clear, I am just a bit confused about the execution.

It sounds to me like they could be simplified to a bunch of context managers, maybe:

```python
# somewhere in the submitter's code
match environment:  # assuming [type, *args]
    case ["docker", image, tag]:
        with setup_docker(task, image, tag) as env:
            ...
    case _:  # fallback to native
        with setup_native(task) as env:
            ...
```

Here I am using the new pattern matching syntax, but a stack of `isintance` calls would work the same.

The other worry I have got is how composable the `Environment` family of values is with the `Submitter` one. All submitters (cf, Dask, SLURM, ...) should work with native, but would that be the case for containers (Docker, Singularity) or others we had in mind like Conda?

I am totally onboard with the idea to dissociate the execution environment from the task abstraction. But I am still feeling the boundary between `Environment` and `Submitter` is not as clear cut than anticipated? I don't have further ideas to improve the proposal at this point, so I won't be opposed to a merge 👍

pydra/engine/environments.py
effigies: I think this comment is out of date.

pydra/engine/environments.py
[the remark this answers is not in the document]
djarecka: I actually don't think that we need setup and teardown at the end...

pydra/engine/environments.py
```diff
@@ -0,0 +1,123 @@
...
+                raise RuntimeError(output["stderr"])
+            else:
+                raise RuntimeError(output["stdout"])
+        # Any outputs that have been created with a re-rooted path need
+        # to be de-rooted
+        # task.finalize_outputs("/mnt/pydra") TODO: probably don't need it
+        return output
+
+
+class Singularity(Docker):
```
effigies, line 85: This is simple enough that I think we could just copy/paste the single additional `bind()` method (or just make it a function, since it's a staticmethod) to keep the class hierarchy flat. If there's a lot of duplication, we could make an intermediate `Container(Environment)` subclass and then make `Singularity(Container)` and `Docker(Container)`.

pydra/engine/specs.py
```diff
@@ -693,20 +693,6 @@ class ContainerSpec(ShellSpec):
     )
```
effigies, line 693: I think we want to get rid of `ContainerSpec`.

pydra/engine/environments.py
```diff
@@ -0,0 +1,160 @@
...
+        ]
+        docker_args.extend(self.xargs)
+        docker_args.extend(
+            " ".join(
+                [f"-v {key}:{val[0]}:{val[1]}" for (key, val) in mounts.items()]
+            ).split()
+        )
+        docker_args.extend(["-w", f"{self.root}{task.output_dir}"])
+        keys = ["return_code", "stdout", "stderr"]
+        # print("\n Docker args", docker_args)
```
effigies, line 111:
```suggestion
```

pydra/engine/tests/test_specs.py
```diff
@@ -53,33 +51,17 @@ def test_shellspec():
...
+#
+#
+# def test_container():
+#     with pytest.raises(TypeError):
+#         spec = ContainerSpec()
+#     spec = ContainerSpec(
+#         executable="ls", image="busybox", container="docker"
+#     )  # (execute, args, image, cont)
+#     assert all([hasattr(spec, attr) for attr in container_attrs])
+#     assert hasattr(spec, "executable")
```
effigies, line 64:
```suggestion
```

effigies (commented):
This is pretty much ready to go. Small comments.

djarecka:
@ghisvail @tclose - want to try this before merging?

tclose:
> @ghisvail @tclose - want to try this before merging?

Don't have the time atm, so happy to go with this and find problems later

PR #520, opened 2022-04-01
simplifying ShellCommandTask.cmdline

djarecka:
simplifying all methods related to `cmdline` for `ShellCommandTask` 
- creating command line only for stateless versions  of the task (`cmdline` is needed when the task is executed and after the state has been removed by the submitter) 
- temporarily removing some asserts from tests since we can't ask for `task.cmdline` when task has a state
- we could bring back the option to create cmdline for tasks with state by creating a copy of a task (in the same way as it is done in the submitter), but likely in a different PR

## Acknowledgment

## Types of changes

## Checklist

- [ ] All tests passing

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
     -   id: check-yaml
     -   id: check-added-large-files
 -   repo: https://github.com/psf/black
-    rev: 22.1.0
+    rev: 22.3.0
     hooks:
     -   id: black
 -   repo: https://github.com/codespell-project/codespell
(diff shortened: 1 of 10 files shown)


djarecka:
@satra - I recalled that you were saying something about `click`, was this related to the issue:https://results.pre-commit.ci/run/github/152126811/1649028674.Fu63lAIhSCKWzxHD9TpUJA

satra:
no that was just that click is not a dependency of pydra. in this case it's likely a function of black version in pre-commit.

djarecka:
I will merge this now. Some of the test will have to be cleaned later.

PR #524, opened 2022-04-01
RF: Submitter logic.

mgxd:
A quick pass to try and clean up some of the logic in the `Submitter` class. This can definitely be reduced further, but hopefully this starts with the untangling.

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -1055,7 +1055,9 @@ class Workflow(TaskBase):
                 "Workflow output cannot be None, use set_output to define output(s)"
             )
         # creating connections that were defined after adding tasks to the wf
-        self._connect_and_propagate_to_tasks()
+        self._connect_and_propagate_to_tasks(
+            propagate_rerun=self.task_rerun and self.propagate_rerun
+        )
 
         checksum = self.checksum
         output_dir = self.output_dir
@@ -1234,7 +1236,7 @@ class Workflow(TaskBase):
         self,
         *,
         propagate_rerun=False,
-        override_task_caches=False
+        override_task_caches=False,
     ):
         """
         Visit each node in the graph and create the connections.
@@ -1244,11 +1246,11 @@ class Workflow(TaskBase):
             self.create_connections(task)
             # if workflow has task_rerun=True and propagate_rerun=True,
             # it should be passed to the tasks
-            if self.task_rerun and self.propagate_rerun:
-                task.task_rerun = self.task_rerun
+            if propagate_rerun:
+                task.task_rerun = True
                 # if the task is a wf, than the propagate_rerun should be also set
                 if is_workflow(task):
-                    task.propagate_rerun = self.propagate_rerun
+                    task.propagate_rerun = True
 
             # ported from Submitter.__call__
             # TODO: no prepare state ?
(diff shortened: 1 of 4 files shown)


pydra/engine/submitter.py
```diff
@@ -1,14 +1,7 @@
...
-from .workers import (
-    SerialWorker,
-    ConcurrentFuturesWorker,
-    SlurmWorker,
-    DaskWorker,
-    SGEWorker,
-)
+from .workers import WORKERS
 from .core import is_workflow
 from .helpers import get_open_loop, load_and_run_async
```
mgxd, line 6: given that these functions are only used here, it might help with readability just having them here.

pydra/engine/submitter.py
```diff
@@ -110,41 +98,37 @@ async def submit(self, runnable, wait=False, rerun=False):
             Coroutines for :class:`~pydra.engine.core.TaskBase` execution.
 
         """
+        if runnable.plugin and runnable.plugin != self.plugin:
+            raise NotImplementedError()
```
mgxd, line 102: This was removed since it wasn't being tested.

original:
```python
# dj: this is not tested!!! TODO
await self.worker.run_el(workflow, rerun=rerun)
```

pydra/engine/submitter.py
```diff
@@ -110,41 +98,37 @@ async def submit(self, runnable, wait=False, rerun=False):
...
+
+        task_pkl = await prepare_runnable_with_state(runnable)
+
+        for sidx in range(len(runnable.state.states_val)):
             if is_workflow(runnable):
-                await self._run_workflow(runnable, rerun=rerun)
+                # job has no state anymore
+                futures.add(
+                    # This unpickles and runs workflow - why are we pickling?
+                    asyncio.create_task(load_and_run_async(task_pkl, sidx, self, rerun))
```
mgxd, line 115: I think this bit is still excessively confusing, and could really benefit from a rework / simplification. `Worker.run_el` should have a streamlined behavior - currently it handles two pathways (state vs stateless)

mgxd (commented):
I went through and left some thought / talking points - this is ready for a review.

pydra/engine/core.py
```diff
@@ -1227,21 +1232,31 @@ def create_dotfile(self, type="simple", export=None, name=None):
...
         Additionally checks if all tasks should be rerun.
         """
         for task in self.graph.nodes:
+            self.create_connections(task)
             # if workflow has task_rerun=True and propagate_rerun=True,
             # it should be passed to the tasks
-            if self.task_rerun and self.propagate_rerun:
-                task.task_rerun = self.task_rerun
+            if propagate_rerun:
+                task.task_rerun = True
```
djarecka, line 1250: I'm wondering if this should be always `True`, even if `self.task_rerun` is `False`?

pydra/engine/submitter.py
djarecka: agree

djarecka:
@mgxd - thank you a lot! Any idea one of the slurm test is failing? (seems to be consistent) If not, I can try to debug tomorrow

pydra/engine/core.py
mgxd: it should never be False, since we call this as:

```python
        self._connect_and_propagate_to_tasks(
            propagate_rerun=self.task_rerun and self.propagate_rerun
        )
```

mgxd:
I haven't looked into the SLURM fail too much, but it seems like it is submitting each FunctionTask within the workflow through SLURM directly, instead CF (which I think it was doing before?).

djarecka:
ok, I'm merging this for now. We can have another round of refactoring in another PR at some point. Thanks @mgxd

Issue #442, opened 2021-03-25
create_dotfile fail if workflow has a splitter


djarecka:
reported by @jw-96. Thanks

tclose:
Has this been fixed?

JeffWigger:
According to #538 it seems to be resolved. I (formerly jw-96) do no longer have the original example that led to this bug report.

tclose:
ok, thanks :)

PR #538, opened 2022-04-17
use cache_dir instead of output_dir

montesmariana:
## Types of changes
- Bug fix (non-breaking change which fixes an issue), fixing Issue #442 

## Summary
Add the option of specifying the output directory for the dotfile (the graph of a workflow) and when it is not specified, use `cache_dir` instead of `output_dir` (because when there is splitting the `output_dir` attribute is a list of paths instead of a path).

## Checklist
I have not added tests or updated documentation (there is no much original documentation on this method that I could find)

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -1189,7 +1189,7 @@ class Workflow(TaskBase):
                     )
         return attr.evolve(output, **output_wf)
 
-    def create_dotfile(self, type="simple", export=None, name=None, output_dir = None):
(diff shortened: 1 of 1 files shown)


djarecka:
@montesmariana - I'm sorry, I completely missed this PR... Thank you so much for fixing it!

I will merge it, but could you please add your name to the [zenodo file](https://github.com/nipype/pydra/blob/master/.zenodo.json). We haven't kept any particular order, but please just keep Satra Ghosh as the last author.

PR #545, opened 2022-05-31
Split command line using quote respecting shlex.split instead of ' '.split

tclose:
## Types of changes

## Summary
Fixes an issue when there are quotes in argument strings to enclose spaces and other special characters.

## Checklist

### DIFF
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -484,10 +484,8 @@ class ShellCommandTask(TaskBase):
                         cmd_el_str = f"{argstr} {value}"
                     else:
                         cmd_el_str = ""
-            # removing double spacing
-            cmd_el_str = cmd_el_str.strip()
             if cmd_el_str:
-                cmd_add += cmd_el_str
+                cmd_add += shlex.split(cmd_el_str)
         return pos, cmd_add
 
     @property
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -39,6 +39,7 @@ Implement processing nodes.
 
 """
 import platform
+import re
 import attr
 import cloudpickle as cp
 import inspect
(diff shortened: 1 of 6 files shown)


pydra/engine/task.py
```diff
@@ -519,10 +518,19 @@ def _run_task(self):
...
                 if self.output_["stderr"]:
-                    raise RuntimeError(self.output_["stderr"])
-                else:
-                    raise RuntimeError(self.output_["stdout"])
+                    msg += "\n\nstderr:\n" + self.output_["stderr"]
+                if self.output_["stdout"]:
+                    msg += "\n\nstderr:\n" + self.output_["stdout"]
+                raise RuntimeError(msg)
+
+    def _join_cmd_args(self, args):
```
tclose, line 528: Not 100% sure about this, feels a bit hacky just skipping the executable but it should work. Any suggestions on a more elegant solution?

tclose:
Hi @djarecka, could you have a look at this when you get the chance? Basically uses shlex.split instead of str.split(' ') when creating command args. This allows you to put quotes around cmd args and handle special cases like spaces in file names.

 The tests that are failing seem to be some issue with the GH action runner rather than my code if you look at them.

tclose:
Ah, so it seems I have some Windows errors, which is annoying as I don't have a Windows test box atm. Seems as though the Windows path backslashes are getting eaten by the Python string. Will try to work out why that might be.

NB: the code-cov patch is down because I have added a switch to the conftest to allow you to enable the raising of exceptions while debugging unittests

djarecka:
Hmm, not sure why windows raises the error just after reading it (also don't have windows)...
But I'm also not sure if I understand the changes in pytest - seems like no test from `test_shelltask.py` was run...

effigies:
I think we're getting bitten by https://github.com/python/cpython/issues/44990. Might be worth looking for a windows-compatible shlex.

effigies:
Actually, what about using: 

```Python
import sys
import shlex

def shell_split(val):
    return shlex.split(var, posix="win" not in sys.platform)
```

Via https://github.com/jrnl-org/jrnl/issues/348

tclose:
Nice find @effigies! Although I had to change their clause because it was matching "dar**win**" :) 

All checks are passing now except the code-cov due to the debugging clause I inserted into conftest. Happy to take this out if you are not a fan.

tclose:
Ok, I believe this is good to go from my end. Just 1.6% down on the coverage patch

djarecka:
Thanks @tclose !

I think we can keep the the debugging clause.

However I do worry that you are having issue with running `test_shelltask.py` on your laptop. Could you create an issue and show the error?

tclose:
Thanks @djarecka, actually after I reinstalled pydra[dev] and reconfigured my VSCode pytest all the tests showed up, so it must have just been something that needed refreshing.

PR #553, opened 2022-07-11
function in output_file_template

djarecka:
## Types of changes

## Summary
allowing output_file_template to be a function that returns a different templates depending on the values of the input fields

## Checklist


yibeichan:
@djarecka thank you! I updated with your code and passed the test locally, but the output from the test seems a bit weird. I'll show you tomorrow morning

PR #561, opened 2022-08-06
changing template formatting when value (or default) is False

djarecka:
## Types of changes

## Summary

when output use template to create the output path and value is `False` the field should not be used in the cmdline, it was inconsistent

## Checklist

### DIFF
--- a/pydra/engine/tests/test_shelltask_inputspec.py
+++ b/pydra/engine/tests/test_shelltask_inputspec.py
@@ -1152,6 +1152,8 @@ def test_shell_cmd_inputs_template_6():
         executable="executable", input_spec=my_input_spec, inpA="inpA"
     )
     assert shelly.cmdline == f"executable inpA -o {str(shelly.output_dir / 'inpA_out')}"
+    # checking if the command is the same
+    assert shelly.cmdline == f"executable inpA -o {str(shelly.output_dir / 'inpA_out')}"
 
     # a string is provided for outA, so this should be used as the outA value
     shelly = ShellCommandTask(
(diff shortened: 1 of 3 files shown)


djarecka:
@yibeichan - this is the fix for the template, I was trying this on the example of `FNIRT`, but you can try on your example if you want and let me know if this makes sense

yibeichan:
@djarecka thank you! so I create a local branch for `pydra`, used your code. then run pytest in `pydra-fsl` on `cluster.py`, but my `task.cmdline` still changed after `task.run()`. Here is my `cluster.py` https://github.com/yibeichan/pydra-fsl/blob/wip-yc-convert/pydra/tasks/fsl/model/cluster.py; here is the test https://github.com/yibeichan/pydra-fsl/blob/wip-yc-convert/pydra/tasks/fsl/model/tests/test_run_cluster.py

pydra/engine/tests/test_shelltask_inputspec.py
```diff
@@ -1170,6 +1172,8 @@ def test_shell_cmd_inputs_template_6():
         executable="executable", input_spec=my_input_spec, inpA="inpA", outA=False
     )
     assert shelly.cmdline == "executable inpA"
+    # checking of the command is the same
+    assert shelly.cmdline == "executable inpA"
```
yibeichan, line 1176: @djarecka isn't line 1176 the same as the line 1174? shouldn't we add something like `shelly.run()` between 1174 and 1176?

djarecka: actually I realized that the `run` wasn't an issue when we were checking this last week, simply repeating `cmdline` was giving a new result, it was a bug... That's why I added this to the test...

djarecka:
thanks, I will check your tests

tclose:
I believe this is how the current main works now

PR #565, opened 2022-08-16
Add support for arbitrary mappings in a FunctionTask

ghisvail:
## Types of changes

## Summary

This PR adds support for returning total or partial mappings of the output spec by the wrapped function, as an alternative to going through a conversion from the returned tuple to a dictionary via `output_names`. Unlike the named tuple case where the shape of the data is known a priori (I have got exactly 2 output variables which I want named to `foo` and `bar`), this PR adds support for functions which may return more, less or exactly these data as an arbitrary mapping.

This could be particularly useful for dynamic task construction use cases, like BIDS readers, parsers and writers.

## Checklist

### DIFF
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -198,6 +198,8 @@ class FunctionTask(TaskBase):
             self.output_ = {output_names[0]: output}
         elif isinstance(output, tuple) and len(output_names) == len(output):
             self.output_ = dict(zip(output_names, output))
+        elif isinstance(output, dict):
+            self.output_ = {key: output.get(key, None) for key in output_names}
         else:
             raise RuntimeError(
                 f"expected {len(self.output_spec.fields)} elements, "
--- a/pydra/engine/tests/test_task.py
+++ b/pydra/engine/tests/test_task.py
@@ -114,6 +114,27 @@ def test_annotated_func(use_validator):
     ]
 
 
+def test_annotated_func_dictreturn(use_validator):
+    """Test mapping from returned dictionary to output spec."""
+
+    @mark.task
+    @mark.annotate({"return": {"sum": int, "mul": int}})
+    def testfunc(a: int, b: int):
+        return dict(sum=a + b, diff=a - b)
+
+    task = testfunc(a=2, b=3)
+    result = task()
+
+    # Part of the annotation and returned, should be exposed to output.
+    assert result.output.sum == 5
+
+    # Part of the annotation but not returned, should be coalesced to None.
+    assert result.output.mul is None
+
+    # Not part of the annotation, should be discarded.
+    assert not hasattr(result.output, "diff")
+
+
 def test_annotated_func_multreturn(use_validator):
     """the function has two elements in the return statement"""
 


ghisvail:
For a concrete example, it would simplify this piece of terrible code very much:

https://github.com/aramis-lab/pydra-bids/blob/main/pydra/tasks/bids/utils/read_bids.py

ghisvail:
A safer alternative could also be to only return if the mapping returned by the inner function contains all keys declared in `output_names`. Right now, I went for the coalesce to `None` option, since this is what we are going for in case the function did not return anything. I am happy with either.

satra:
a side note: for dynamic tasks as being targeted here, the cache checksum/location will only depend on inputs (e.g. the query string/template), not the underlying database. hence either the source directory would need to be provided and cached  as a separate input(a potentially expensive operation), or another default input should be associated that uses a random number generator as a default.  the current forcing option of `self.rerun_task` may not suffice in all cases, especially if the same task gets called multiple times.

ghisvail:
Good point. How does caching work today with nipype for things like `BIDSDataGrabber` and `DataSink`? Would it work differently with `pydra` fundamentally?

satra:
> How does caching work today with nipype for things like BIDSDataGrabber and DataSink?

caching is local to a workflow and nipype1 does not have the concept of reusing the same task in multiple places or even across workflows. this is only available in pydra. in nipype1, simply setting the rerun flag would force that interface to rerun, but because there is no shared rerun capabilities, this particular issue does not come into play. so dynamic interfaces are simply set to rerun by default in nipype1.

djarecka:
@ghisvail - can't access the link with the example anymore.

perhaps you can also add some test, e.g. `p`ydra/mark/tests/test_functions.py` and/or `pydra/engine/tests/test_task.py`

ghisvail:
Yes, I am planning to add some unit tests covering this particular usage.

I'll ping you once done :+1

ghisvail:
Yes, I am planning to add some unit tests covering this particular usage.

I'll ping you once done :+1

ghisvail:
I added a test case which covers the present, absent and undefined mappings from returned `dict` to annotation.

A natural extension to this PR would be to handle functions annotated with a [TypedDict](https://peps.python.org/pep-0589/) return value.

ghisvail:
> @ghisvail - can't access the link with the example anymore.

I did some refactoring of the code in the meantime. This is what it looks like today:

https://github.com/aramis-lab/pydra-bids/blob/main/pydra/tasks/bids/utils.py

Granted it's likely a terrible way of making dynamic tasks for reasons I have not thought of, but I sense there could be a nice pattern emerging from this.

ghisvail:
@djarecka do you think we could get this one in for the release too?

djarecka:
if you think that you could use it, then sure

effigies:
Going forward, I think it would be good to deprecate `{'name': type}` for `TypedDict`, which didn't exist when we initially came up with the return annotations.

ghisvail:
> Going forward, I think it would be good to deprecate `{'name': type}` for `TypedDict`, which didn't exist when we initially came up with the return annotations.

We might not need to deprecate it, since `TypedDict` can be constructed [dynamically](https://peps.python.org/pep-0589/#alternative-syntax). So we could normalize tasks internals to use `TypedDict` but keep `annotate` as a user-friendly wrapper to help generate the `TypedDict` definition from the mapping passed in `annotate`.

What do you think of this plan?

effigies:
Sounds good to me.

ghisvail:
Ok, then I'll find some time to start hacking on it 👍

djarecka:
@ghisvail - you want to still work in this PR, or you want me to merge this one and open a new one?

ghisvail:
@djarecka please go ahead with merging. I'll open a follow-up PR for `TypedDict`.

PR #567, opened 2022-08-17
Make Makefile compatible with macOS

ghisvail:
It looks like macOS distribution of `rm` does not understand long versions of
flag names, such as --force or --recursive. This PR switches to their
respective short versions (-f and -r) to increase compatibility.


djarecka:
thank you!

PR #575, opened 2022-09-02
[wip] adding a part for pandas.df to hash_value

djarecka:
## Types of changes

## Summary

adding a section to `hash_value` that is for `pandas.DF`: changing df to dict before calculating the hash value.

@yibeichan - could you please see if this small change fixes some of your issues? it did fix the issue that I had with running my workflow within the jupyter notebook...

## Checklist

### DIFF
--- a/pydra/engine/helpers.py
+++ b/pydra/engine/helpers.py
@@ -704,6 +704,11 @@ def hash_value(value, tp=None, metadata=None, precalculated=None):
                 hash_value(el, tp, metadata, precalculated)
                 for el in ensure_list(value.tolist())
             ]
+        elif type(value).__module__ == "pandas.core.frame":
+            return [
+                hash_value(el, tp, metadata, precalculated)
+                for el in ensure_list(value.to_dict())
+            ]
         else:
             return value
 


yibeichan:
@djarecka thank you! I'll try it today and let you now

yibeichan:
@djarecka this change is specific for a list of `pandas.Dataframe` as inputs, correct? 
my notebook works fine with or without this change because I'm using python 3.7? If this can solve your problem, then it probably can solve potential problems in higher python as well.
However, I exported my notebook as `.py` and got errors (not related to your change here)
>File "/Users/yibeichen/miniconda3/envs/pydra/lib/python3.7/asyncio/futures.py", line 181, in result
    raise self._exception
concurrent.futures.process.BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.

this error happens after `secondlevel_estimation` when starting statistical testing.
Only `.py` script has this error, notebook works fine.

tclose:
I think this should be covered by the sequence hashing functionality shouldn't it @effigies ?

effigies:
Seems like it's worth just testing.

PR #581, opened 2022-09-20
More descriptive debugging graph errors and bugfix in values validator

tclose:
## Types of changes
- Bug fix (non-breaking change which fixes an issue): 
- New feature (non-breaking change which adds functionality):

## Summary
* `_allowed_values_validator` now accepts LazyField objects
*  more detailed error messages when there is a problem with the workflow graph and the dependencies of some nodes cannot be satisfied

## Checklist

### DIFF
--- a/pydra/engine/helpers.py
+++ b/pydra/engine/helpers.py
@@ -482,7 +482,7 @@ def _check_special_type(tp, name):
 def _allowed_values_validator(instance, attribute, value):
     """checking if the values is in allowed_values"""
     allowed = attribute.metadata["allowed_values"]
-    if value is attr.NOTHING:
+    if value is attr.NOTHING or isinstance(value, LazyField):
         pass
     elif value not in allowed:
         raise ValueError(
--- a/pydra/engine/helpers.py
+++ b/pydra/engine/helpers.py
@@ -112,7 +112,7 @@ def load_result(checksum, cache_locations):
     if not cache_locations:
         return None
     # TODO: if there are issues with loading, we might need to
-    # TODO: sleep and repeat loads (after checkin that there are no lock files!)
+    # TODO: sleep and repeat loads (after checking that there are no lock files!)
     for location in cache_locations:
         if (location / checksum).exists():
             result_file = location / checksum / "_result.pklz"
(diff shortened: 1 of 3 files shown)


tclose:
@djarecka do you know why this would be failing the SLURM check? I don't think I changed anything that would impact on it

```
plugin = 'slurm'
tmpdir = local('/tmp/pytest-of-root/pytest-0/popen-gw0/test_wf_upstream_error3_slurm_0')

    def test_wf_upstream_error3(plugin, tmpdir):
        """task2 dependent on task1, task1 errors, task-level split on task 1
        goal - workflow finish running, one output errors but the other doesn't
        """
        wf = Workflow(name="wf", input_spec=["x"], cache_dir=tmpdir)
        wf.add(fun_addvar_default(name="addvar1", a=wf.lzin.x))
        wf.inputs.x = [1, "hi"]  # TypeError for adding str and int
        wf.addvar1.split("a")  # task-level split
        wf.plugin = plugin
        wf.add(fun_addvar_default(name="addvar2", a=wf.addvar1.lzout.out))
        wf.set_output([("out", wf.addvar2.lzout.out)])
    
        with pytest.raises(Exception) as excinfo:
            with Submitter(plugin=plugin) as sub:
                sub(wf)
>       assert "addvar1" in str(excinfo.value)
E       AssertionError: assert 'addvar1' in 'Event loop is closed'
E        +  where 'Event loop is closed' = str(RuntimeError('Event loop is closed'))
E        +    where RuntimeError('Event loop is closed') = <ExceptionInfo RuntimeError('Event loop is closed') tblen=11>.value
```

djarecka:
@tclose - this test is flaky, perhaps I should do more reruns, before reporting fails.

I wonder if you had some test cases that you were testing this on and could be added to the repo?

tclose:
> @tclose - this test is flaky, perhaps I should do more reruns, before reporting fails.
> 
> I wonder if you had some test cases that you were testing this on and could be added to the repo?

I fixed the "real" case that was failing but I have been meaning to create a test with an incorrectly designed task, which will trigger this case

djarecka:
ok, I will merge this now, so it is in the newest release. I'm sure this will be an update to the previous error report. I've created an issue to think about the tests: #590

PR #584, opened 2022-09-26
[ENH] Add `_graph_checksums` to `input_spec` if missing

NicolasGensollen:
## Types of changes

## Summary

Following #573, this PR proposes to automatically add `_graph_checksums` to the `input_spec` provided to the workflow when these specifications are provided directly as a `SpecInfo` instance.
This is already done automatically when `input_spec` is provided as a list of strings.
This PR also proposes to put the input sanitization code in a separate helper function.

## Checklist

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -798,8 +798,8 @@ class TaskBase:
 
 
 def _sanitize_input_spec(
-        input_spec: ty.Union[BaseSpec, ty.List[str]],
-        wf_name: str,
+    input_spec: ty.Union[BaseSpec, ty.List[str]],
+    wf_name: str,
 ) -> BaseSpec:
     """Makes sure the provided input specifications are valid.
 
@@ -831,9 +831,7 @@ def _sanitize_input_spec(
             return input_spec
         elif isinstance(input_spec, SpecInfo):
             if not any([x == BaseSpec for x in input_spec.bases]):
-                raise ValueError(
-                    "Provided SpecInfo must have BaseSpec as it's base."
-                )
+                raise ValueError("Provided SpecInfo must have BaseSpec as it's base.")
             if "_graph_checksums" not in {f[0] for f in input_spec.fields}:
                 input_spec.fields.insert(0, graph_checksum_input)
             return input_spec
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -798,9 +798,9 @@ class TaskBase:
 
 
 def _sanitize_input_spec(
-    input_spec: ty.Union[BaseSpec, ty.List[str]],
+    input_spec: ty.Union[SpecInfo, ty.List[str]],
     wf_name: str,
-) -> BaseSpec:
+) -> SpecInfo:
     """Makes sure the provided input specifications are valid.
 
     If the input specification is a list of strings, this will
@@ -808,7 +808,7 @@ def _sanitize_input_spec(
 
     Parameters
     ----------
-    input_spec : BaseSpec or List[str]
+    input_spec : SpecInfo or List[str]
         Input specification to be sanitized.
 
     wf_name : str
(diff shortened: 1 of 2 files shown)


pydra/engine/core.py
```diff
@@ -797,6 +797,66 @@ def _reset(self):
...
+        Sanitized input specifications.
+
+    Raises
+    ------
+    ValueError
+        If provided `input_spec` is None.
+    """
+    graph_checksum_input = ("_graph_checksums", ty.Any)
+    if input_spec:
+        if isinstance(input_spec, BaseSpec):
```
NicolasGensollen, line 830: It looks like this line wasn't tested. I tried adding a new test for it, but it seems like giving a `BaseSpec` or a derived class instance to the `Workflow` constructor gives an error:


```python
>>> from pydra.engine.specs import ShellSpec
>>> spec = ShellSpec(executable="foo")
>>> from pydra.engine import Workflow
>>> wf = Workflow(name="bar", input_spec=spec)
  File "<stdin>", line 1, in <module>
  File "/Users/nicolas.gensollen/GitRepos/pydra/pydra/engine/core.py", line 917, in __init__
    super().__init__(
  File "/Users/nicolas.gensollen/GitRepos/pydra/pydra/engine/core.py", line 140, in __init__
    klass = make_klass(self.input_spec)
  File "/Users/nicolas.gensollen/GitRepos/pydra/pydra/engine/helpers.py", line 258, in make_klass
    fields = spec.fields
AttributeError: 'ShellSpec' object has no attribute 'fields'
```

satra: the previous PR change that was merged specifically insisted on `BaseSpec`, since a Workflow should never require anything else. so a `ShellSpec` instance in theory should not be passed to a Workflow. do you get the same issue if you use a `BaseSpec`? (i don't think a release was made since that change, so you would have to test with the github master branch)

NicolasGensollen: Yes, using a `BaseSpec` or any derived class instance gives the same error:

```python
>>> from pydra.engine.specs import BaseSpec
>>> from pydra.engine import Workflow
>>> wf = Workflow(name="bar", input_spec=BaseSpec())
  File "<stdin>", line 1, in <module>
  File "/Users/nicolas.gensollen/GitRepos/pydra/pydra/engine/core.py", line 883, in __init__
    super().__init__(
  File "/Users/nicolas.gensollen/GitRepos/pydra/pydra/engine/core.py", line 140, in __init__
    klass = make_klass(self.input_spec)
  File "/Users/nicolas.gensollen/GitRepos/pydra/pydra/engine/helpers.py", line 258, in make_klass
    fields = spec.fields
AttributeError: 'BaseSpec' object has no attribute 'fields'
```

I am on the latest commit on master.

satra: sorry, this indeed needs some update. it cannot be a `BaseSpec` by itself, it needs to be something of this nature: https://github.com/nipype/pydra/blob/master/pydra/engine/core.py#L855 

In fact the `SpecInfo` version should also check for graph_checksums as a field. i think this is primarily why we avoided allowing this to simply be a named list, but if you are up for it to fix this up in your PR that would be great.

technically any input needs to be a SpecInfo, hence BaseSpec should probably be removed from the typing list as well.

i know @tclose looked at this recently, so please chime in.

NicolasGensollen: Thanks for your fast reply @satra. If I understand correctly, the user shouldn't be able to pass a `BaseSpec` instance to the `Workflow` constructor, but either a `SpecInfo` instance, or a list of strings which is then converted to a proper `SpecInfo` instance. 

So, unless I missed something, this is what needs to be done:

- Remove `BaseSpec` from the type hint and update the docstring accordingly
- Raise a `ValueError` when a `BaseSpec` instance is passed to the constructor (basically remove lines 830 and 831 in this PR)
- Make sure the input specifications have a `_graph_checksums` attribute (implemented in this PR for `SpecInfo` and already implemented for list of strings).

satra: thank you @NicolasGensollen  for tracking this down.  your steps are right. i don't think it makes sense to have `BaseSpec` as an input, but it's been a long time since we wrote that so the thinking process is no longer in memory.

PR #585, opened 2022-09-27
Provide paths to ShellTask output files

tclose:
## Types of changes

## Summary
Fixes passing of explicit output paths to shell "output" input fields that define "output_file_template", which otherwise are used in the `cmdline` string but not when collecting the outputs

## Checklist

### DIFF
--- a/pydra/engine/specs.py
+++ b/pydra/engine/specs.py
@@ -462,26 +462,28 @@ class ShellOutSpec:
             try:
                 additional_out[fld.name] = getattr(inputs, fld.name)
             except AttributeError:
-                if (
-                    fld.default is None or fld.default == attr.NOTHING
-                ) and not fld.metadata:  # TODO: is it right?
-                    raise AttributeError("File has to have default value or metadata")
-                elif fld.default != attr.NOTHING:
-                    additional_out[fld.name] = self._field_defaultvalue(fld, output_dir)
-                elif fld.metadata:
-                    if (
-                        fld.type in [int, float, bool, str, list]
-                        and "callable" not in fld.metadata
-                    ):
-                        raise AttributeError(
-                            f"{fld.type} has to have a callable in metadata"
-                        )
-                    additional_out[fld.name] = self._field_metadata(
-                        fld, inputs, output_dir, outputs
-                    )
+                pass
             else:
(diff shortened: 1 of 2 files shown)


yibeichan:
(continue your example from brainhack) so `output_file_template` will automatically generate a new file which has the same filename as your input file. In your case, you will get `text_zipped.txt` if your input file is `test.txt`. 
Maybe you want to give  `output_callable` a try? Here is an example https://github.com/nipype/pydra-fsl/blob/6957eb67ea8785f8aae1d2aa858a9a1a85a4b51a/pydra/tasks/fsl/model/filmgls.py#L6
the function defined here is a `callable` and will be used here https://github.com/nipype/pydra-fsl/blob/6957eb67ea8785f8aae1d2aa858a9a1a85a4b51a/pydra/tasks/fsl/model/filmgls.py#L276
I might be wrong. Dorota knows better. She is out of town this week and will come back next week.

tclose:
> (continue your example from brainhack) so `output_file_template` will automatically generate a new file which has the same filename as your input file. In your case, you will get `text_zipped.txt` if your input file is `test.txt`. 

Yes, this is the case when the out_file path isn't explicitly provided. But if you explicitly provide a path to out_file (required for mrconvert, which uses the output extension to determine the desired format) then the collected path and the one used in the cmdline will differ and 
attrs.NOTHING will be returned when pydra finds out that the path doesn't exist.

> Maybe you want to give  `output_callable` a try? Here is an example https://github.com/nipype/pydra-fsl/blob/6957eb67ea8785f8aae1d2aa858a9a1a85a4b51a/pydra/tasks/fsl/model/filmgls.py#L6
> 
> the function defined here is a `callable` and will be used here https://github.com/nipype/pydra-fsl/blob/6957eb67ea8785f8aae1d2aa858a9a1a85a4b51a/pydra/tasks/fsl/model/filmgls.py#L276


Thanks, you could get it to work using a callable, but it seems to me that this shouldn't be necessary.


> I might be wrong. Dorota knows better. She is out of town this week and will come back next week.

No worries, thanks for the input.

djarecka:
thanks! It is definitely needed option in case someone wants to provide the name directly!

PR #588, opened 2022-10-15
Initial attempt at adding infile tracking

rcali21:
Adding infile tracking for BIDS-PROV.

### DIFF
--- a/pydra/engine/audit.py
+++ b/pydra/engine/audit.py
@@ -188,7 +188,7 @@ class Audit:
             "label": label,
             "command": command,
             "startedAtTime": now(),
-            "Used": inputs
+            "Used": inputs,
         }
         self.audit_message(start_message, AuditFlag.PROV)
 


rcali21:
WIP-Adding complimentary test.

djarecka:
@rcali21 - thanks for opening this, but remember that a task could use various files (or directories) as an input, not all would be called `in_file`

rcali21:
> @rcali21 - thanks for opening this, but remember that a task could use various files (or directories) as an input, not all would be called `in_file`

Hi Dorota,

Yes, I remember you and I discussing this. This was more of a placeholder for now. Am I remembering correctly that you wanted this to just be a UUID, or was it something else entirely?

rcali21:
> not sure if I understand correctly, `uuid` of the `Input` entity will be used in the `Used` field of the `Activity` (i.e. task), but you have to create a completely new `Input` entity that has several fields.

Yeah, I think I was misunderstanding it then. So "Used" should be implemented after "Entity" is implemented. I can rename the PR to focus on that.

djarecka:
correct, start from `Input` entity, similarly `Output` will be a separate entity

PR #593, opened 2022-11-02
Working on adding entity_message

rcali21:
WIP-Entity message in BIDS Prov


rcali21:
Hi @djarecka,

Couple questions....

When I have the following JN test:
```
shelly = pydra.ShellCommandTask(name='shelly', in_file=file, executable=cmd, audit_flags=AuditFlag.PROV, messengers=PrintMessenger())
print(shelly.inputs.in_file)
```
I am unable to retrieve the in_file from the print statement and instead I'm left with "Inputs has no attribute 'in_file'" even though I define it as a path to a file in a previous line. I will need this in order to fill the 'at_location' field in entity by getting the abs. path of the in_file.

Additionally, you will see that I have added self.audit_message again at the end of audit.py. I'm assuming there is a more elegant way to add another message to self.audit_message.

rcali21:
Wondering if we should strip the filename and just return dir path. [1cc91ef](https://github.com/nipype/pydra/pull/593/commits/1cc91ef4db4fd412d323613135e992b40ac4e951)

djarecka:
> Hi @djarecka,
> 
> Couple questions....
> 
> When I have the following JN test:
> 
> ```
> shelly = pydra.ShellCommandTask(name='shelly', in_file=file, executable=cmd, audit_flags=AuditFlag.PROV, messengers=PrintMessenger())
> print(shelly.inputs.in_file)
> ```
> 
> I am unable to retrieve the in_file from the print statement and instead I'm left with "Inputs has no attribute 'in_file'" even though I define it as a path to a file in a previous line. I will need this in order to fill the 'at_location' field in entity by getting the abs. path of the in_file.
> 
I don't see `input_spec` in your example, so `in_file` is not treated as an input, we talked about it last time

rcali21:
Will add accompanying pytest for   [d0770b4](https://github.com/nipype/pydra/pull/593/commits/d0770b41ff7ef4be009e4f116e7d45c073dd2015) later today.

djarecka:
I think your current tests stopped working, it might be good to figure out this first

pydra/engine/audit.py
```diff
@@ -214,15 +214,15 @@ def audit_task(self, task):
         start_message = {
             "@id": self.aid,
             "@type": "task",
-            "Label": label,
+            "Activity_Label": label,
```
rcali21, line 217: @djarecka You will see that I have change "Label" to "Acitivty_Label" and "Entity_Label", respecitvely. I'm not sure if there exists a means of parsing JSON with duplicate key entries (i.e., two "Labels"). This is my proposed workaround in the meantime.

rcali21: Note that the duplicate entry is what is in the BEP document, so it will need to be changed there as well.

djarecka: what do you mean by duplicate entry?

pydra/engine/audit.py
```diff
@@ -214,15 +214,15 @@ def audit_task(self, task):
...
             "Command": command,
             "StartedAtTime": now(),
             "AssociatedWith": version_cmd,
         }
 
         entity_message = {
             "@id": self.aid,
-            "Label": print(entity_label),
+            "Entity_Label": print(entity_label),
```
djarecka, line 220: I think you don't need `print` function here

pydra/engine/audit.py
```diff
@@ -173,15 +173,20 @@ def audit_task(self, task):
...
             command = task.cmdline
         # assume function task
         else:
             command = None
-        # if hasattr(task.inputs, "in_file"):
-        #     input_file = task.inputs.in_file
-        # else:
-        #     input_file = None
+
+        if hasattr(task.inputs, "in_file"):
```
djarecka, line 184: remember that inputs can have more inputs that are file, not only `in_file`

pydra/engine/audit.py
```diff
@@ -204,15 +209,24 @@ def audit_task(self, task):
...
             "@type": "task",
-            "Label": label,
+            "Activity_Label": label,
             "Command": command,
             "StartedAtTime": now(),
             "AssociatedWith": version_cmd,
         }
 
+        entity_message = {
+            "@id": self.aid,
```
djarecka, line 219: I don't think you should be using the same id here as for activity

djarecka:
you fixed the tests, nice! you should add more tests for entity now, I also left some comments, we can discuss them later today if something is not clear

rcali21:
@djarecka What are your thoughts on adding a test that uses a task that requires file manipulation? Would be helpful for me in further development. Is it best to use a nifti file living on the repo like in nipype, or should we just create a scratch .txt file and then delete it each run?

djarecka:
> @djarecka What are your thoughts on adding a test that uses a task that requires file manipulation? Would be helpful for me in further development. Is it best to use a nifti file living on the repo like in nipype, or should we just create a scratch .txt file and then delete it each run?

I'm not sure if I understand what do you mean. You mean manipulation in place (i.e. saving to the same location after manipulation)? That is consider not the best practice, so I wouldn't do it for the test, but I'm not sure if this is really what you want, don't see any reason why that would be important to you...
Or do you mean just having a test that takes file as an input and creates another file as an output? That would definitely make sense! For these tests I would suggest some text files, are always smaller.

rcali21:
@djarecka When you have a moment, could you look over the most recent commits? Thank you, Dorota.

pydra/engine/audit.py
```diff
@@ -173,15 +174,21 @@ def audit_task(self, task):
...
-        # else:
-        #     input_file = None
+        entity_label = type(label)
+
+        command = task.cmdline if hasattr(task.inputs, "executable") else None
+
+        attr_list = attr_fields(task.inputs)
+        for attrs in attr_list:
+            if attrs.type in [File, Directory]:
+                input_name = attrs.name
```
djarecka, line 184: it will work well if only one element is either `File` or `Directory`, but if you have more than one element you will be overwriting `input_name`, `input_path`, etc. You have to store these variables for each element.

pydra/engine/audit.py
```diff
@@ -173,15 +174,21 @@ def audit_task(self, task):
...
+                input_path = os.path.abspath(getattr(task.inputs, input_name))
+                file_hash = hash_file(input_path)
+
+            else:
```
djarecka, line 188: I believe this is not needed. For now, we said that we will not create any entity for inputs that are not files or directories

pydra/engine/audit.py
rcali21: Thanks for the response. I'm a bit confused. I attempted to implement a list of File/Directories and then iterate the list but it didn't work as expected.

djarecka: Creating a list to iterate over later would be fine, but I don't see where you create the list you're writing about. You should probably create one before l.182, but I don't see any...

pydra/engine/audit.py
```diff
@@ -218,13 +224,16 @@ def audit_task(self, task):
         }
         entity_id = f"uid:{gen_uuid()}"
         entity_message = {
```
djarecka, line 226: you should create entity message for each file from the input

rcali21: Do you recommend putting everything into a function and then calling it for each input?

rcali21: Disregard the last message, I believe I fixed it.

pydra/engine/audit.py
```diff
@@ -173,15 +174,30 @@ def audit_task(self, task):
         import subprocess as sp
 
         label = task.name
+        entity_label = type(label)
```
djarecka, line 177: isn't that always string? why do you take `type(label)`?

pydra/engine/audit.py
```diff
@@ -173,15 +174,30 @@ def audit_task(self, task):
...
+        attr_list = attr_fields(task.inputs)
+        for attrs in attr_list:
+            if attrs.type in [File, Directory]:
+                input_name = attrs.name
+                input_path = os.path.abspath(getattr(task.inputs, input_name))
+                file_hash = hash_file(input_path)
+                entity_id = f"uid:{gen_uuid()}"
+                entity_message = {
+                    "@id": entity_id,
+                    "Label": print(entity_label),
```
djarecka, line 194: why to you call `print` function here?

pydra/engine/audit.py
```diff
@@ -173,15 +174,30 @@ def audit_task(self, task):
...
+                    "AtLocation": input_path,
+                    "GeneratedBy": "test",
```
djarecka, line 196: since for now you don't cover workflows, you can have None here

pydra/engine/tests/test_task.py
```diff
@@ -998,18 +1007,28 @@ def testfunc(a: int, b: float = 0.1) -> ty.NamedTuple("Output", [("out", float)]
...
+            if "@type" in data:
                 if "AssociatedWith" in data:
-                    assert None == data["AssociatedWith"]
-    assert any(json_content)
+                    assert "testfunc" in data["Label"]
+
+            if "@type" in data:
+                if data["@type"] == "input":
+                    assert None == data["Label"]
+                    # placeholder for atlocation until
```
djarecka, line 1022: is this test finished or not?

pydra/engine/tests/test_task.py
```diff
@@ -1028,21 +1047,92 @@ def test_audit_shellcommandtask(tmpdir):
...
     command_content = []
 
     for file in glob(str(message_path) + "/*.jsonld"):
         with open(file, "r") as f:
             data = json.load(f)
-            print(data)
-            if "Label" in data:
-                label_content.append(True)
+
+            if "@type" in data:
```
djarecka, line 1057: I believe all messages should have `@type`, am I right?

pydra/engine/tests/test_task.py
```diff
@@ -1028,21 +1047,92 @@ def test_audit_shellcommandtask(tmpdir):
...
+        audit_flags=AuditFlag.PROV,
+        messengers=PrintMessenger(),
+    )
+    shelly.cache_dir = tmpdir
+    shelly()
+    message_path = tmpdir / shelly.checksum / "messages"
+    for file in glob.glob(str(message_path) + "/*.jsonld"):
+        with open(file, "r") as f:
+            data = json.load(f)
+            print(file_in)
```
djarecka, line 1131: please remove all prints

pydra/engine/tests/test_task.py
```diff
@@ -1028,21 +1047,92 @@ def test_audit_shellcommandtask(tmpdir):
...
+            if "AtLocation" in data:
+                assert data["AtLocation"] == [file_in, file_in_2]
```
djarecka, line 1133: I'm not sure why you expect list here. But the test pass because it never gets here!

pydra/engine/tests/test_task.py
```diff
@@ -998,18 +1007,28 @@ def testfunc(a: int, b: float = 0.1) -> ty.NamedTuple("Output", [("out", float)]
     funky.cache_dir = tmpdir
     funky()
     message_path = tmpdir / funky.checksum / "messages"
-    # go through each jsonld file in message_path and check if the label field exists
-    json_content = []
+    print(message_path)
+
     for file in glob(str(message_path) + "/*.jsonld"):
```
djarecka, line 1012: `glob(str(message_path) + "/*.jsonld")` is an empty list

pydra/engine/tests/test_task.py
rcali21: WIP

pydra/engine/tests/test_task.py
rcali21: WIP

pydra/engine/tests/test_task.py
rcali21: WIP

pydra/engine/tests/test_task.py
rcali21: WIP

djarecka:
@rcali21 - thanks for fixing most of the issues, but I'm not sure why you just removed the second file from the input in test `test_audit_shellcommandtask_file`. As we discussed that is a good test that checks what happens when you have multiple files as an input. It just needed to be fixed

rcali21:
> 

Resolved in

rcali21:
> @rcali21 - thanks for fixing most of the issues, but I'm not sure why you just removed the second file from the input in test `test_audit_shellcommandtask_file`. As we discussed that is a good test that checks what happens when you have multiple files as an input. It just needed to be fixed

Resolved in https://github.com/nipype/pydra/pull/593/commits/662ade1d2866e08f7479a6c2f716d16150893d25.

djarecka:
Thanks @rcali21 for fixing the test! I know we were talking about adding output in this PR, but I will merge this now, and output can be done in a separate PR. Also, the connection of input and output to the task will have to be added

PR #594, opened 2022-11-10
Add datalad task

yibeichan:
## Types of changes

## Summary

as discussed with @djarecka, add datalad interface (adopted from mriqc) to pydra task 
## Checklist

        but the test didn't pass locally, got an error `TypeError: __init__() missing 1 required positional argument: 'func'`; will discuss with Dorota through meeting


tclose:
since this needs datalad, I reckon it would be probably be better in it's own task package `pydra-tasks-datalad`

effigies:
Needs to be rewritten in new syntax.

PR #595, opened 2022-11-17
Handling input fields checking for Mandatory and Mutually exclusive fields

ablachair:
## Types of changes

## Summary
We identified a possible issue with Pydra task input fields checking.

Input fields can be marked as '_mandatory_' as well as '_mutually exclusive_'.
If 
1. field_A is mandatory
2. field_B is mandatory
3. field_A and field_B are 'mutually exclusive'

then only field_A or field_B is needed, **not both**.

Nipype follows this behavior, but it seems like `check_fields_input_spec()` of the Pydra engine (pydra/pydra/engine/specs.py) is checking those conditions sequentially, not concurrently.
Compared with `_check_mandatory_inputs()` (nipype/interfaces/base/core.py)

## Note
The unit test might be misplaced. Should it go in `test_shelltask_inputspec.py` ?

## Checklist

### DIFF
--- a/pydra/engine/specs.py
+++ b/pydra/engine/specs.py
@@ -162,10 +162,15 @@ class BaseSpec:
             mdata = fld.metadata
             # checking if the mandatory field is provided
             if getattr(self, fld.name) is attr.NOTHING:
-                if mdata.get("mandatory"):
-                    raise AttributeError(
-                        f"{fld.name} is mandatory, but no value provided"
-                    )
+                if mdata.get("mandatory"):    
+                    # checking if the mandatory field is provided elsewhere in the xor list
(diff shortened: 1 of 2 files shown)


ablachair:
Fixes Australian-Imaging-Service/pipelines#223

djarecka:
@ablachair - thank you so much for the fix!
Perhaps you can indeed move the tests to `test_shelltask_inputspec.py`, but it's not as important.

Could you please add your name to [zenodo file](https://github.com/nipype/pydra/blob/master/.zenodo.json). We haven't kept any specific order, but just keep Satra Ghosh as the last one.

djarecka:
I will merge it now, but please add yourself to zenodo when you have a chance

PR #597, opened 2022-11-23
MNT: Migrate project to flit-scm

ghisvail:
## Types of changes

- Enhancement

## Summary

With `setup.py` soon to become legacy in favour of the more declarative `pyproject.toml`, this PR proposes to migrate the project to [Flit](https://flit.pypa.io/) and perform dynamic versioning with `flit-scm` instead of `versioneer`.

The very same migration has been successfully experimented with in [pydra-nipype1](https://github.com/nipype/pydra-nipype1), which makes me confident the same can be achieved in the main Pydra repository. The main benefit is to future-proof project management – `setup.py` is going away sooner or later, better be safe than sorry when it happens.

## Checklist

- [x] Convert project metadata to `pyproject.toml`
- [x] Convert dynamic versioning to `flit-scm`
- [x] Adjust CI workflows to the new build
- [x] Run the tests through CI


djarecka:
can this allow for installing with `pip install -e .`?

ghisvail:
> can this allow for installing with `pip install -e .`?

Yes, and also optional extras like `pip install -e .[doc]` or `pip install -e .[test]`

pyproject.toml
```diff
@@ -1,6 +1,97 @@
...
+    "pytest-timeout",
+    "codecov",
+    "numpy",
+    "pyld",
+    "psutil",
+    "python-dateutil",
+    "tornado",
+    "boutiques",
+    "pympler",
+]
```
effigies, line 76: It's possible to include other optional-dependencies of the current project:

```suggestion
]
# Aliases
tests = ["pydra[test]"]
docs = ["pydra[doc]"]
all = ["pydra[doc,dev]"]
```

This will ensure that we don't break things for people who were previously using these.

pyproject.toml
```diff
@@ -1,6 +1,97 @@
...
+
+[project.optional-dependencies]
+dask = [
+    "dask",
+    "distributed",
+]
+dev = [
+    "black",
+    "pre-commit",
+]
```
effigies, line 49:
```suggestion
dev = [
    "black",
    "pre-commit",
    "pydra[test]",
]
```

effigies (approved):
Looks great! Thanks so much for taking the time.

A couple suggestions to restore some dropped "extras".

pyproject.toml
ghisvail: I was not aware of this trick actually 👍

pyproject.toml
ghisvail: MNT: Include test dependencies in dev group

pyproject.toml
effigies: Learned it here: https://discuss.python.org/t/aliased-or-aggregated-optional-dependencies/20488/2

PR #601, opened 2022-11-28
FIX: Enable testing with Python 3.11

### LINKED ISSUE #569: DeprecationWarning: The explicit passing of coroutine objects to asyncio.wait() is deprecated since Python 3.8, and scheduled for removal in Python 3.11
Running the test suite triggers this warning (I am on a Python 3.9 environment), which might become a problem for the future release of Python 3.11.

From my limitted understanding, `asyncio.wait` now needs to be fed with tasks explicitly created with `asyncio.create_task` instead of raw coroutines declared with `async def`.

I could not find an obvious fix for it, so I am leaving it to the experts :-)

ghisvail:
Fails due to [asyncio.wait](https://docs.python.org/3.11/library/asyncio-task.html?highlight=asyncio%20wait#asyncio.wait) now forbidding coroutines directly. Those need to be wrapped in tasks first.

Closes #569

### DIFF
--- a/.github/workflows/testpydra.yml
+++ b/.github/workflows/testpydra.yml
@@ -17,6 +17,9 @@ jobs:
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+        exclude:
+          - os: 'windows-latest'
+            python-version: '3.11'
       fail-fast: false
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/testpydra.yml
+++ b/.github/workflows/testpydra.yml
@@ -30,7 +30,9 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - name: Update build tools
-      run: python -m pip install --upgrade pip build
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install build wheel
 
     - name: Build pydra
       run: python -m build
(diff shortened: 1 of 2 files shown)


.github/workflows/testpydra.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
-        python-version: [3.7, 3.8, 3.9, "3.10"]
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
```
effigies, line 19: Windows is waiting on https://github.com/lxml/lxml/pull/356. Maybe we should exclude that from the matrix for now and create a PR to re-include it to remind us to check periodically?

```suggestion
        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
        exclude:
          - os: 'windows-latest'
            python-version: '3.11'
```

ghisvail: It's actually [merged](https://github.com/lxml/lxml/pull/360) and scheduled for release in version `4.9.2`. People are pushing for it but no idea as to when exactly.

ghisvail: TIL, I did not know about this `exclude` option. It worked beautifully.

ghisvail:
Might be worth doing a new release with this fix for those working (or running tests) with Python 3.11.

djarecka:
yes, will do it tonight! Want to add one more thing before release.

djarecka:
@ghisvail - btw. can you please add yourself to the zenodo file, looks like you're still not there

ghisvail:
I am pretty sure both of you asked me at least once to do it 😆 

It's done 👍

PR #609, opened 2022-12-14
MNT: Add git-archive support for setuptools_scm

effigies:
## Types of changes
- Maintenance task: Add and test git-archive support

## Summary
Versioneer permitted `git archive` to store some limited version information in the artifacts, such as those generated by GitHub. setuptools_scm needs special prodding to do this.

This adds and tests the feature. In the process, I've split the build and test stages from the main tests so that we can test different installation methods.

## Checklist

### DIFF
--- a/.git_archival.txt
+++ b/.git_archival.txt
@@ -1,0 +1,4 @@
+node: $Format:%H$
+node-date: $Format:%cI$
+describe-name: $Format:%(describe:tags=true,match=*[0-9]*)$
+ref-names: $Format:%D$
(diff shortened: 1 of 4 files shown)


effigies:
Getting bitten by https://github.com/pytest-dev/pytest/issues/1596. Fortunately we're not trying to use the `--dask` option anywhere in CI, so just wrapping it up in a `try`/`except`.

ghisvail (approved):
Looks good to me. Sad for `dask` but looks like we are not the only one bitten by this issue.

PR #614, opened 2022-12-31
ENH: Add support for argstr formatting of any iterable

### LINKED ISSUE #613: Proper way to format a tuple parameter
I have got an interface which expects a parameter `foo` as a triple with the following format `--foo 1 2 3`

I reached for the following definition as a first attempt:

```python
input_spec = pydra.specs.SpecInfo(
    name="Input",
    fields=[
        (
            "foo",
            ty.Tuple[int, int, int],
            {
                "help_string": "foo triple",
                "argstr": "--foo {foo}",
            },
        ),
    ],
    bases=(pydra.specs.ShellSpec,),
```

Which does not work since it produces `--foo (1, 2, 3)`.

If I use `--foo...` and instantiate the interface with a list instead of tuple then I get `--foo 1 --foo 2 --foo 3`.

If I use `formatter` instead of `argstr` with `"formatter": lambda field: f"--foo {' '.join(field)}"`, I get an error because formatter seem to be  called with each element of the list instead of the entire list (`TypeError: sequence item 0: expected str instance, int found`).

I am probably missing another option, hence this support request 🙏

ghisvail:
See #613 for context. I looked at the code path @effigies suggested where the fix should be implemented.

This PR should facilitate composition of interfaces producing and consuming iterable types, not just lists. Sets and tuples provide additional guarantees which are quite nice, such as uniqueness and size boundness respectively. Since strings and bytes also satisfy the `Iterable` protocol in Python, these need to be discarded explicitly in the instance type check. 

Closes #613

### DIFF
--- a/pydra/engine/task.py
+++ b/pydra/engine/task.py
@@ -458,7 +458,11 @@ class ShellCommandTask(TaskBase):
                 cmd_add.append(argstr)
         else:
             sep = field.metadata.get("sep", " ")
-            if argstr.endswith("...") and isinstance(value, list):
+            if (
+                argstr.endswith("...")
+                and isinstance(value, ty.Iterable)
+                and not isinstance(value, (str, bytes))
+            ):
                 argstr = argstr.replace("...", "")
                 # if argstr has a more complex form, with "{input_field}"
                 if "{" in argstr and "}" in argstr:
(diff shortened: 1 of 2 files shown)


ghisvail:
I took this opportunity to fix a few typos and warnings which PyCharm complained about.

effigies:
LGTM.

ghisvail:
Is it ok if I merge this?

djarecka:
yes, sorry, thank you! I will create a new release during the weekend

Issue #622, opened 2023-02-28
Intermittent workflow failures (>=0.21?)


tclose:
**What version of Pydra are you using?** 0.21
**What were you trying to do/What did you expect will happen?** run a workflow successfully
**What actually happened?** workflow intermittently fails with error

```
message:
E                         Traceback (most recent call last):
E                           File "/opt/miniconda-latest/envs/arcana/lib/python3.11/site-packages/pydra/engine/core.py", line 1124, in _run
E                             await self._run_task(submitter, rerun=rerun)
E                           File "/opt/miniconda-latest/envs/arcana/lib/python3.11/site-packages/pydra/engine/core.py", line 1152, in _run_task
E                             await submitter.expand_workflow(self, rerun=rerun)
E                           File "/opt/miniconda-latest/envs/arcana/lib/python3.11/site-packages/pydra/engine/submitter.py", line 190, in expand_workflow
E                             await task._run(self, rerun=rerun)
E                           File "/opt/miniconda-latest/envs/arcana/lib/python3.11/site-packages/pydra/engine/core.py", line 1124, in _run
E                             await self._run_task(submitter, rerun=rerun)
E                           File "/opt/miniconda-latest/envs/arcana/lib/python3.11/site-packages/pydra/engine/core.py", line 1152, in _run_task
E                             await submitter.expand_workflow(self, rerun=rerun)
E                           File "/opt/miniconda-latest/envs/arcana/lib/python3.11/site-packages/pydra/engine/submitter.py", line 171, in expand_workflow
E                             blocked = _list_blocked_tasks(graph_copy)
E                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E                           File "/opt/miniconda-latest/envs/arcana/lib/python3.11/site-packages/pydra/engine/submitter.py", line 310, in _list_blocked_tasks
E                             blocking.append(pred, ", ".join(matching_name))
E                         TypeError: list.append() takes exactly one argument (2 given)
```

[...]

PR #623, opened 2023-02-28
FIX: Type error in blocking task list, detailed exception message

tclose:
## Types of changes

## Summary

Fixes bug described in https://github.com/nipype/pydra/issues/622

## Checklist

### DIFF
--- a/pydra/engine/tests/test_submitter.py
+++ b/pydra/engine/tests/test_submitter.py
@@ -1,7 +1,5 @@
 from dateutil import parser
 import re
-import shutil
-import attrs
 import subprocess as sp
 import time
 
@@ -579,33 +577,41 @@ def test_sge_no_limit_maxthreads(tmpdir):
     assert job_1_endtime > job_2_starttime
 
 
-@mark.task
-def add_together(x, y):
-    return x + y
-
-
 def test_wf_with_blocked_tasks(tmpdir):
-    dummy_task = add_together(name="dummy", x=1, y=2)
-    lf = LazyField(node=dummy_task, attr_type="output")
-    lf.field = "out"
     wf = Workflow(name="wf_with_blocked_tasks", input_spec=["x"])
-    wf.add(sleep_add_one(name="taska", x=wf.lzin.x))
-    wf.add(sleep_add_one(name="taskb", x=lf))  # wf.taska.lzout.out))
-    # wf.add(sleep_add_one(name="taskc", x=wf.taska.lzout.out))
-    # wf.add(add_together(name="taskd", x=wf.taskb.lzout.out, y=wf.taskc.lzout.out))
+    wf.add(identity(name="taska", x=wf.lzin.x))
+    wf.add(alter_input(name="taskb", x=wf.taska.lzout.out))
+    wf.add(to_tuple(name="taskc", x=wf.taska.lzout.out, y=wf.taskb.lzout.out))
     wf.set_output([("out", wf.taskb.lzout.out)])
 
-    wf.inputs.x = 1
+    wf.inputs.x = A(1)
 
     wf.cache_dir = tmpdir
 
-    with Submitter("serial") as sub:
-        sub(wf)
+    with pytest.raises(Exception, match="graph is not empty,"):
+        with Submitter("serial") as sub:
+            sub(wf)
 
-    # with pytest.raises(Exception) as exc:
-    # wf_graph = wf.graph.copy()
-    # runnable_tasks, _ =  get_runnable_tasks(wf_graph)
-    # assert runnable_tasks == []
-    # blocked = _list_blocked_tasks(wf_graph)
 
-    # assert "graph is not empty" in str(exc)
+class A:
+    def __init__(self, a):
+        self.a = a
+
+    def __hash__(self):
+        return hash(self.a)
+
+
+@mark.task
+def identity(x):
+    return x
+
+
+@mark.task
+def alter_input(x):
+    x.a = 2
+    return x
+
+
+@mark.task
+def to_tuple(x, y):
+    return (x, y)
--- a/pydra/engine/tests/test_submitter.py
+++ b/pydra/engine/tests/test_submitter.py
@@ -1,6 +1,7 @@
 from dateutil import parser
 import re
 import shutil
+import attrs
 import subprocess as sp
 import time
 
@@ -14,10 +15,11 @@ from .utils import (
     gen_basic_wf_with_threadcount_concurrent,
 )
 from ..core import Workflow
-from ..submitter import Submitter
+from ..submitter import Submitter, get_runnable_tasks, _list_blocked_tasks
 from ... import mark
 from pathlib import Path
 from datetime import datetime
+from pydra.engine.specs import LazyField
 
 
 @mark.task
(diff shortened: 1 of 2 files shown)


effigies:
Is it possible to add a test that exercises this branch?

tclose:
I've been trying to think how to do that. How would a create a workflow with a broken execution graph, where some of the nodes are unrunnable? (I can't remember the exact case that motivated me to write that error explanation code)

effigies:
Is it possible to create a loop like nodeA -> nodeB -> nodeA?

tclose:
I tried this

```python


@mark.task
def add_together(x, y):
    return x + y


def test_wf_with_blocked_tasks(plugin_dask_opt, tmpdir):
    wf = Workflow(name="wf_with_blocked_tasks", input_spec=["x"])
    wf.add(add_together(name="taska", x=wf.lzin.x))
    wf.add(add_together(name="taskb", x=wf.taska.lzout.out, y=wf.taska.lzout.out))

    # Create a cyclic loop
    wf.taska.y = wf.taskb.lzout.out
    wf.set_output([("out", wf.taskb.lzout.out)])

    wf.inputs.x = [1]

    wf.cache_dir = tmpdir

    # with pytest.raises(Exception) as exc:
    with Submitter(plugin=plugin_dask_opt) as sub:
        sub(wf)
```

but it produces the following traceback instead


```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pydra/engine/submitter.py:42: in __call__
    self.loop.run_until_complete(self.submit_from_call(runnable, rerun))
/usr/local/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/base_events.py:649: in run_until_complete
    return future.result()
pydra/engine/submitter.py:68: in submit_from_call
    await runnable._run(self, rerun=rerun)
pydra/engine/core.py:1125: in _run
    result.output = self._collect_outputs()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pydra.engine.core.Workflow object at 0x10eaf4ca0>

    def _collect_outputs(self):
        output_klass = make_klass(self.output_spec)
        output = output_klass(**{f.name: None for f in attr.fields(output_klass)})
        # collecting outputs from tasks
        output_wf = {}
        for name, val in self._connections:
            if not isinstance(val, LazyField):
                raise ValueError("all connections must be lazy")
            try:
                val_out = val.get_value(self)
                output_wf[name] = val_out
            except (ValueError, AttributeError):
                output_wf[name] = None
                # checking if the tasks has predecessors that raises error
                if isinstance(getattr(self, val.name)._errored, list):
>                   raise ValueError(
                        f"Tasks {getattr(self, val.name)._errored} raised an error"
                    )
E                   ValueError: Tasks ['taska'] raised an error
```

djarecka:
@tclose - do you remember if the changes indeed helped at the time you had the issue?

tclose:
@djarecka The changes were what I used to track down and debug the issue I was having. It was something like one node failing and it not getting picked up and then downstream nodes sitting there blocked, but I can't remember what exactly now

tclose:
I expect it would help a lot to debug the intermittent failures I'm experiencing atm too

tclose:
Why is codecov complaining the coverage drops by 5% when I only changed two lines that aren't getting hit?

effigies:
> Why is codecov complaining the coverage drops by 5% when I only changed two lines that aren't getting hit?

Because codecov is garbage.

djarecka:
sorry, but it's still not completely clear to me  how this helps with debugging, since you're commenting `get_runnable_tasks(graph_copy)` and suggesting to   "Uncomment to debug".

tclose:
> sorry, but it's still not completely clear to me how this helps with debugging, since you're commenting `get_runnable_tasks(graph_copy)` and suggesting to "Uncomment to debug".

sorry, that isn't very clear. It is just that that exception is triggered by `get_runnable_tasks` returning no runnable tasks, so if you uncomment that line put a breakpoint there you can step through and find out what the problem is. I forgot to take it out before making my previous PR, but thought it might be useful to keep it there as a reference. Happy to take it out though

tclose:
I've tried a few ways to hit this branch but keep getting stuck. If we merge these changes, I should be able to have a better idea of what the issue that is causing this branch to be hit in my code and then could try to design a test around that.

tclose:
Ah, I remember now. I fixed this bug in my local branch and managed to hit the issue again. It is caused by the hash of the task being altered at some point in the workflow (somewhere deep in a nested object I imagine) and then the downstream node not being able to find the predecessor node at the hash it expects to.

The error message helped a lot :) Still not sure how to generate a test for it easily though.

effigies:
Do we coerce inputs to a different type at runtime? That might change the hash.

tclose:
> Do we coerce inputs to a different type at runtime? That might change the hash.

I don't remember seeing that. I'm passing around fairly deep objects, which was not really what I expect Pydra was designed for, so I ran into trouble before where one of nested objects was not being copied between nodes and a downstream node updated the object that was used to generate the upstream nodes hash.

tclose:
Using the insight (see #622) that this error is caused by hashes of node inputs being unstable, I have managed to come up with a test to hit this error.

Seems as though this is something that needs a bit deeper thought though, as relying on hashes not changing is a bit of a weakness.

tclose:
@djarecka @effigies , I reckon this is ready to merge now. Not sure why the SLURM test is failing. If you go to Codecov, the changes actually increase the coverage by 0.55% not reduce them

pydra/engine/tests/test_submitter.py
```diff
@@ -575,3 +575,43 @@ def test_sge_no_limit_maxthreads(tmpdir):
...
+
+
+@mark.task
+def identity(x):
+    return x
+
+
+@mark.task
+def alter_input(x):
+    x.a = 2
```
effigies, line 611: Interesting reproduction. Here's a thought for how we could change this to make this use case smoother:

User writes a function that modifies its inputs. After execution, the runner re-checks the input hash and raises an error on change, rather than waiting for a workflow deadlock. The user then re-marks with `@mark.unsafe_task`, which causes the inputs to be `deepcopy`ed before execution. The input hash remains unchanged, and the workflow continues.

This doesn't resolve the hash nondeterminism. That's caused by https://github.com/nipype/pydra/blob/b5fe4c0eb7f937e70db15bc087d86fe90f401ff3/pydra/engine/helpers.py#L672-L674

Given that it's based on the string representation, you could probably resolve that on your end by sorting the set in `__str__` or `__repr__`. 

That said, there's a [StackOverflow](https://softwareengineering.stackexchange.com/questions/405243/how-to-perform-consistent-hashing-on-any-python-object-that-works-with-hash) on this exact problem, and the answer is basically that there's no way to hash arbitrary objects. But we could use `functools.singledispatch` to pass this problem off to users:

```Python
@functools.singledispatch
def hash_obj(obj: object) -> bytes:
    # Works for generic objects with __dict__
    dict_rep = ":".join(":".join(key, hash_obj(val)) for key, val in obj.__dict__.items())
    return sha256(f"{obj.__class__}:{dict_rep}".encode()).hexdigest() 

# Then define something for builtin classes with no __dict__
@hash_obj.register
def _(obj: int) -> bytes: ...

@hash_obj.register
def _(obj: str) -> bytes: ...

@hash_obj.register
def _(obj: dict) -> bytes: ...
```

We could even handle things like `set()` and `frozenset()` by default, and so only types that were really obscure would need to register their own. And we could either just let them decorate:


```Python
@pydra.util.hash_obj.register
def _(obj: File) -> bytes:
    ...
```

or provide a registration function

```Python
pydra.utils.register_hash(File, myhashfun)
```

djarecka:
yes, this is interesting case. I think I was assuming that input should not changed and I did not include tests that do it... I like the approach @effigies is suggesting: checking the input after execution. 

As for the issue with hashes, I've already started adding some extra rules for specific subjects, e.g. for the [numpy arrays](https://github.com/nipype/pydra/blob/b5fe4c0eb7f937e70db15bc087d86fe90f401ff3/pydra/engine/helpers.py#L702), after I had issues. But giving option to the user might ba a good solution (this probably can be in a different PR)

effigies (approved):
We should probably move these discussions out of a PR that's ready to merge...

effigies:
Created #625 and #626.

Issue #619, opened 2023-02-09
Combining `requires` and `output_file_template` for a field


ghisvail:
I am trying to combine the effects of `requires` and `output_file_template` but failed so far.

Here is my latest attempt:

```python
@attrs.define(kw_only=True)
class MyShellSpec(pydra.specs.ShellSpec):
    mandatory_arg: str = attrs.field(
        metadata={"help_string": "mandatory arg", "mandatory": True}
    )

    enable_bar: bool  = attrs.field(
        metadata={"help_string": "optional support for bar", "argstr": "-a"}
    )

    output_arg: str = attrs.field(
        metadata={
            "help_string": "output arg",
            "argstr": "-o",
            "output_file_template": "{mandatory_arg}_foo",
        }
    )

    maybe_output_arg: str = attrs.field(
        metadata={
            "help_string": "output arg if optional arg provided",
            "argstr": "-m",
            "output_file_template": "{mandatory_arg}_bar",
            "requires": {"enable_bar"},
        }
    )  
```

where I would want `maybe_output_arg` to get a templated default value **only** if an optional argument is provided.
The reasoning behind it is that the underlying executable generates additional files based on whether some optional parameters have been provided.

Assuming this is not feasible,

- Is this something we would want to allow?
- If not, what's the alternative pattern to achieve the same behavior?

Cheers,
Ghis

*EDIT: Fix typo in code snippet*

satra:
@ghisvail - i don't see `optional_arg` anywhere. perhaps there should at least be something that checks that variables in requires are present in the spec.

ghisvail:
My bad, optional -> enable_bar

Le jeu. 9 févr. 2023, 20:48, Satrajit Ghosh ***@***.***> a
écrit :

ghisvail:
> perhaps there should at least be something that checks that variables in requires are present in the spec.

Indeed :sweat_smile:, this mistake was not intended but you're right `requires` should only reference a field present in the spec.

I fixed the code snippet.

satra:
i would say this is likely a order of operations resolution problem in the spec processor. do you have the inclination to send a PR ?

ghisvail:
> do you have the inclination to send a PR ?

I can have a look

PR #627, opened 2023-03-09
REF: More detailed runtime checks for input spec

ghisvail:
Whilst working on #619, I had a hard time figuring out how runtime checks were working in the input spec.

I took this opportunity to break the checks down into more digestible chunks, with intermediate variables and more meaningful names. In addition, we are now getting the list of alternative fields to set in case a mandatory argument is unset but declare other fields in `xor`, instead of just reporting the field as mandatory but unset.

On a more general note, I would have expected to find these checks in `ShellSpec`, since the specific metadata we are manipulating only make sense in this context. Food for thoughts for a wider refactoring maybe.

## Types of changes

- Refactoring (non-breaking change, does not fix an issue)

## Checklist

### DIFF
--- a/pydra/engine/specs.py
+++ b/pydra/engine/specs.py
@@ -156,55 +156,69 @@ class BaseSpec:
 
         """
         fields = attr_fields(self)
-        names = []
-        require_to_check = {}
-        for fld in fields:
-            mdata = fld.metadata
-            # checking if the mandatory field is provided
-            if getattr(self, fld.name) is attr.NOTHING:
-                if mdata.get("mandatory"):
-                    # checking if the mandatory field is provided elsewhere in the xor list
-                    in_exclusion_list = mdata.get("xor") is not None
-                    alreday_populated = in_exclusion_list and [
-                        getattr(self, el)
-                        for el in mdata["xor"]
-                        if (getattr(self, el) is not attr.NOTHING)
-                    ]
-                    if (
-                        alreday_populated
-                    ):  # another input satisfies mandatory attribute via xor condition
-                        continue
+
+        for field in fields:
+            field_is_mandatory = bool(field.metadata.get("mandatory"))
+            field_is_unset = getattr(self, field.name) is attr.NOTHING
+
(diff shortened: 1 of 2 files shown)


pydra/engine/specs.py
```diff
@@ -156,55 +156,69 @@ def check_fields_input_spec(self):
...
-                        if (getattr(self, el) is not attr.NOTHING)
-                    ]
-                    if (
-                        alreday_populated
-                    ):  # another input satisfies mandatory attribute via xor condition
-                        continue
+
+        for field in fields:
+            field_is_mandatory = bool(field.metadata.get("mandatory"))
+            field_is_unset = getattr(self, field.name) is attr.NOTHING
```
effigies, line 162: Optimization: Do not build alternatives if no problem exists.

```suggestion
            field_is_unset = getattr(self, field.name) is attr.NOTHING
            
            if not field_is_mandatory or not field_is_mandatory:
                continue
```

The `continue` on L198 does this after constructing `alternative_fields` and `required_fields`.

pydra/engine/specs.py
```diff
@@ -156,55 +156,69 @@ def check_fields_input_spec(self):
...
+            required_fields = {
+                name: getattr(self, name) is not attr.NOTHING
+                for name in field.metadata.get("requires", [])
+                if name != field.name
+            }
+
+            # Raise error if field is mandatory and unset
+            # or no suitable alternative is provided.
+            if field_is_unset:
+                if field_is_mandatory:
```
effigies, line 181: Using the above suggestion, these checks can be assumed `True`.

effigies:
Deleted some bad suggestions. Will do normal review so you don't see the noise.

ghisvail:
> The logic looks good. I have some suggestions that I put in a commit that I think make it a little easier to follow (https://github.com/nipype/pydra/commit/2eb09b77e010525579cdc6324b929a260c0addb3).

I like them. I added them on your behalf in a follow-up commit.

PR #629, opened 2023-03-13
FIX: Provide templated fields to cmdline only when requirements are met

### LINKED ISSUE #619: Combining `requires` and `output_file_template` for a field

ghisvail:
## Types of changes

## Summary

Issue #619 showed an example where a templated field can be provided in the generated command-line string,
despite being guarded by another boolean option via `requires`. This PR ensures that the helper function responsible for computing interpolated fields only consider templated fields for which all their required fields are set.

## Checklist

Closes #619

### DIFF
--- a/pydra/engine/helpers_file.py
+++ b/pydra/engine/helpers_file.py
@@ -591,9 +591,19 @@ def template_update(inputs, output_dir, state_ind=None, map_copyfiles=None):
 
     from .specs import attr_fields
 
+    # Collect templated inputs for which all requirements are satisfied.
     fields_templ = [
-        fld for fld in attr_fields(inputs) if fld.metadata.get("output_file_template")
+        field
+        for field in attr_fields(inputs)
+        if field.metadata.get("output_file_template")
+        and all(
+            [
+                getattr(inputs, required_field) is not attr.NOTHING
+                for required_field in field.metadata.get("requires", [])
+            ]
+        )
     ]
+
     dict_mod = {}
     for fld in fields_templ:
         if fld.type not in [str, ty.Union[str, bool]]:
--- a/pydra/engine/helpers_file.py
+++ b/pydra/engine/helpers_file.py
@@ -597,10 +597,8 @@ def template_update(inputs, output_dir, state_ind=None, map_copyfiles=None):
         for field in attr_fields(inputs)
         if field.metadata.get("output_file_template")
         and all(
-            [
-                getattr(inputs, required_field) is not attr.NOTHING
-                for required_field in field.metadata.get("requires", [])
-            ]
+            getattr(inputs, required_field) is not attr.NOTHING
+            for required_field in field.metadata.get("requires", ())
         )
     ]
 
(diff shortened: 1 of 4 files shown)


pydra/engine/helpers_file.py
```diff
@@ -591,9 +591,19 @@ def template_update(inputs, output_dir, state_ind=None, map_copyfiles=None):
...
     fields_templ = [
-        fld for fld in attr_fields(inputs) if fld.metadata.get("output_file_template")
+        field
+        for field in attr_fields(inputs)
+        if field.metadata.get("output_file_template")
+        and all(
+            [
+                getattr(inputs, required_field) is not attr.NOTHING
+                for required_field in field.metadata.get("requires", [])
+            ]
```
effigies, line 603: Two slight optimizations:

1) Do not construct a list. `all()` will short-circuit and avoid even finishing iterating over `field.metadata['requires']` the first time it hits a `False` value.
2) `()` is often a better empty iterable than `[]`, since tuples are immutable and the interpreter can optimize by having one global empty tuple rather than constructing a new empty list.

```suggestion
            getattr(inputs, required_field) is not attr.NOTHING
            for required_field in field.metadata.get("requires", ())
```

effigies (approved):
This makes sense to me. We should get a different failure if `requires` are missing and the field is mandatory.

Is there any chance of documenting this? It would be nice to have a cookbook of specs showing how to achieve goals like conditional command line templates.

ghisvail:
> We should get a different failure if requires are missing and the field is mandatory.

I think there is a wider discussion to be had on `requires` (and maybe `xor`). For instance, FSL's Eddy features requirements for some options based on a minimum value on a dependent field. Kind of like a feature flag followed by feature specific flags (here it's an optional processing step enabled when the model order is higher than 1).

I have a feeling a more general form of `requires` could be:

`requires: mapping[str, Callable[T, bool]] | T | None] | iterable[str] | None`

where the current form `iterable[str]` stands for all fields are defined, and the more general form `mapping[...]` means all fields are defined and satisfy a runtime check against a constant or a callable, if defined.

> Is there any chance of documenting this? It would be nice to have a cookbook of specs showing how to achieve goals like conditional command line templates.

I agree that a cookbook of specs would be welcome. We have a few ones in the test suite but they are pretty abstract. I think once the model for `requires` (and maybe `xor`) is well specified, I would feel more confident in working on a cookbook-like extension to the docs.

PR #631, opened 2023-03-21
FIX: Allow any iterable for ShellOutSpec requires

ghisvail:
## Types of changes

## Summary

Specifying the `requires` metadata for a field in `ShellOutSpec` as a `set` instead of a `list` bypasses part of the metadata checker resulting in an Exception further down the implementation.

This fix allow to specify the `requires` metadata for `ShellOutSpec` as any iterable (including a set), instead of being restricted to a `list`. It's then explicitly transformed to a list of list by the metadata checker, if necessary.

### DIFF
--- a/pydra/engine/specs.py
+++ b/pydra/engine/specs.py
@@ -636,7 +636,7 @@ class ShellOutSpec:
 
         if "requires" in fld.metadata:
             # if requires is a list of list it is treated as el[0] OR el[1] OR...
-            required_fields = list(fld.metadata["requires"])
+            required_fields = ensure_list(fld.metadata["requires"])
             if all([isinstance(el, list) for el in required_fields]):
                 field_required_OR = required_fields
             # if requires is a list of tuples/strings - I'm creating a 1-el nested list
--- a/pydra/engine/specs.py
+++ b/pydra/engine/specs.py
@@ -636,11 +636,12 @@ class ShellOutSpec:
 
         if "requires" in fld.metadata:
             # if requires is a list of list it is treated as el[0] OR el[1] OR...
-            if all([isinstance(el, list) for el in fld.metadata["requires"]]):
-                field_required_OR = fld.metadata["requires"]
+            required_fields = list(fld.metadata["requires"])
+            if all([isinstance(el, list) for el in required_fields]):
+                field_required_OR = required_fields
             # if requires is a list of tuples/strings - I'm creating a 1-el nested list
-            elif all([isinstance(el, (str, tuple)) for el in fld.metadata["requires"]]):
-                field_required_OR = [fld.metadata["requires"]]
+            elif all([isinstance(el, (str, tuple)) for el in required_fields]):
+                field_required_OR = [required_fields]
             else:
                 raise Exception(
                     f"requires field can be a list of list, or a list "


pydra/engine/specs.py
```diff
@@ -636,11 +636,12 @@ def _check_requires(self, fld, inputs):
 
         if "requires" in fld.metadata:
             # if requires is a list of list it is treated as el[0] OR el[1] OR...
-            if all([isinstance(el, list) for el in fld.metadata["requires"]]):
-                field_required_OR = fld.metadata["requires"]
+            required_fields = list(fld.metadata["requires"])
```
effigies, line 639: What if we get a string? This will turn `"field"` into `['f', 'i', 'e', 'l', 'd']`.

```suggestion
            required_fields = ensure_list(fld.metadata["requires"])
```

Or, if we don't want to allow this:

```suggestion
            if isinstance(fld.metadata["requires"], str):
                raise ValueError(...)
            required_fields = list(fld.metadata["requires"])
```

pydra/engine/specs.py
```diff
@@ -636,11 +636,12 @@ def _check_requires(self, fld, inputs):
...
+            if all([isinstance(el, list) for el in required_fields]):
+                field_required_OR = required_fields
             # if requires is a list of tuples/strings - I'm creating a 1-el nested list
```
effigies, line 642: I'm a little confused by this one. The type checking is divorced from the semantics. Why should this be a list of lists of tuples or strings? It's hard to reason about the effects of converting an arbitrary iterable to a list when the meaning of the tuples and strings (both iterables) are opaque.

Instead of having this here in this function, I'd rather have a separate function:

```Python
def canonicalize_requires(requires):
    """Canonicalize a requirement to a nested list of lists of requirements

    Fully specified requirements are a list of lists of strings or tuples that have the form
    ``[["abc", "def"], ...]`` or ``[[("abc", "def"), ("ghi", "jkl")], ...]``. For brevity, we permit the following
    forms.

    >>> # some example code

    Non-list iterables are also accepted:

    >>> canonicalize_requirements({"abc", "def"})
    [["abc", "def"]]
    """
    ...
```

ghisvail: I did not want to change the semantics too much in this PR so I did not attempt to refactor the code too much. The code flow could be improved.

The idea remains that I should be able to do:

```python
metadata={
    ...,
    "requires": {"foo", "bar"},
}
```

or

```python
metadata={
    ...,
    "requires": ["foo", "bar"],
}
```

and the most straightfoward way was to normalize `Iterable` to `list` internally first.

pydra/engine/specs.py
ghisvail: Since `"requires": "foo"` makes sense, maybe we could allow it? Then, `ensure_list` would be the solution.

effigies (approved):
If you want to keep refactoring/modifying semantics for a separate PR, that's fine by me.

djarecka:
`AND` is covered by default when you have multiple elements, e.g. [here](https://github.com/nipype/pydra/blob/master/pydra/engine/tests/test_shelltask.py#L3605). 

So if you have:
```
"requires": [
     ["file1", "additional_inp_A"],
     ["file1", "additional_inp_B"],
  ],
```
That means ("file1" AND "additional_inp_A") OR ("file1" AND "additional_inp_B")

PR #632, opened 2023-03-22
FIX: Allow staticmethod be passed to callable metadata in ouput spec

ghisvail:
## Types of changes

## Summary

This PR provides a fix allowing the `callable` field metadata be defined as a staticmethod to the output spec class.

Before this fix, the following would fail:

```python

@attrs.define(kw_only=True)
class MyOutSpec(pydra.specs.ShellOutSpec):

    @staticmethod
    def get_foo(field):
        return "foo.bar"

    foo: pydra.specs.File = attrs.field(
        metadata={
            help="some output help",
            callable=get_foo,
        }
    )
```

The reason for that is that the actual callable is wrapped inside the `staticmethod` descriptor and does not expose the `Callable` interface, which makes runtime reflection with `inspect` fail with a `TypeError`. The solution is just to recover the wrapped method inside the descriptor.
 
## Checklist

### DIFF
--- a/pydra/engine/specs.py
+++ b/pydra/engine/specs.py
@@ -593,7 +593,12 @@ class ShellOutSpec:
                     return attr.NOTHING
                 return val
         elif "callable" in fld.metadata:
-            call_args = inspect.getfullargspec(fld.metadata["callable"])
+            callable_ = fld.metadata["callable"]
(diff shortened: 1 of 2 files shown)


satra:
@ghisvail - can you add your example as a test?

PR #633, opened 2023-03-23
CI: Add dependabot config

ghisvail:
Add [dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file) configuration watching updates to GitHub actions.


effigies:
Do we have some tests that only run on master, so PRs will always look like regressions?

PR #644, opened 2023-05-02
ENH: Add generic object hasher

### LINKED ISSUE #626: Centralize object hashing and provide a mechanism for types to register a hash
Right now we have hashing split up in a few places:

https://github.com/nipype/pydra/blob/b5fe4c0eb7f937e70db15bc087d86fe90f401ff3/pydra/engine/helpers.py#L677-L708

https://github.com/nipype/pydra/blob/b5fe4c0eb7f937e70db15bc087d86fe90f401ff3/pydra/engine/helpers.py#L672-L674

https://github.com/nipype/pydra/blob/b5fe4c0eb7f937e70db15bc087d86fe90f401ff3/pydra/engine/helpers_file.py#L70-L168

An alternative approach could be to use `functools.singledispatch`:

```Python
@functools.singledispatch
def hash_obj(obj: object) -> bytes:
    # Works for generic objects with __dict__
    dict_rep = ":".join(":".join(key, hash_obj(val)) for key, val in obj.__dict__.items())
    return sha256(f"{obj.__class__}:{dict_rep}".encode()).hexdigest() 
```

This defines a cryptographic hash for a generic object that applies recursively. We would need some bottom types that don't have `__dict__`:

```Python
@hash_obj.register
def _(obj: int) -> bytes: ...

@hash_obj.register
def _(obj: str) -> bytes: ...

@hash_obj.register
def _(obj: dict) -> bytes: ...
```

And each type would be able to declare how much is needed to uniquely identify it across instances. We could add `set()` and `frozenset()` to ensure that these known-problematic builtin types are consistent. And then provide a means for a downstream tool to register a type with our hasher, such as:

```Python
@pydra.util.hash_obj.register
def _(obj: MyType) -> bytes:
    ...
```
or

```Python
pydra.utils.register_hash(MyType, myhashfun)
```

effigies:
## Types of changes

## Summary
This is an initial proposal for how to generically hash an object. The basic scheme is to create a single hash object that accepts byte-strings, and then iterates over `bytes` representations of components of the object to be hashed.

The job of each object is then not to tell how to hash itself but to provide a stable representation of itself in `bytes` that is unique. By creating generators, it is not necessary to keep the full `bytes` representation of the object in memory at any time, which will be critical if file contents are to be hashed.

I propose to use `blake2b` because it achieves MD5-like speed and we can use custom digest sizes to keep digests readable. Blake2 also has a [tree mode](https://docs.python.org/3/library/hashlib.html#tree-mode), where multiple hashes can be combined. This may be useful for computing an interface hash that is created from the hashes of all inputs, but I have not attempted to think through this yet.

There is a pathological case for repeated objects:

```Python
In [101]: s = {1,2,3}

In [103]: t = {4, 5, 6}

In [104]: hash_object({'a': s, 'b': t, 'c': s})
Out[104]: b'G\x8a\xa0:\x0e\xd4\xc4>2\xc9F\xedz\x9ax\x1b'

In [105]: hash_object({'a': s, 'b': t, 'c': t})
Out[105]: b'G\x8a\xa0:\x0e\xd4\xc4>2\xc9F\xedz\x9ax\x1b'
```

This would be mitigated by mapping objects directly to hashes.

Closes #626.

## Checklist

### DIFF
--- a/pydra/engine/helpers.py
+++ b/pydra/engine/helpers.py
@@ -7,7 +7,6 @@ from pathlib import Path
 from filelock import SoftFileLock, Timeout
 import os
 import sys
-from hashlib import sha256
 from uuid import uuid4
 import subprocess as sp
 import getpass
@@ -32,6 +31,7 @@ from .specs import (
     MultiOutputFile,
 )
 from .helpers_file import hash_file, hash_dir, copyfile, is_existing_file
+from ..utils.hash import hash_object
 
 
 def ensure_list(obj, tuple2list=False):
@@ -671,7 +671,7 @@ def get_open_loop():
 
 def hash_function(obj):
     """Generate hash of object."""
-    return sha256(str(obj).encode()).hexdigest()
+    return hash_object(obj).hex()
 
 
 def hash_value(value, tp=None, metadata=None, precalculated=None):
--- a/pydra/engine/helpers_file.py
+++ b/pydra/engine/helpers_file.py
@@ -6,11 +6,14 @@ import os
 import os.path as op
 import re
 import shutil
+import stat
 import posixpath
 import logging
 from pathlib import Path
 import typing as ty
 from copy import copy
+
+from ..utils.hash import hash_object
 
 related_filetype_sets = [(".hdr", ".img", ".mat"), (".nii", ".mat"), (".BRIK", ".HEAD")]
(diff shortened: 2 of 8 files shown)


effigies:
Subsumed by #662.

PR #645, opened 2023-05-03
STY: Miscellaneous cleanups

effigies:
## Types of changes
* Style, slight refactor

## Summary
While looking into type checking failures, some style issues caught my eye. Rather than bundle them in, I am putting them in a standalone PR. This PR has three commits, each of which can be considered separately, if you disagree.

* Replace `__file__` calls with `importlib.resources` where appropriate. I wrapped it in a pytest fixture to save repetition. This is considered good hygiene, as `__file__` may not be available when packages are zipped.
* Drop `f"...{str(x)}..."` idiom. This is a straightforward win for readability.
* Just ran `pyupgrade --py38-plus`, followed by `black` for cleanup. This auto-upgrades from idioms that are no longer needed to ones that are universally supported in 3.8+.

## Checklist


djarecka:
lgtm! thank you!

effigies:
Are these random 3.11 failures expected?

djarecka:
no, i  didn't notice it... and it's not even test failure, but some `INTERNALERROR`...

effigies:
It happened on 99df677708b1a25608b705daadb120202558971d as well, so at least I'm not introducing it with something I've done here. My guess is it's an out-of-memory error causing an xdist worker to get killed.

PR #621, opened 2023-02-27
WIP: Static checking with mypy

ghisvail:
*Work in progress*

See methodology [here](https://mypy.readthedocs.io/en/stable/existing_code.html#introduce-stricter-options).

PR #648, opened 2023-05-03
MNT: Type-annotate Pydra

effigies:
## Types of changes

## Summary
Building off of #621. Interested in feedback as we go.

## Checklist

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -10,7 +10,7 @@ import typing as ty
 from copy import deepcopy
 from uuid import uuid4
 
-import cloudpickle as cp
+import cloudpickle as cp  # type: ignore[import]
 from filelock import SoftFileLock
 import shutil
 from tempfile import mkdtemp
--- a/pydra/engine/helpers.py
+++ b/pydra/engine/helpers.py
@@ -2,7 +2,7 @@
 import asyncio
 import asyncio.subprocess as asp
 import attr
-import cloudpickle as cp
+import cloudpickle as cp  # type: ignore[import]
 from pathlib import Path
 from filelock import SoftFileLock, Timeout
 import os
--- a/pydra/engine/specs.py
+++ b/pydra/engine/specs.py
@@ -8,6 +8,8 @@ from glob import glob
 
 from .helpers_file import template_update_single
 
+T = ty.TypeVar("T")
+
 
 def attr_fields(spec, exclude_names=()):
     return [field for field in spec.__attrs_attrs__ if field.name not in exclude_names]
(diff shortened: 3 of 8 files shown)


effigies:
Okay, it looks like we're going to have fights with type checkers about `File` and `Directory`. If we use `File = NewType("File", Path)`, then Python 3.10+ will happily accept it, but prior to that, there are pickling errors. We can't subclass `Path`, which is a known limitation of Python and won't be fixed until Python 3.12 at a minimum (I'm not sure of the status of that effort). Importing `NewType` from `typing_extensions` doesn't give us 3.10+ behavior in 3.7.

@djarecka What exactly are the semantics of these? It looks like they are labels but not actually instantiable as classes containing paths, so when we write things like

```Python
@task
def dirname(a: File) -> Directory:
    return Path(a).parent
```

We get a type failure. And we can't wrap `Directory(Path(a).parent)`, since `Directory()` doesn't take args.

One possibility could be to just implement a very thin `os.PathLike` interface:

```Python
class FileObj:
    def __init__(self, path: ty.Union[str, os.PathLike]):
        self._path = Path(path)
    def __fspath__(self):
        return str(self._path)
    def __repr__(self):
        return repr(self._path)

class File(FileObj): ...
class Directory(FileObj): ...
```

I *think* this would get us close enough to 3.10-style NewTypes, so we could eventually get rid of these. But it may make more sense to get rid of/deprecate these concepts altogether.

djarecka:
I think `File` and `Directory` was used a bit as placeholders for future class, so we can use something like `FileObj`.

I'm not sure what you mean by "concept" that we might want to get rid of.

effigies:
Sorry, I just meant the classes. I'm not sure how all they're used, so it's unclear what we need to replace if we just delete them.

ghisvail:
I'd propose to define `File` and `Directory` as aliases to `os.PathLike` which is the generic protocol for paths in Python. `File` would represent a pathlike instance pointing to a file, `Directory` to a directory.

Anything satisfying `os.PathLike` can be converted to `pathlib.[Pure]Path` underneath, so it should not be a problem for the internal Pydra engine machinery. Besides, it would satisfy estabished good OOP practices of "committing to an interface, not an implementation", thereby allowing alternatives like `fsspec` to be used as input arguments as well.

ghisvail:
We also need to make sure we would not be accidentally breaking pieces of formatting or callable logic like [this](https://github.com/aramis-lab/pydra-fsl/blob/94e2e9e0c0211b02e4cbb8ddb54212018e506667/src/pydra/tasks/fsl/utils/fslselectvols.py#L57) which relies isinstance / pattern matching on types.

djarecka:
@effigies - I mostly used them in Input/Output spec so it's clear what people expect to get. I also use this type when doing the current hashing of setting the bindings to the container. Not saying that it would be impossible without them, but not sure if I want to remove them completely... Can think more

ghisvail:
Given the major changes recently introduced, do you think it's worth keeping this issue open?

effigies:
No.

PR #651, opened 2023-05-05
RF: Change File and Directory to aliases of Path

effigies:
This attempts to replace File/Directory with TypeAliases, which I think is probably the only reasonable way to make Pydra type-checkable.

The File/Directory stuff is pretty magical. Something somewhere is changing the `output_dir` after using it to do input mapping when running container tasks, but I can't figure out where and I've run out of time to spend on this. Possibly someone else will see it.

### DIFF
--- a/pydra/engine/audit.py
+++ b/pydra/engine/audit.py
@@ -1,10 +1,11 @@
 """Module to keep track of provenance information."""
 import os
 import json
+from pathlib import Path
 import attr
 from ..utils.messenger import send_message, make_message, gen_uuid, now, AuditFlag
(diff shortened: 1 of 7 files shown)


ghisvail:
Just out of curiosity, why can't `File` and `Directory` be aliases to `os.PathLike`?

effigies:
Maybe we can. As a protocol, they're not instantiable, so I didn't want to add the complication, but maybe that will help us be more careful.

effigies:
FWIW, we might be able to go back to `NewType` after the next `typing_extensions` release: https://github.com/python/typing_extensions/issues/156

PR #653, opened 2023-05-11
`cmd_arg` and `cmd_out` syntactic sugar funcs for attrs.field in `ShellSpec`/`ShellOutSpec`

tclose:
## Types of changes

- New feature (non-breaking change which adds functionality)

## Summary
Adds the `pydra.mark.shell_commands.cmd_arg()` and `pydra.mark.shell_commands.cmd_out()` methods to streamline the definition of input/output specs for shell command task definitions. Instead of creating an attrs.field by hand and setting the metadata directly, `cmd_arg` and `cmd_out` do this for you. 

The main benefit is that `cmd_arg` and `cmd_out` specify the valid metadata fields that can be added to the input and output fields, making the code more self-documenting. It also makes the task definitions slightly less verbose.

## Checklist

- [x] I have added tests to cover my changes (if necessary)
- [x] I have updated documentation (if necessary)

PR #655, opened 2023-05-15
`shell_task` decorator 

tclose:
## Types of changes

## Summary
Extends PR #653 to include `@shell_task` decorator, which streamlines the code required to define a shell command task. See discussion https://github.com/nipype/pydra/discussions/647 for rationale behind this approach

## Checklist


tclose:
I have now implemented the `@shell_task` decorator, which can be used as per the discussion thread like

```python

@shell_task
class Ls:
    executable = "ls"

    class Inputs(RecursiveOptions):
        directory: os.PathLike = shell_arg(
            help_string="the directory to list the contents of",
            argstr="",
            mandatory=True,
            position=-1,
        )
        hidden: bool = shell_arg(
            help_string="display hidden FS objects",
            argstr="-a",
            default=False,
        )

    class Outputs:
        entries: list = shell_out(
            help_string="list of entries returned by ls command",
            callable=list_entries,
        )
```

or alternatively as a function in order to dynamically generate interfaces

```python
Ls = shell_task(
    "Ls",
    executable="ls",
    input_fields={
        "directory": {
            "type": os.PathLike,
            "help_string": "the directory to list the contents of",
            "argstr": "",
            "mandatory": True,
            "position": -1,
        },
        "hidden": {
            "type": bool,
            "help_string": "display hidden FS objects",
            "argstr": "-a",
        },
    output_fields={
        "entries": {
            "type": list,
            "help_string": "list of entries returned by ls command",
            "callable": list_entries,
        }
    },
    inputs_bases=[RecursiveOptions],
)
```

The decorator/function automatically generates the `input_spec` and `output_spec` `SpecInfor` class attributes, but I think they could be redundant now, as the attrs-generated classes for inputs and outputs are generated at `MyTask.Inputs` and `MyTask.Outputs`, respectively, with `output_file_template` generated outputs automatically inserted.

If this syntax looks alright I can update the docs to match it.

If there is interest, I could also look at refreshing the function `task` decorator so that it can be used in the same way for dynamic classes and so it also generates `Inputs`/`Outputs` classes so we can do without `input_spec` and `output_spec` all together.

pydra/mark/tests/test_shell.py
```diff
@@ -0,0 +1,467 @@
...
+    assert "y_out" in [a.name for a in attrs.fields(A.Outputs)]
+
+
+def get_file_size(y: Path):
+    result = os.stat(y)
+    return result.st_size
+
+
+def test_shell_bases_dynamic(A, tmpdir):
+    B = shell_task(
```
djarecka, line 282: so I understand that this has the same `input_spec` and `executable`/`args` as `A`?

Could we also overwrite `executable` or `args`?

pydra/mark/tests/test_shell.py
```diff
@@ -0,0 +1,467 @@
...
+    xpath = tmpdir / "x.txt"
+    ypath = tmpdir / "y.txt"
+    Path.touch(xpath)
+
+    b = B(x=xpath, y=str(ypath))
+
+    result = b()
+
+    assert b.inputs.x == xpath
+    assert result.output.y == str(ypath)
```
djarecka, line 303: perhaps we should also check `result.output.out_file_size` to check if this is correctly added to the `output_spec`

pydra/mark/tests/test_shell.py
```diff
@@ -0,0 +1,467 @@
...
```
djarecka, line 323: see comment in the previous test

pydra/mark/tests/test_shell.py
```diff
@@ -0,0 +1,467 @@
...
+        input_fields={
+            "hidden": {
+                "type": bool,
+                "argstr": "-a",
+                "help_string": "show hidden files",
+                "default": False,
+            }
+        },
+        bases=[A],
+        inputs_bases=[A.Inputs],
```
djarecka, line 358: not sure if I understand the idea behind this syntax (i.e. separately providing `bases` and `inputs_bases`

pydra/mark/tests/test_shell.py
```diff
@@ -0,0 +1,467 @@
...
+                    help_string="list of entries returned by ls command",
+                    callable=list_entries,
+                )
+
+
+def test_shell_decorator_misuse(A):
+    with pytest.raises(
+        RuntimeError, match=("`shell_task` should not be provided any other arguments")
+    ):
+        shell_task(A, executable="cp")
```
djarecka, line 467: it might be an answer from one of my previous question. Is this a misuse because you try to overwrite `executable`?
Would it rise an error if `args` provided?
Perhaps the error message could be less general?

djarecka (commented):
I would have to get used to this syntax, but I can see that this might be more readable!
I left some comments under the tests to be sure if I fully understand the idea and syntax. Perhaps a very short description to less trivial tests would be useful

tclose:
closing to see where the #692 discussion goes first

Issue #641, opened 2023-04-21
Issue with composing tasks with output_file_template in a workflow


ghisvail:
I am attempting to build workflows which compose shell command tasks together, some of which providing default output names via `output_file_template`. However, these trigger an exception at runtime if they are not explicitly set when running the workflow, which defeats their purpose in my opinion.

The following minimal example defines a toy backup function wrapping the `cp` command and using a templated output for the target file. The code works fine on the task alone, but fails when wrapped with a worfklow with:

`future: <Task finished name='Task-2' coro=<ConcurrentFuturesWorker.exec_as_coro() done, defined at .../pydra/engine/workers.py:169> exception=Exception("target_file has to be str or bool, but LF('backup', 'target_file') set")>`

I would expect both runs to be successful and yield the same result modulo the temporary directory prefix.

```python
import os
import pathlib
import tempfile

import attrs
import pydra


class BackupFile(pydra.engine.ShellCommandTask):
    """Example task performing a file backup with cp.

    If not target file is provided, it uses a default file suffixed by copy.
    """

    @attrs.define(kw_only=True)
    class InputSpec(pydra.specs.ShellSpec):
        source_file: os.PathLike = attrs.field(
            metadata={"help_string": "source file", "mandatory": True, "argstr": ""}
        )

        target_file: str = attrs.field(
            metadata={
                "help_string": "target file",
                "argstr": "",
                "output_file_template": "{source_file}_copy",
            }
        )

    input_spec = pydra.specs.SpecInfo(name="Input", bases=(InputSpec,))

    executable = "cp"


def backup(name="backup", **kwargs) -> pydra.Workflow:

    wf = pydra.Workflow(input_spec=["source_file", "target_file"], name=name, **kwargs)

    wf.add(
        BackupFile(
            name="backup_file",
            source_file=wf.lzin.source_file,
            target_file=wf.lzin.target_file,
        )
    )

    wf.set_output({"target_file": wf.backup_file.lzout.target_file})

    return wf


if __name__ == "__main__":

    # Execute with standalone backup task.
    with tempfile.TemporaryDirectory() as td:
        source_file = (pathlib.Path(td) / "source1.file")
        source_file.touch()

        task = BackupFile(source_file=source_file)
        result = task()

        print(result.output.target_file)

    # Execute backup task wrapped in a workflow.
    with tempfile.TemporaryDirectory() as td:
        source_file = (pathlib.Path(td) / "source2.file")
        source_file.touch()

[...]

PR #657, opened 2023-05-22
FIX: Discard lazy fields in template type checks

ghisvail:
See issue #641 for context.

## Types of changes

## Summary
This PR discards lazy fields in checks performed in template updates.

I am not very happy with this change, but it's the only way I found so far to bypass this issue without major surgery of the code. Not being able to use `output_file_template` in a workflow is quite a bummer.

Open to other suggestions if any 🙏 

## Triage

```python
from os import PathLike
from attrs import define, field
from pydra.engine import Workflow
from pydra.engine.specs import ShellSpec, SpecInfo
from pydra.engine.task import ShellCommandTask


class BackupFile(ShellCommandTask):
    """Example task performing a file backup with cp.

    If not target file is provided, it uses a default file suffixed by copy.
    """

    @define(kw_only=True)
    class InputSpec(ShellSpec):
        source_file: PathLike = field(
            metadata={"help_string": "source file", "mandatory": True, "argstr": ""}
        )

        target_file: str = field(
            metadata={"help_string": "target file", "argstr": "", "output_file_template": "{source_file}_copy"}
        )

    input_spec = SpecInfo(name="Input", bases=(InputSpec,))

    executable = "cp"


def backup(name="backup", **kwargs) -> Workflow:
    wf = Workflow(input_spec=["source_file", "target_file"], name=name, **kwargs)

    wf.add(
        BackupFile(
            name="backup_file",
            source_file=wf.lzin.source_file,
            target_file=wf.lzin.target_file,
        )
    )

    wf.set_output({"target_file": wf.backup_file.lzout.target_file})

    return wf


if __name__ == "__main__":
    from pathlib import Path
    from tempfile import TemporaryDirectory

    # Execute with standalone backup task.
    with TemporaryDirectory() as td:
        source_file = Path(td) / "source1.file"
        source_file.touch()

        task = BackupFile(source_file=source_file)
        result = task()

        print(result.output.target_file)

    # Execute backup task wrapped in a workflow.
    with TemporaryDirectory() as td:
        source_file = Path(td) / "source2.file"
        source_file.touch()

        task = backup(source_file=source_file)
        result = task()

        print(result.output.target_file)
```

Closes: #641

### DIFF
--- a/pydra/engine/helpers_file.py
+++ b/pydra/engine/helpers_file.py
@@ -628,7 +628,7 @@ def template_update_single(
     based on the value from inputs_dict
     (checking the types of the fields, that have "output_file_template)"
     """
-    from .specs import File, MultiOutputFile, Directory
+    from .specs import File, MultiOutputFile, Directory, LazyField
 
     # if input_dict_st with state specific value is not available,
     # the dictionary will be created from inputs object
@@ -642,7 +642,9 @@ def template_update_single(
                 "has to be a string or Union[str, bool]"
             )
         inp_val_set = inputs_dict_st[field.name]
-        if inp_val_set is not attr.NOTHING and not isinstance(inp_val_set, (str, bool)):
+        if inp_val_set is not attr.NOTHING and not isinstance(
+            inp_val_set, (str, bool, LazyField)
+        ):
             raise Exception(
                 f"{field.name} has to be str or bool, but {inp_val_set} set"
             )


ghisvail:
Without the fix, an exception is raised (`Exception: target_file has to be str or bool, but LF('backup', 'target_file') set`). With the fix, the path to the target file is printed.

djarecka:
if this works for you we can have it temporarily, but I would not close the issue. Sorry, was planning to look at this earlier, but will try to check it this week.

ghisvail:
Fixed by #662

PR #669, opened 2023-06-27
MNT: Drop Python 3.7 support

effigies:
## Types of changes

## Summary
- Drops Python 3.7 support


effigies:
@mgxd Do you know where your Dockerfile is that you used to build https://hub.docker.com/r/mgxd/slurm? We need to update it with a newer Python.

djarecka:
There is also a newer container that was found here: https://github.com/nipype/pydra/discussions/667
but I haven't had time to check why some tests are failing

PR #673, opened 2023-07-10
Fix DaskWorker: `AttributeError: 'tuple' object has no attribute '_run'`

adi611:
## Types of changes

## Summary

This is a fix to DaskWorker which previously failed on some tests with the following error: `AttributeError: 'tuple' object has no attribute '_run'`. (Refer discussion #665).

## Checklist

### DIFF
--- a/.github/workflows/testdask.yml
+++ b/.github/workflows/testdask.yml
@@ -1,0 +1,42 @@
+name: Dask Tests
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8, 3.9, "3.10"]
+      fail-fast: false
+
+    steps:
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+
+      - name: Checkout Pydra repo
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ github.repository }}
+
+      - name: Install pydra with Dask and test dependencies
+        run: |
+          pip install -e ".[test,dask]"
+
+      - name: Run tests
+        run: |
+          pytest -v --dask pydra/engine --cov pydra --cov-config .coveragerc --cov-report xml:cov.xml
+
+      - name: Upload to codecov
+        run: codecov -f cov.xml -F unittests -e GITHUB_WORKFLOW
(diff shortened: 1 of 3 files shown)


djarecka:
@adi611 - thank you for this change, but I understand that dask is still not tested in CI, so we should add it and fix the tests

djarecka:
we still have to understand why this specific test is failing

adi611:
The tests `test_duplicate_input_on_split_wf` and `test_inner_outer_wf_duplicate` in `test_workflow.py` are failing after reaching the timeout for Python versions `3.9` and `3.10`. They also fail for Python version `3.8`, but as it is an older version, I am currently disregarding it. These tests are not Dask tests; instead, they run on the `cf` plugin. It seems to be a GA specific issue since the tests run successfully locally and on Google Colab. Please let me know if I should consider increasing the timeout limit for these tests or if there are any other possible solutions.

This also indicates that all the Dask tests pass for Python versions `3.9` and `3.10`.

djarecka:
for some reason i'm having problems with seeing reports from the tests right now..:( What do you mean this tests are not dask tests, I'm guessing that since it only happens with the "dask" run, some tests are stuck when running the dask plugin, am I right?

Do they have a reasonable running time on google collab?

adi611:
Sorry for the confusion. The tests `test_duplicate_input_on_split_wf` and `test_inner_outer_wf_duplicate` in `test_workflow.py` are the only failing tests for Python `3.9` and `3.10`. These tests run on `cf` only, using `plugin="cf"`. The tests which run on both `dask` and `cf` have `plugin=plugin_dask_opt` which is not the case here. So what I meant was the reason these tests are failing may not be a dask related one.

Moreover these tests are passing on Google Colab, as can be seen here: [github gist for python 3.10](https://colab.research.google.com/gist/adi611/f0c6689a1e538ef5587b5e04a63d5178/untitled16.ipynb), [github gist for python 3.9](https://colab.research.google.com/gist/adi611/0695067a7ed043b4ca22fcc8bd1008b2/untitled17.ipynb).

djarecka:
@adi611 - it looks like the change in timeout didn't help, and I actually check one of the GA report (for 3.9), and I can see that there is an issue with another test, with `test_wf_3nd_st_1` that doesn't have timeout, but it fails (even with rerun). It also fails on my laptop... 

Are you really able to run without any issues this test locally?

adi611:
@djarecka - The test `test_wf_3nd_st_1` was not failing before increasing the timeout. In fact, for my fork of the repo the tests actually pass for Python `3.9`, and while they do fail for Python `3.10` , it is only for the tests `test_duplicate_input_on_split_wf` and `test_inner_outer_wf_duplicate`. It can be seen [here](https://github.com/adi611/pydra/actions/runs/5577146618/job/15657744300).

Even though the results appear to be inconsistent across platforms, running the test files individually does not result to any error in any platform. I have created a new GA workflow to verify the same. This will be helpful in understanding the cause of the problem.

djarecka:
ok, thank you. I've approved the workflows, so should run now

ghisvail:
Superseded by #686

PR #693, opened 2023-08-31
Omit false output templates

tclose:
## Types of changes

## Summary

Optional output fields with `output_file_template`  metadata (i.e. of type=`ty.Union[Path, bool]`) were getting printed to the command line regardless of whether a value for the field was set to False.

## Checklist

### DIFF
--- a/pydra/engine/helpers_file.py
+++ b/pydra/engine/helpers_file.py
@@ -120,6 +120,7 @@ def template_update(inputs, output_dir, state_ind=None, map_copyfiles=None):
         field
         for field in attr_fields(inputs)
         if field.metadata.get("output_file_template")
+        and getattr(inputs, field.name) is not False
         and all(
             getattr(inputs, required_field) is not attr.NOTHING
             for required_field in field.metadata.get("requires", ())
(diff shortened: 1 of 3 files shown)


pydra/engine/helpers_file.py
```diff
@@ -120,6 +120,7 @@ def template_update(inputs, output_dir, state_ind=None, map_copyfiles=None):
         field
         for field in attr_fields(inputs)
         if field.metadata.get("output_file_template")
+        and getattr(inputs, field.name) is not False
```
effigies, line 123: Do you want it to be specifically `False`, or is falsey better? Consider `Optional[Path]`, where the default is `None` instead. It's unclear when a value would be falsey and useful.

```suggestion
        and getattr(inputs, field.name)
```

I'm not 100% sure what would be least surprising here.

effigies (approved):
Test looks good. Q about logic.

pydra/engine/helpers_file.py
tclose: The only falsey value would be the empty string, given the field is typed ty.Union[str, bool]. However, in the proposed syntax changes they would change to ty.Union[Path, bool] so the only falsey value would be False in any case

tclose: Ah, just reread your comment. For Optional[Path] fields, we would want the None values to be omitted wouldn't we? So falsey would be better

tclose: Double ah, I remembered why it is explicitly not False instead of falsey, we want to use the template when the value is attrs.NOTHING. Have reverted the suggested change

Issue #681, opened 2023-08-02
MNT: Check that the checksum of a task doesn't change during execution


tclose:
If the checksum of a task changes during execution then Pydra won't know where to look for the results and bugs out with a cryptic message (NoneType doesn't have attribute ...). We should at least cache the checksum and detect when it has changed and provide a more descriptive error message. Ideally, we would store checksums for each of the input fields to help narrow down the issue in the error message.

For example, I had a task where the output directory was specified as type `Directory` instead of `Path` and therefore its hash changed after the output data was written to it. We should be able to catch this case and prompt the user to consider changing it to a path.

tclose:
This has been implemented

PR #698, opened 2023-09-11
Hash change guards

tclose:
## Types of changes

## Summary

Addresses #681 issue

* Detects changes in the hashes of task/workflow inputs and raises informative error messages identifying which fields have changed
* Reinstates behaviour of "blocked tasks" unittest, so it once again triggers "graph is not empty" exception
* On "graph not empty" exception, check Workflow.inputs._graph_checksums to determine which task's checksums have changed and raise informative message

## Checklist

### DIFF
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -281,9 +281,9 @@ class TaskBase:
 
         """
         if is_workflow(self) and self.inputs._graph_checksums is attr.NOTHING:
-            self.inputs._graph_checksums = [
-                (nd.name, nd.checksum) for nd in self.graph_sorted
-            ]
+            self.inputs._graph_checksums = {
+                nd.name: nd.checksum for nd in self.graph_sorted
+            }
 
         if state_index is not None:
             inputs_copy = copy(self.inputs)
@@ -1142,9 +1142,9 @@ class Workflow(TaskBase):
         """
         # if checksum is called before run the _graph_checksums is not ready
         if is_workflow(self) and self.inputs._graph_checksums is attr.NOTHING:
-            self.inputs._graph_checksums = [
-                (nd.name, nd.checksum) for nd in self.graph_sorted
-            ]
+            self.inputs._graph_checksums = {
+                nd.name: nd.checksum for nd in self.graph_sorted
+            }
 
         input_hash = self.inputs.hash
         if not self.state:
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -821,8 +821,8 @@ class TaskBase:
 
         Returns
         -------
-        result :
-
+        result : Result
+            the result of the task
         """
         # TODO: check if result is available in load_result and
         # return a future if not
@@ -894,27 +894,33 @@ class TaskBase:
         details = ""
         for changed in hash_changes:
             field = getattr(attr.fields(type(self.inputs)), changed)
+            val = getattr(self.inputs, changed)
+            field_type = type(val)
             if issubclass(field.type, FileSet):
                 details += (
-                    f"- '{changed}' field is of file-type {field.type}. If it "
-                    "is intended to contain output data then the type of the field in "
-                    "the interface class should be changed to `pathlib.Path`. Otherwise, "
-                    "if the field is intended to be an input field but it gets altered by "
-                    "the task in some way, then the 'copyfile' flag should be set to "
-                    "'copy' in the field metadata of the task interface class so copies of "
-                    "the files/directories in it are passed to the task instead\n"
+                    f"- {changed}: value passed to the {field.type} field is of type "
+                    f"{field_type} ('{val}'). If it is intended to contain output data "
+                    "then the type of the field in the interface class should be changed "
+                    "to `pathlib.Path`. Otherwise, if the field is intended to be an "
+                    "input field but it gets altered by the task in some way, then the "
+                    "'copyfile' flag should be set to 'copy' in the field metadata of "
+                    "the task interface class so copies of the files/directories in it "
+                    "are passed to the task instead.\n"
                 )
             else:
                 details += (
-                    f"- the {field.type} object passed to '{changed}' field appears to "
-                    f"have an unstable hash. The {field.type}.__bytes_repr__() method "
-                    "can be implemented to provide stable hashes for this type. "
-                    "See pydra/utils/hash.py for examples.\n"
+                    f"- {changed}: the {field_type} object passed to the {field.type}"
+                    f"field appears to have an unstable hash. This could be due to "
+                    "a stochastic/non-thread-safe attribute(s) of the object\n\n"
+                    f"The {field.type}.__bytes_repr__() method can be implemented to "
+                    "bespoke hashing methods based only on the stable attributes for "
+                    f"the `{field_type.__module__}.{field_type.__name__}` type. "
+                    f"See pydra/utils/hash.py for examples. Value: {val}\n"
                 )
         if hash_changes:
             raise RuntimeError(
                 f"Input field hashes have changed during the execution of the "
-                f"'{self.name}' {type(self).__name__}.\n{details}"
+                f"'{self.name}' {type(self).__name__}.\n\n{details}"
             )
         logger.debug(
             "Input values and hashes for '%s' %s node:\n%s\n%s",
--- a/pydra/engine/core.py
+++ b/pydra/engine/core.py
@@ -558,15 +558,9 @@ class TaskBase:
                 os.chdir(cwd)
         # Check for any changes to the input hashes that have occurred during the execution
         # of the task
-        hash_changes = self.inputs.hash_changes()
-        if hash_changes:
-            raise RuntimeError(
-                f"Hashes have changed for {hash_changes} input fields during the "
-                f"execution of {self}. Please check all output files/directories are "
-                "typed with `pathlib.Path` instead of `fileformats` classes"
-            )
-        self.hooks.post_run(self, result)
 
+        self.hooks.post_run(self, result)
+        self._check_for_hash_changes()
         return result
 
     def _collect_outputs(self, output_dir):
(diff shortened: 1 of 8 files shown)


tclose:
I believe this is good for review now. The new SLURM test keeps on getting stuck somewhere but I assume that it isn't related

tclose:
NB: I'm not sure why the coverage is down. It seems to be an issue with the dask submitter code

pydra/engine/tests/test_specs.py
```diff
@@ -27,7 +27,7 @@
 
 def test_basespec():
     spec = BaseSpec()
-    assert spec.hash == "06fe829a5dca34cc5f0710b454c24808"
+    assert spec.hash == "0b1d98df22ecd1733562711c205abca2"
 
 
 def test_runtime():
```
ghisvail, line 33: I am a bit confused why the hash values should change. Aren't the test values in the specifications still the same as before?

ghisvail (approved):
LGTM with minor comments / questions which might be due to my own ignorance or overlook.

pydra/engine/submitter.py
```diff
@@ -168,24 +168,52 @@ async def expand_workflow(self, wf, rerun=False):
...
+                            "results predecessors:\n\n"
                         )
+                        # Get blocked tasks and the predecessors they are waiting on
+                        outstanding = {
+                            t: [
+                                p for p in graph_copy.predecessors[t.name] if not p.done
+                            ]
+                            for t in graph_copy.sorted_nodes
+                        }
+                        graph_checksums = dict(wf.inputs._graph_checksums)
```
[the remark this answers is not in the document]
tclose: I believe the main reason `_graph_checksums` has been made a "protected member" is to separate it from the user namespace accessed via `_inputs`, i.e. I think PyCharm is being a little overzealous on this issue

pydra/engine/tests/test_specs.py
```diff
@@ -27,7 +27,7 @@
 
 def test_basespec():
     spec = BaseSpec()
-    assert spec.hash == "06fe829a5dca34cc5f0710b454c24808"
+    assert spec.hash == "0b1d98df22ecd1733562711c205abca2"
```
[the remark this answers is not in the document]
tclose: Yes, it doesn't quite feel right to hard code this. However, it does flag whether the hash is stable across versions (or isn't in this case). Ideally it would be so that caches wouldn't be invalidated by new Pydra versions, but in this case, I'm not sure whether there is an easy way around this

pydra/engine/tests/test_specs.py
tclose: The test values are the same, just the type of the outer container and the index that `_graph_checksums` is inserted changes

djarecka:
indeed looks like the coverage has dropped, what is weird, since for some time we didn't test `Dask` at all...

pydra/engine/tests/test_specs.py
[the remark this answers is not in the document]
djarecka: I believe I added these kind of tests to not change the hashing way without noticing

djarecka (approved):
Thank you! I think it is really a great improvement to the debugging options for Pydra

tclose:
> indeed looks like the coverage has dropped, what is weird, since for some time we didn't test `Dask` at all...

I don't trust codecov in this instance (and others). The dask test is passing so it couldn't be missing the Dask submitter, which is where the bulk of the coverage drop comes from (otherwise it would be a coverage increase I believe)

pydra/engine/core.py
```diff
@@ -480,8 +482,8 @@ def _modify_inputs(self):
         modified_inputs = template_update(
             self.inputs, self.output_dir, map_copyfiles=map_copyfiles
         )
-        if modified_inputs:
-            self.inputs = attr.evolve(self.inputs, **modified_inputs)
+        for name, value in modified_inputs.items():
+            setattr(self.inputs, name, value)
```
effigies, line 486: This is okay for now, but it feels messy. I think it would be good to find some way to keep using attrs in the usual way, and keep track of the changes inside `Task` instead of `Task.inputs._hashes`.

pydra/engine/tests/test_submitter.py
```diff
@@ -1,10 +1,13 @@
 from dateutil import parser
 import re
```
effigies, line 2:
```suggestion
import random
import re
```

pydra/engine/tests/test_submitter.py
```diff
@@ -573,42 +576,93 @@ def test_sge_no_limit_maxthreads(tmpdir):
...
-    with Submitter("serial") as sub:
-        sub(wf)
+def test_hash_changes_in_task_inputs_unstable(tmp_path):
+    @attrs.define
+    class Unstable:
+        value: int  # type: ignore
 
+        def __bytes_repr__(self, cache) -> ty.Iterator[bytes]:
+            """Bytes repr based on time-stamp -> inherently unstable"""
+            yield struct.pack("!I", int(time.time()))
```
effigies, line 597:
```suggestion
        def __bytes_repr__(self, cache) -> ty.Iterator[bytes]:
            """Random 128-bit bytestring"""
            yield random.randbytes(16)
```

effigies (commented):
I guess I could have made suggestions...

pydra/engine/core.py
```diff
@@ -280,7 +280,9 @@ def checksum_states(self, state_index=None):
 
         """
         if is_workflow(self) and self.inputs._graph_checksums is attr.NOTHING:
-            self.inputs._graph_checksums = [nd.checksum for nd in self.graph_sorted]
+            self.inputs._graph_checksums = [
+                (nd.name, nd.checksum) for nd in self.graph_sorted
+            ]
```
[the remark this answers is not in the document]
tclose: It doesn't need to be a dict as we never look up the checksum by name, the names are just used to provide more detailed debugging messages

pydra/engine/core.py
tclose: In the syntax changes I am proposing, the input spec would be subsumed into the task class in any case. So I might hold off and see what people think of them first

pydra/engine/core.py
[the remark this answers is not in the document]
tclose: Actually, I take that back. I do look up the checksum by name but just turn it into a dictionary at that point.

pydra/utils/hash.py
```diff
@@ -58,12 +58,12 @@ class UnhashableError(ValueError):
     """Error for objects that cannot be hashed"""
 
 
-def hash_function(obj):
+def hash_function(obj, cache=None):
```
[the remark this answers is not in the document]
tclose: I'm not sure many people will call this function directly, will they?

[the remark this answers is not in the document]
tclose: I went with `cache: ty.Optional[Cache]` for Py 3.8 & 3.9 support

pydra/engine/tests/test_submitter.py
```diff
@@ -573,42 +576,93 @@ def test_sge_no_limit_maxthreads(tmpdir):
...
 
-class A:
-    def __init__(self, a):
-        self.a = a
+    @mark.task
+    def unstable_input(unstable: Unstable) -> int:
+        time.sleep(1)  # Ensure the timestamp changes during the task run
```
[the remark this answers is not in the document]
tclose: ok, done

pydra/engine/core.py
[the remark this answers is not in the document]
tclose: I have changed it to a dict in the latest commit

pydra/engine/core.py
djarecka: and why this is changing?  is this  related to the issue with pydra-ml?

pydra/engine/core.py
```diff
@@ -281,13 +281,15 @@ def checksum_states(self, state_index=None):
...
         """
         if is_workflow(self) and self.inputs._graph_checksums is attr.NOTHING:
-            self.inputs._graph_checksums = [nd.checksum for nd in self.graph_sorted]
+            self.inputs._graph_checksums = {
+                nd.name: nd.checksum for nd in self.graph_sorted
+            }
 
         if state_index is not None:
-            inputs_copy = deepcopy(self.inputs)
+            inputs_copy = copy(self.inputs)
```
djarecka, line 289: are we sure, that removing deepcopy here will be always ok?

pydra/engine/core.py
tclose: I can't actually remember why that was changed to be like that, but in any it has been changed further now with the changes to fix the pydra-ml-raised issue

pydra/engine/core.py
tclose: I'm not 100% but pretty confident. It looks like the copy is only done so that the state-array values can be resolved to single values on L294

```
        if state_index is not None:
            inputs_copy = copy(self.inputs)
            for key, ind in self.state.inputs_ind[state_index].items():
                val = self._extract_input_el(
                    inputs=self.inputs, inp_nm=key.split(".")[1], ind=ind
                )
                setattr(inputs_copy, key.split(".")[1], val)
```

tclose: https://github.com/nipype/pydra/blob/4e1d4a89f40857453ac8cb86f1061426a8ebcb3e/pydra/engine/core.py#L288-L294

djarecka: yes, I think that was indeed mostly needed for the tasks with states

pydra/engine/core.py
```diff
@@ -548,13 +572,14 @@ def _run(self, rerun=False, environment=None, **kwargs):
...
                 # removing the additional file with the checksum
                 (self.cache_dir / f"{self.uid}_info.json").unlink()
-                # # function etc. shouldn't change anyway, so removing
-                orig_inputs = {
-                    k: v for k, v in orig_inputs.items() if not k.startswith("_")
-                }
-                self.inputs = attr.evolve(self.inputs, **orig_inputs)
+                # Restore original values to inputs
+                for field_name, field_value in orig_inputs.items():
+                    setattr(self.inputs, field_name, field_value)
```
satra, line 577: why not call `attr.evolve` directly here if the startswith is not relevant in construction?

tclose: No reason, I just didn't think about it after I had changed a few things back and forth. Updated it now

pydra/engine/core.py
```diff
@@ -548,13 +572,13 @@ def _run(self, rerun=False, environment=None, **kwargs):
                 save(output_dir, result=result, task=self)
                 # removing the additional file with the checksum
                 (self.cache_dir / f"{self.uid}_info.json").unlink()
-                # # function etc. shouldn't change anyway, so removing
-                orig_inputs = {
-                    k: v for k, v in orig_inputs.items() if not k.startswith("_")
```
satra, line 553: @tclose - do we no longer have this scenario of attributes starting with `_`?

tclose:
@satra I remembered/realised why I went with setattr instead of evolve, because the `_hashes` attribute I'm using to store the original hashes before the run gets wiped out

pydra/engine/core.py
tclose: I remembered/realised why I went with setattr instead of evolve, because the `_hashes` attribute I'm using to store the original hashes before the run gets wiped out (did a hard reset to previous commit to revert)

tclose:
@djarecka I believe this is ready to merge now

djarecka:
I'm good with merging, thank you @tclose !

Issue #683, opened 2023-08-03
MNT: Reimplement file-hash caching in new hashing framework


tclose:
A casualty of the new hashing framework introduced by [pydra#662](https://github.com/nipype/pydra/pull/662) was the removal of file-hash caching (only calculating the file-hash once per task. For large files this could be a significant performance regression so it would be good to work out how to add it back in.

### Suggestions
1. Just cache the checksum in the task object (if we place guards on it changing post-execution, see [pydra:#681](https://github.com/nipype/pydra/issues/681), this might be sufficient)
2. Return file mtime as part of __bytes_repr__ and use this to create a local cache. This mapping could be potentially stored on disk for persistence between runs

tclose:
@effigies

My idea for this feature is for the `bytes_repr` function overload of file-classes (and potentially any other type that you want to cache the hashes for) to yield a "time-stamp" object consisting of a key (file path) and modification time as the first item in the generator.

The calling `hash_single` function can check for these key/mtime pairs in a "hashes cache" dict loaded from the cache directory, and return the cached hash if present. Otherwise it proceeds through the remaining byte chunks, calculates the hash and then saves it into the hashes cache using the key/mtime

effigies:
Yes, that makes sense to me. I would probably genericly type the yield type as `Union[CacheKey, bytes]`, where CacheKey is a `tuple[Hashable, ...]` newtype. You could imagine some type wanting something besides `(Path, int)`, e.g., an S3 path with a version key.

PR #700, opened 2023-09-11
Caching of file-set hashes by local path and mtimes

tclose:
## Types of changes

## Summary

Will address #683 by

- [x] `bytes_repr()` overloads for FileSets to yield a "time-stamp" object consisting of a key (file path) and modification time as the first item in the generator.
- [x] `hash_single()` checks for these key/mtime pairs in a global "local hashes cache" dict
    - returns cached hash, if present
    - otherwise it proceeds through the remaining byte chunks, and calculates the hash
- [x] calculated hashes are saved into/loaded from the cache directory using a hash of the key/mtime 

## Checklist

## Notes
I'm pretty happy with how this turned out with the exception of a few of wrinkles. Any suggestions would be most welcome
- There isn't a clean way to specify the location of the persistent hash cache in the top-level code or put it in the `cache_location` path (my original plan) given that `checksum` and `hash` are object properties instead of methods
    - I have therefore just dumped it in `~/.pydra-hash-cache`
    - not ideal to have to (effectively) hard-code this and having it user dependent
    - Having it in the user directory does mean that if the same cache directory is accessed from different machines (e.g. shared network drive) then local paths won't clash (although chance of mtimes being the same would be vanishingly small)
- The keys of the caches themselves are hashed and stored as files in the persistent cache directory, and will therefore never be cleaned up.
    - This could be ok as they are very small files so if it grows as time goes on it probably won't have much impact
    - This is how pydra caches work in general
- The resolution of mtime differs depending on OS, with Ubuntu and Windows having quite low resolution, i.e. order of seconds
    - I have had to put a sleep call inside the hash test to ensure the mtimes are different
    - Could cause issues if you were updating the contents of files within a loop and spinning off different workflows (not sure whether this would ever happen in practice)
    - No easy way to disable the mtime caching behaviour if this does become a problem

### DIFF
--- a/pydra/engine/tests/test_node_task.py
+++ b/pydra/engine/tests/test_node_task.py
@@ -1563,3 +1563,39 @@ def test_task_state_cachelocations_updated(plugin, tmp_path):
     # both workflows should be run
     assert all([dir.exists() for dir in nn.output_dir])
     assert all([dir.exists() for dir in nn2.output_dir])
+
+
+def test_task_files_cachelocations(plugin_dask_opt, tmp_path):
+    """
+    Two identical tasks with provided cache_dir that use file as an input;
+    the second task has cache_locations and should not recompute the results
+    """
+    cache_dir = tmp_path / "test_task_nostate"
+    cache_dir.mkdir()
+    cache_dir2 = tmp_path / "test_task_nostate2"
+    cache_dir2.mkdir()
+    input_dir = tmp_path / "input"
+    input_dir.mkdir()
+
+    input1 = input_dir / "input1.txt"
+    input1.write_text("test")
+    input2 = input_dir / "input2.txt"
+    input2.write_text("test")
+
+    nn = fun_file(name="NA", filename=input1, cache_dir=cache_dir)
+    with Submitter(plugin=plugin_dask_opt) as sub:
+        sub(nn)
+
+    nn2 = fun_file(
+        name="NA", filename=input2, cache_dir=cache_dir2, cache_locations=cache_dir
+    )
+    with Submitter(plugin=plugin_dask_opt) as sub:
+        sub(nn2)
+
+    # checking the results
+    results2 = nn2.result()
+    assert results2.output.out == "test"
+
+    # checking if the second task didn't run the interface again
+    assert nn.output_dir.exists()
+    assert not nn2.output_dir.exists()
--- a/pydra/engine/tests/test_node_task.py
+++ b/pydra/engine/tests/test_node_task.py
@@ -1,10 +1,15 @@
 import os
 import shutil
 import attr
+import typing as ty
 import numpy as np
+import time
 from unittest import mock
+from pathlib import Path
 import pytest
 import time
+from fileformats.generic import File
+import pydra.mark
 
 from .utils import (
     fun_addtwo,
@@ -1599,3 +1604,62 @@ def test_task_files_cachelocations(plugin_dask_opt, tmp_path):
     # checking if the second task didn't run the interface again
     assert nn.output_dir.exists()
     assert not nn2.output_dir.exists()
+
+
+class OverriddenContentsFile(File):
+    """A class for testing purposes, to that enables you to override the contents
+    of the file to allow you to check whether the persistent cache is used."""
+
+    def __init__(
+        self,
+        fspaths: ty.Iterator[Path],
+        contents: ty.Optional[bytes] = None,
+        metadata: ty.Dict[str, ty.Any] = None,
+    ):
+        super().__init__(fspaths, metadata=metadata)
+        self._contents = contents
+
+    def byte_chunks(self, **kwargs) -> ty.Generator[ty.Tuple[str, bytes], None, None]:
+        if self._contents is not None:
+            yield (str(self.fspath), iter([self._contents]))
+        else:
+            yield from super().byte_chunks(**kwargs)
+
+    @property
+    def contents(self):
+        if self._contents is not None:
+            return self._contents
+        return super().contents
+
+
+def test_task_files_persistentcache(tmp_path):
+    """
+    Two identical tasks with provided cache_dir that use file as an input;
+    the second task has cache_locations and should not recompute the results
+    """
+    test_file_path = tmp_path / "test_file.txt"
+    test_file_path.write_bytes(b"foo")
+    cache_dir = tmp_path / "cache-dir"
+    cache_dir.mkdir()
+    test_file = OverriddenContentsFile(test_file_path)
+
+    @pydra.mark.task
+    def read_contents(x: OverriddenContentsFile) -> bytes:
+        return x.contents
+
+    assert (
+        read_contents(x=test_file, cache_dir=cache_dir)(plugin="serial").output.out
+        == b"foo"
+    )
+    test_file._contents = b"bar"
+    # should return result from the first run using the persistent cache
+    assert (
+        read_contents(x=test_file, cache_dir=cache_dir)(plugin="serial").output.out
+        == b"foo"
+    )
+    time.sleep(2)  # Windows has a 2-second resolution for mtime
+    test_file_path.touch()  # update the mtime to invalidate the persistent cache value
+    assert (
+        read_contents(x=test_file, cache_dir=cache_dir)(plugin="serial").output.out
+        == b"bar"
+    )  # returns the overridden value
--- a/pydra/engine/tests/test_node_task.py
+++ b/pydra/engine/tests/test_node_task.py
@@ -322,11 +322,10 @@ def test_task_init_7(tmp_path):
     output_dir1 = nn1.output_dir
 
     # changing the content of the file
+    time.sleep(2)  # need the mtime to be different
     file2 = tmp_path / "file2.txt"
-    with mock.patch("time.time") as t:
-        t.return_value = time.time() + 10  # mock mtime to ensure different
-        with open(file2, "w") as f:
-            f.write("from pydra")
+    with open(file2, "w") as f:
+        f.write("from pydra")
 
     nn2 = fun_file_list(name="NA", filename_list=[file1, file2])
     output_dir2 = nn2.output_dir
--- a/pydra/engine/tests/test_node_task.py
+++ b/pydra/engine/tests/test_node_task.py
@@ -321,10 +321,13 @@ def test_task_init_7(tmp_path):
     output_dir1 = nn1.output_dir
 
     # changing the content of the file
-    time.sleep(2)  # need the mtime to be different
     file2 = tmp_path / "file2.txt"
-    with open(file2, "w") as f:
-        f.write("from pydra")
+    with mock.patch("time.time") as t:
+        t.return_value = (
+            time.time() + 10
+        )  # mock mtime writing to ensure it is different
+        with open(file2, "w") as f:
+            f.write("from pydra")
 
     nn2 = fun_file_list(name="NA", filename_list=[file1, file2])
     output_dir2 = nn2.output_dir
--- a/pydra/engine/tests/test_node_task.py
+++ b/pydra/engine/tests/test_node_task.py
@@ -323,9 +323,7 @@ def test_task_init_7(tmp_path):
     # changing the content of the file
     file2 = tmp_path / "file2.txt"
     with mock.patch("time.time") as t:
-        t.return_value = (
-            time.time() + 10
-        )  # mock mtime writing to ensure it is different
+        t.return_value = time.time() + 10  # mock mtime to ensure different
         with open(file2, "w") as f:
             f.write("from pydra")
 
--- a/pydra/engine/tests/test_node_task.py
+++ b/pydra/engine/tests/test_node_task.py
@@ -2,6 +2,7 @@ import os
 import shutil
 import attr
 import numpy as np
+from unittest import mock
 import pytest
 import time
 
--- a/pydra/engine/tests/test_node_task.py
+++ b/pydra/engine/tests/test_node_task.py
@@ -3,6 +3,7 @@ import shutil
 import attr
 import numpy as np
 import pytest
+import time
 
 from .utils import (
     fun_addtwo,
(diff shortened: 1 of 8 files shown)


ghisvail:
I am not confident I understand enough of the previous and proposed caching methods to have an opinion and assess this PR.

AFAIK, the previous caching mechanism generated cache folders for each task (including the workflow) in a common cache location which was set to a temporary folder, unless overridden with the `cache_dir` argument.

Each task's cache folder was composed of its name and a hash value, the latter being computed from the task's input field values. The cache would be re-used, if the name and hash values did not change, and the same cache folder was specified as `cache_dir`. Otherwise, it would recompute everything in a new cache location somewhere in temp.

Could you confirm whether my summary is accurate and whether this PR is proposing to change the fundamentals of this mechanism?

tclose:
> I am not confident I understand enough of the previous and proposed caching methods to have an opinion and assess this PR.

No worries, I just put you all down as reviewers, so you were notified. Don't feel like you need to contribute if the area isn't familiar

> AFAIK, the previous caching mechanism generated cache folders for each task (including the workflow) in a common cache location which was set to a temporary folder, unless overridden with the `cache_dir` argument.
> 
> Each task's cache folder was composed of its name and a hash value, the latter being computed from the task's input field values. The cache would be re-used, if the name and hash values did not change, and the same cache folder was specified as `cache_dir`. Otherwise, it would recompute everything in a new cache location somewhere in temp.
> 
> Could you confirm whether my summary is accurate and whether this PR is proposing to change the fundamentals of this mechanism?

Yes, your understanding is correct. Tis is how the hashing works, both currently and in this PR. This PR seeks to improve (or more accurately, restore) performance by caching the hashes of file/directory types themselves, so files/directories don't need to be rehashed (which can be an expensive operation) each time the checksum is accessed.

It does this by hashing the path and mtime of the file/directory (not to be confused with the hashing of the file/directory contents), and using it as a key to look up a "hash cache" (not to be confused with the execution cache) that contains previously computed file/directory hashes.

effigies:
> * I have had to put a sleep call inside the hash test to ensure the mtimes are different

Better to mock the call than to sleep. Here's an example where I've done that with `time.time()`:
https://github.com/nipy/nibabel/blob/5f37398a2f8211c175b798eb42298b963c693ae0/nibabel/tests/test_openers.py#L457-L464

tclose:
> > * I have had to put a sleep call inside the hash test to ensure the mtimes are different
> 
> Better to mock the call than to sleep. Here's an example where I've done that with `time.time()`: https://github.com/nipy/nibabel/blob/5f37398a2f8211c175b798eb42298b963c693ae0/nibabel/tests/test_openers.py#L457-L464

Interesting, I assumed that the value for the mtime was controlled by the file-system not Python.

pydra/utils/hash.py
```diff
@@ -51,19 +53,109 @@
...
+    _hashes: ty.Dict[CacheKey, Hash] = attrs.field(factory=dict)
+
+    @location.validator
+    def location_validator(self, _, location):
+        if not os.path.isdir(location):
+            raise ValueError(
+                f"Persistent cache location '{location}' is not a directory"
+            )
+
+    def get_hash(self, key: CacheKey, calculate_hash: ty.Callable) -> Hash:
```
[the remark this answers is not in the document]
tclose: ok, sounds good

pydra/utils/hash.py
```diff
@@ -51,19 +53,109 @@
...
+    if persistent_cache is None:
+        # FIXME: Ideally the default location would be inside one of the cache_locations
+        # but can't think of a clean way to pass the cache_locations down to this part
+        # of the code, so just dumping in the home directory instead
+        persistent_cache = (Path("~") / ".pydra-hash-cache").expanduser()
+        try:
+            if not persistent_cache.exists():
+                persistent_cache.mkdir()
+        except Exception:
+            persistent_cache = None
```
ghisvail, line 154: I believe we should be careful where to place stuff in the home directory.

We could follow the old "wild west" convention of placing files in a single `.pydra` folder at the root of `$HOME`, but this is generally considered sloppy, and sooner or later someone will complain and ask whether we can support established standards for their platform of choice.

Luckily, there is a [package](https://pypi.org/project/platformdirs/) implementing this already. I guess we would be concerned by this [section](https://platformdirs.readthedocs.io/en/latest/api.html#cache-directory).

ghisvail (changes requested):
Some suggestions and a question mark regarding the cache path in the home folder. Maybe also worth having a look to mtime mocking instead of sleep steps as suggested by Chris.

Looks solid otherwise 👍

djarecka:
@tclose - you can try with new slurm testing workflow, should work now!

pydra/utils/hash.py
djarecka: - perhaps we could just use a specific environmental variable if specified by the user?
- Since this is meant to be used within specific run, should we have option to clean the files that are e.g. a week old?

djarecka (approved):
lgtm! just left comments regarding the location of the cache directory

pydra/utils/hash.py
```diff
@@ -51,19 +53,109 @@
...
+            if key_path.exists():
+                return Hash(key_path.read_bytes())
+            hsh = calculate_hash()
+            key_path.write_bytes(hsh)
+        return Hash(hsh)
+
+
+def persistent_cache_converter(
+    path: ty.Union[Path, str, PersistentCache, None]
+) -> PersistentCache:
```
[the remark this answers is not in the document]
tclose: I believe this is a strange limitation of attrs, that converters can't be classmethods. I presume it is in order to preserve performance

pydra/utils/hash.py
tclose: That package looks good @ghisvail, although I not sure this is worth adding another dependency for, what are everyone's thoughts?

In either case, an env var is a good idea @djarecka. Do we have a naming convention for them?

> Since this is meant to be used within specific run, should we have option to clean the files that are e.g. a week old?

I don't see this needing to be constrained to a specific run, but probably no harm to run a clean-up of week/month-old files after a workflow completes

pydra/utils/hash.py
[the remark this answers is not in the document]
tclose: Actually, I had this wrong. It is just a limitation of mypy. I have added it as a classmethod and just ignored the mypy warning

tclose:
> > * I have had to put a sleep call inside the hash test to ensure the mtimes are different
> 
> Better to mock the call than to sleep. Here's an example where I've done that with `time.time()`: https://github.com/nipy/nibabel/blob/5f37398a2f8211c175b798eb42298b963c693ae0/nibabel/tests/test_openers.py#L457-L464

@effigies I'm not sure this works on Windows (it seems to work on Ubuntu)

pydra/utils/hash.py
ghisvail: > although I not sure this is worth adding another dependency for

Every piece of software which cache stuff to home end up writing (usually badly) some subset of what's covered in `platformdirs`. It's not worth adding if we are confident to implement it *right*. If not, the package is dependency free and [well tested](https://github.com/platformdirs/platformdirs/actions/workflows/check.yml), so the risks and additional bloat to Pydra would be minimal.

> we could just use a specific environmental variable if specified by the user

We could indeed (say `PYDRA_CACHE_DIR`), but the suitable fallback value is platform dependent. That's where the value of `platformdirs` comes in.

> I don't see this needing to be constrained to a specific run, but probably no harm to run a clean-up of week/month-old files after a workflow completes

I am trying to picture a use case for this: if I run workflow A and months later workflow B, should I expect Pydra to clean cached files up from workflow A (and any other which are old enough w.r.t. workflow B)?

tclose: > > although I not sure this is worth adding another dependency for
> 
> Every piece of software which cache stuff to home end up writing (usually badly) some subset of what's covered in `platformdirs`. It's not worth adding if we are confident to implement it _right_. If not, the package is dependency free and [well tested](https://github.com/platformdirs/platformdirs/actions/workflows/check.yml), so the risks and additional bloat to Pydra would be minimal.

Sounds ok to me, I looked into pulling out just the bits we need from that package, and for Mac+Linux it was trivial but Windows is pretty complicated so decided against it.

> > we could just use a specific environmental variable if specified by the user
> 
> We could indeed (say `PYDRA_CACHE_DIR`), but the suitable fallback value is platform dependent. That's where the value of `platformdirs` comes in.

👍

> 
> > I don't see this needing to be constrained to a specific run, but probably no harm to run a clean-up of week/month-old files after a workflow completes
> 
> I am trying to picture a use case for this: if I run workflow A and months later workflow B, should I expect Pydra to clean cached files up from workflow A (and any other which are old enough w.r.t. workflow B)?

Yes, if they are not touched by Workflow B and they are over a month old they get cleaned up.

tclose: It is still showing up as "1 change requested" from you, I think it might have been the method name change (which I implemented in the last round of changes). Is that right?

tclose: @ghisvail I realised that *platformdirs* is already dependency of *black* and *virtualenv* so have used it to locate the hash cache as you suggested.

tclose:
@effigies, I think that I have addressed the outstanding issues with this PR so it just needs your review. However, I had to revert the mtime mocking as it doesn't appear to work for Windows (see https://github.com/nipype/pydra/actions/runs/6307685788/job/17124695852), unless you have some ideas on how to do it.

djarecka:
@tclose - the recent commit to this PR comes from some rebasing?

tclose:
> @tclose - the recent commit to this PR comes from some rebasing?

I rebased it on top of master after the environment PR was merged and fixed up a few things

djarecka:
ok, trying to figure out if I should review it again, since I already accepted

pydra/utils/hash.py
```diff
@@ -52,19 +54,144 @@
...
+    _hashes: ty.Dict[int, Hash] = attrs.field(factory=dict)
+
+    def __getitem__(self, object_id: int) -> Hash:
+        return self._hashes[object_id]
+
+    def __setitem__(self, object_id: int, hsh: Hash):
+        self._hashes[object_id] = hsh
+
+    def __contains__(self, object_id):
+        return object_id in self._hashes
```
effigies, line 180: `self.persistent` never gets used. How is this supposed to work?

effigies (commented):
Sorry, I've completely lost the plot on this PR. I don't understand what's going on, but there are some things that look wrong to me.

pydra/utils/hash.py
```diff
@@ -52,19 +54,144 @@
...
+        Returns
+        -------
+        Hash
+            _description_
+        """
+        try:
+            return self._hashes[key]
+        except KeyError:
+            pass
+        key_path = self.location / blake2s(str(key).encode()).hexdigest()
```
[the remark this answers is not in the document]
tclose: I think I just copied it from somewhere. I don't really know the pros & cons of different methods. Which would you recommend?

pydra/utils/hash.py
tclose: It is used on L241 (https://github.com/tclose/pydra/blob/a031ea5dfc1571d2171e0df4157bb97114b9b1eb/pydra/utils/hash.py#L241)

pydra/utils/hash.py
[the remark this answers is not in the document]
djarecka: looks like we were using `blake2s` before, but indeed doesn't seem to be the best choice (probably copied from somewhere). We can just use `blake2 ` or try [blake3](https://github.com/BLAKE3-team/BLAKE3), but could be also done in a separate pr

djarecka:
sorry, I'm sure we discussed it at some point, but I'm not sure about one important thing...  looks like if I move file around my filesystem I have no way of reusing the previous tasks now, is that right?

djarecka:
sorry, I was wrong, the task has correct hash when the file is the same with a different path. 

Perhaps you can add this test: https://github.com/djarecka/pydra/blob/db782c20890797bd58eb1d52545b7104f7d41aa4/pydra/engine/tests/test_node_task.py#L1568

(I was planning to create a PR to you, but I must have merged to the branch more things to my branch)

tclose:
> sorry, I was wrong, the task has correct hash when the file is the same with a different path.
> 
> Perhaps you can add this test: https://github.com/djarecka/pydra/blob/db782c20890797bd58eb1d52545b7104f7d41aa4/pydra/engine/tests/test_node_task.py#L1568
> 
> (I was planning to create a PR to you, but I must have merged to the branch more things to my branch)

Ok, nice addition

djarecka:
I've just realized that we should also have a similar test when the persistent cache is used, but realized that it's not enough to set `PYDRA_HASH_CACHE`. Sorry, for asking for more work, but can we have some test when the persistent cache is used together with running task or workflow

tclose:
@djarecka I have added a new test to ensure that the persistent cache gets hit during the running of tasks. Let me know if there is anything else you need me to do

pydra/engine/tests/test_node_task.py
```diff
@@ -1599,3 +1604,62 @@ def test_task_files_cachelocations(plugin_dask_opt, tmp_path):
...
+            yield from super().byte_chunks(**kwargs)
+
+    @property
+    def contents(self):
+        if self._contents is not None:
+            return self._contents
+        return super().contents
+
+
+def test_task_files_persistentcache(tmp_path):
```
djarecka, line 1635: @tclose - where are you setting the persistent cach path? i.e. `PYDRA_HASH_CACHE`

tclose: At @ghisvail's suggestion, the hash cache is stored in a system-dependent user cache directory using `platformdirs.user_cache_dir` by default (e.g. `/Users/<username>/Library/Caches/pydra/<version-number>` on MacOS).

I have just tweaked the code so that it is now put in a `hashes` subdirectory of that user cache dir (accessible in the `pydra.utils.user_cache_dir` variable) just in case any other cache data needs to be stored at some point in the future.

djarecka: oh, got it! Sorry I missed that. I've just realized that I made a typo when I was setting `PYDRA_HASH_CACHE` and that's why it was not saving the hashes there, and was confused where this is being saved..

djarecka:
@tclose - thanks so much for the work!

PR #709, opened 2023-09-29
Bugfix for optional output template files (i.e. of type Union[Path, bool]) when set to False

tclose:
## Types of changes

## Summary

For options such as the following where an optional output file (with a template) type Union[Path, bool], where the value is set to False, the collection of outputs was failing when attempting to coerce the boolean to a File

```
(
    "unused",
    attr.ib(
        type=ty.Union[Path, bool],
        default=False,
        metadata={
            "argstr": "--not-used",
            "output_file_template": "out.txt",
            "help_string": "dummy output",
        },
    ),
),

```

This bug-fix omits these fields from the collected outputs

## Checklist
- [ x ] I have added tests to cover my changes (if necessary)

### DIFF
--- a/pydra/engine/specs.py
+++ b/pydra/engine/specs.py
@@ -452,9 +452,14 @@ class ShellOutSpec:
             input_value = getattr(inputs, fld.name, attr.NOTHING)
             if input_value is not attr.NOTHING:
                 if TypeParser.contains_type(FileSet, fld.type):
-                    label = f"output field '{fld.name}' of {self}"
-                    input_value = TypeParser(fld.type, label=label).coerce(input_value)
-                additional_out[fld.name] = input_value
+                    if input_value is not False:
+                        label = f"output field '{fld.name}' of {self}"
+                        input_value = TypeParser(fld.type, label=label).coerce(
+                            input_value
+                        )
+                        additional_out[fld.name] = input_value
+                else:
+                    additional_out[fld.name] = input_value
             elif (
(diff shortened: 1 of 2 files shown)


pydra/engine/specs.py
```diff
@@ -451,9 +451,14 @@ def collect_additional_outputs(self, inputs, output_dir, outputs):
...
-                    input_value = TypeParser(fld.type, label=label).coerce(input_value)
-                additional_out[fld.name] = input_value
+                    if input_value is not False:
+                        label = f"output field '{fld.name}' of {self}"
+                        input_value = TypeParser(fld.type, label=label).coerce(
+                            input_value
+                        )
+                        additional_out[fld.name] = input_value
+                else:
+                    additional_out[fld.name] = input_value
```
tclose, line 461: This line handles the case where the is a corresponding input with the same name as an output but it isn't of file-type. Is this behaviour intended that the output receives the value of the input field with the same name by default, and therefore should be tested, or just removed?

effigies (approved):
This seems to work. I agree it would be good to understand the use case for L462 and either remove or test it.

pydra/engine/specs.py
djarecka: i think it should be removed. I'm not sure what I had in mind...

ghisvail: I removed it. We can reinstate it later if a use case for it emerges.

effigies:
Is one the nested `if`s redundant?

PR #729, opened 2023-12-30
add slurm dockerfile

adi611:
## Types of changes

## Summary

add slurm dockerfile for reference

## Checklist


tclose:
Would be good to link this directly to the image used in the CI

djarecka:
we have this image in CI, but we should probably update it

tclose:
We could set it up as a package that gets automatically re-built at regular intervals

PR #733, opened 2024-02-25
BYO worker

tclose:
## Types of changes

## Summary
Adds the ability to supply your own worker, i.e. one not already defined in the pydra core, without it needing to monkey patch the pydra.engine.workers.WORKERS dict

## Checklist

- [ x ] I have added tests to cover my changes (if necessary)

### DIFF
--- a/pydra/engine/submitter.py
+++ b/pydra/engine/submitter.py
@@ -1,9 +1,10 @@
 """Handle execution backends."""
 
 import asyncio
+import typing as ty
 import pickle
 from uuid import uuid4
-from .workers import WORKERS
+from .workers import Worker, WORKERS
 from .core import is_workflow
 from .helpers import get_open_loop, load_and_run_async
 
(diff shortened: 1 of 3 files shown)


pydra/engine/submitter.py
```diff
@@ -16,24 +17,31 @@
...
         """
         Initialize task submission.
 
         Parameters
         ----------
-        plugin : :obj:`str`
-            The identifier of the execution backend.
+        plugin : :obj:`str` or :obj:`pydra.engine.core.Worker`
+            Either the identifier of the execution backend or the backend itself.
             Default is ``cf`` (Concurrent Futures).
```
effigies, line 29: Should document the kwargs.

```suggestion
        \*\*kwargs ...
```

pydra/engine/submitter.py
```diff
@@ -16,24 +17,31 @@
...
 
         """
         self.loop = get_open_loop()
         self._own_loop = not self.loop.is_running()
-        self.plugin = plugin
-        try:
-            self.worker = WORKERS[self.plugin](**kwargs)
-        except KeyError:
-            raise NotImplementedError(f"No worker for {self.plugin}")
+        if isinstance(plugin, Worker):
```
effigies, line 33: Do we want to allow people to pass uninitialized classes, as well as pre-initialized? Here's an approach that looks up a class for `str`, initializes a class for `type`:

```python
if isinstance(plugin, str):
    plugin = WORKERS.get(plugin)
    if plugin is None:
        raise NotImplementedError(f"No worker for {plugin}")
if isinstance(plugin, type):
    plugin = plugin(**kwargs)

self.worker = plugin
self.plugin = plugin.plugin_name
```

Given that we're allowing initialization options, it seems weird not to permit classes, to me. But I can also see a case for keeping the options fewer. Up to you.

Alternately, we could get rid of `kwargs`. People can use defaults or pass a fully initialized `Worker`.

effigies (approved):
I agree with this approach in general. Small comments for your consideration.

pydra/engine/submitter.py
tclose: done

pydra/engine/submitter.py
tclose: Have changed it to a class instead of an instance

PR #734, opened 2024-03-17
Add note instead of catching and raising unhashable exception

tclose:
## Types of changes

## Summary

This is minor change to stop catching specific exceptions and re-raising generic UnhashbleExceptions, in favour of adding a note to the original exception to aid debugging in an editor. Currently, when debugging in an editor, the execution breaks at the UnhashableException rather than the underlying cause of the exception.

The only change in functionality is that of the type of exception raised

## Checklist


effigies (approved):
It's a shame that `raise x from y` disrupts debugging. That does seem like an appropriate way to add context.

PR #742, opened 2024-03-25
Handle None types

### LINKED ISSUE #744: TypeError: TypeParser doesn't know how to handle args ((<class 'pathlib.Path'>, <class 'NoneType'>)) for <class 'types.UnionType'>
The following task definition:

```python
@task
@annotate({"return": {"smoothed_image": Path}})
def smooth_image(input_image: Path, smoothed_image: Path | None) -> Path:
    ...
```

triggers an error:

`TypeError: TypeParser doesn't know how to handle args ((<class 'pathlib.Path'>, <class 'NoneType'>)) for <class 'types.UnionType'> types in 'smoothed_image' field of Inputs `

Notation `X | None` is becoming popular, and sometimes autoformatted by linters like Ruff. It would be great if Pydra's internal parsing mechanism were able to inject the necessary required / optional metadata to the fields, or at least not crash since this is a valid function definition.

tclose:
## Types of changes

## Summary

* Handles None in field types, e.g. `int | None` instead of `typing.Optional[int]`

## Checklist

_added by @ghisvail_
Closes #744

### DIFF
--- a/pydra/utils/tests/test_typing.py
+++ b/pydra/utils/tests/test_typing.py
@@ -120,11 +120,24 @@ def test_type_check_basic15():
     TypeParser(ty.Union[Path, File, float])(lz(int))
 
 
+@pytest.mark.skipif(sys.version_info < (3, 10), reason="No UnionType < Py3.10")
+def test_type_check_basic15a():
+    TypeParser(Path | File | float)(lz(int))
+
+
 def test_type_check_basic16():
     with pytest.raises(
         TypeError, match="Cannot coerce <class 'float'> to any of the union types"
     ):
         TypeParser(ty.Union[Path, File, bool, int])(lz(float))
+
+
+@pytest.mark.skipif(sys.version_info < (3, 10), reason="No UnionType < Py3.10")
+def test_type_check_basic16a():
+    with pytest.raises(
+        TypeError, match="Cannot coerce <class 'float'> to any of the union types"
+    ):
+        TypeParser(Path | File | bool | int)(lz(float))
 
 
 def test_type_check_basic17():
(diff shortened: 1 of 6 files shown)


pydra/utils/typing.py
```diff
@@ -20,6 +21,11 @@
     # Python < 3.8
     from typing_extensions import get_origin, get_args  # type: ignore
 
+try:
+    UNION_TYPES = (ty.Union, types.UnionType)
+except AttributeError:
+    UNION_TYPES = (ty.Union,)  # Python < 3.10
```
ghisvail, line 27: Favor explicit version checks over try / except, see rationales [here](https://learn.scientific-python.org/development/patterns/backports/#conditional-usage)

```suggestion
if sys.version_info >= (3, 10):
    UNION_TYPES = (ty.Union, types.UnionType)
else:
    UNION_TYPES = (ty.Union,)
```

ghisvail:
I did a superficial read of the diff and what I checked so far looked sensible 👍

pydra/utils/typing.py
tclose: accepted

pydra/utils/tests/test_typing.py
```diff
@@ -194,6 +207,12 @@ def test_type_check_fail2():
         TypeParser(ty.Union[Path, File])(lz(int))
 
 
+@pytest.mark.skipif(sys.version_info < (3, 10), reason="No UnionType < Py3.10")
+def test_type_check_fail2a():
+    with pytest.raises(TypeError, match="to any of the union types"):
+        TypeParser(Path | File)(lz(int))
```
ghisvail, line 213: Isn't it possible to run this test with a `from __future__ import annotations` for Python < 3.10?

tclose: `from __future__ import annotations` just leaves all annotations as strings I believe. Probably need to have a separate test module to handle that case, but since it doesn't look like that will be the way that types are handled going forward I have tended to avoid it

ghisvail (approved):

PR #743, opened 2024-03-26
Special handling of bytes_repr for mock filesets

tclose:
## Types of changes

## Summary
Implements special bytes_repr function for mocked filesets in(i.e. produced by Nifti1.mock("/path/to/mock")) as the fileset one yields the file mtimes in the cache key, which obviously don't exist for mocked files.

## Checklist

### DIFF
--- a/pydra/utils/hash.py
+++ b/pydra/utils/hash.py
@@ -55,7 +55,7 @@ __all__ = (
 )
 
 Hash = NewType("Hash", bytes)
-CacheKey = NewType("CacheKey", ty.Tuple[ty.Hashable, ty.Hashable])
+CacheKey = NewType("CacheKey", ty.Tuple[ty.Hashable, ...])
 
 
 def location_converter(path: ty.Union[Path, str, None]) -> Path:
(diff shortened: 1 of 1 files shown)


pydra/utils/hash.py
```diff
@@ -472,11 +472,29 @@ def bytes_repr_fileset(
...
+    cls = type(fileset)
+    yield f"{cls.__module__}.{cls.__name__}:".encode()
+    for key, chunk_iter in fileset.byte_chunks():
+        yield (",'" + key + "'=").encode()
+        yield from chunk_iter
+
+
+# Need to disable the mtime cache key for mocked filesets. Used in doctests
+@register_serializer(MockMixin)
+def bytes_repr_mock_fileset(
```
effigies, line 491: I don't like putting test code in main modules. What about adding `pydra.testing.fixtures` and including:

```
@pytest.fixture(scope='session')
def register_mock_mixin_serializer():
    @register_serializer(MockMixin)
    ...
```

Then in `conftest.py`:

```
from pydra.testing.fixtures import register_mock_mixin_serializer
```

This would allow downstream tools that need this feature to also put this import in their `conftest.py`.

tclose: I agree with the sentiment in general, but in this case I reckon that might be creating a gotcha that would take a bit of tracking down how to solve for little practical benefit, i.e if someone creates a new task package and wants to write a doctest using `NiftiGz.mock()`, then they would need to know that they also have to add this fixture

effigies (approved):

PR #750, opened 2024-04-12
WIP - BEP 028 prov 

yibeichan:
This PR was made during the 2024 BIDS meeting in Seattle after discussions with @effigies, some other discussions can be found here #https://github.com/bids-standard/BEP028_BIDSprov/issues/129
BEP028_prov doc see [here](https://docs.google.com/document/d/1vw3VNDof5cecv2PkFp7Lw_pNUTUo8-m8V4SIdtGJVKs/edit) 
## Types of changes

`context.json`file
- replace [openprov context](https://raw.githubusercontent.com/openprov/prov-jsonld/69964ed16818f78dc5f71bdf97add026288291d4/context.json) with [BEP028 prov context](https://raw.githubusercontent.com/bids-standard/BEP028_BIDSprov/master/context.json)
- adding customized fields `wasStartedBy`, `wasEndedBy`, and `hadActivity`. 
    - Those three were used in `audit.py` but not defined in the prov context. Those are important for connecting `audit_start`, `monitor`, and `finalize_audit`. 
    - If we view `audit_start`, `monitor`, and `finalize_audit` as subactivities of the activity of a given task, the current BEP028 doesn't define the connection between activities but only allows connecting two activities through entities. 

`audit.py`
- fixed typos such as updating `startedAtTime` as `StartedAtTime`
- dropped `entity_generated`, replaced it with a more specific field; in this case is `runtime` but it can be something else depending on the situation. `entity_generated` can be confused with `entity` in the prov doc, which is supposed to be either input files or output files related to an activity.
- changing [`"AssociatedWith": version_cmd`](https://github.com/nipype/pydra/blob/e52e32b05816df9908b612fbc8425faed9ead86b/pydra/engine/audit.py#L227) to 
a dictionary, which is supposed to be an `Agent`, but I couldn't get the `Label`, which is `Name of the software`
    - I put a comment about [`command`](https://github.com/nipype/pydra/blob/e52e32b05816df9908b612fbc8425faed9ead86b/pydra/engine/audit.py#L184C9-L184C16), currently we only get command when the task is a shell command task. If we want to be consistent with the prov, we probably should enable command for function tasks too. See the anonymous (that's Chris) comment on the prov doc.

## Summary

so the above changes are based on what we have in the prov doc, we probably need to dive deeper into `audit.py` and `messenger.py` (worth discussing in the next pydra meeting @djarecka)
@effigies suggested we collect messages for a workflow to generate prov records. 
for example, we can collect all messages at `finalize_audit` level into [`FileMessenger`](https://github.com/nipype/pydra/blob/e52e32b05816df9908b612fbc8425faed9ead86b/pydra/utils/messenger.py#L78C7-L78C20) using [`collect_messages`](https://github.com/nipype/pydra/blob/e52e32b05816df9908b612fbc8425faed9ead86b/pydra/utils/messenger.py#L174)

## Checklist

### DIFF
--- a/pydra/engine/audit.py
+++ b/pydra/engine/audit.py
@@ -62,8 +62,7 @@ class Audit:
             start_message = {
                 "@id": self.aid,
                 "@type": "task",
-                "startedAtTime": now(),
-                "executedBy": user_id,
+                "StartedAtTime": now(),
             }
 
         os.chdir(self.odir)
@@ -84,7 +83,7 @@ class Audit:
                     {
                         "@id": self.mid,
                         "@type": "monitor",
-                        "startedAtTime": now(),
+                        "StartedAtTime": now(),
                         "wasStartedBy": self.aid,
                     },
                     AuditFlag.PROV,
@@ -97,7 +96,7 @@ class Audit:
             result.runtime = gather_runtime_info(self.resource_monitor.fname)
             if self.audit_check(AuditFlag.PROV):
                 self.audit_message(
-                    {"@id": self.mid, "endedAtTime": now(), "wasEndedBy": self.aid},
+                    {"@id": self.mid, "EndedAtTime": now(), "wasEndedBy": self.aid},
                     AuditFlag.PROV,
                 )
                 # audit resources/runtime information
@@ -107,14 +106,14 @@ class Audit:
                     **{
                         "@id": self.eid,
                         "@type": "runtime",
-                        "prov:wasGeneratedBy": self.aid,
+                        "GeneratedBy": self.aid,
                     }
                 )
                 self.audit_message(entity, AuditFlag.PROV)
                 self.audit_message(
                     {
                         "@type": "prov:Generation",
-                        "entity_generated": self.eid,
+                        "runtime": self.eid,
                         "hadActivity": self.mid,
                     },
                     AuditFlag.PROV,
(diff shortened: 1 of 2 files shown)


satra:
@yibeichan - i don't think we should drop the prov context - we are still based on the prov model. that's more general and the bids context should still conform to prov.

yibeichan:
@satra do you mean the original one? [openprov context](https://raw.githubusercontent.com/openprov/prov-jsonld/69964ed16818f78dc5f71bdf97add026288291d4/context.json)
so we don't have to directly use/cite [BEP028 prov context](https://raw.githubusercontent.com/bids-standard/BEP028_BIDSprov/master/context.json) in pydra?

satra:
essentially bids prov was a simplification to keep the keys readable to people and perhaps should stay. however, the openprov context should be included or referenced. 

here is a linkml based context generated for prov: https://github.com/linkml/linkml-prov/blob/main/prov/jsonld/prov.model.context.jsonld

technically speaking we should be able to generate whatever we want as prov and then transform it to bep28 if we wanted using the bids context. pydra doesn't have to generate bep28, but could result in bep28. the following should work.

pydra jsonld + context -> expand -> compact/frame using bep28 context.

djarecka:
@satra -  what is relation between the context from the openprov repo and the one from linkml? do you suggest switching to the one from linkml. 
Seems to me that openprov doesn't have many useful objects, like `wasGeneratedBy`, etc, that is part of bids-prov

When you're saying transformation to bids-prov, you mean we would have to ignore some of the properties. I understand that they both point to prov model, just the coverage is different.

satra:
the linkml one looks more comprehensive and since chris did it, i have some trust in it as well. so i would lean towards using it and that also allows us to use the linkml model for other things.

PR #752, opened 2024-04-24
Fixes bug in super->subclass "auto casting" with union types

tclose:
## Types of changes

## Summary
* streamlines typing error messages
* fixes bug in super->subclass "auto casting" when the sub-classes are part of a union. If the super class is a super class of any of the union args then it passes the check.
* fixes bug with super->sub-class auto-casting
* adds handling for MultiInputObj

## Checklist


tclose:
Combined with #746 as they are both typing bugfixes and would have merge conflicts

Issue #761, opened 2024-07-15
splitting list of torch.Tensor causes "repeating" behavior of the output


wilke0818:
- What version of Pydra are you using? 
  - 0.23
- What were you trying to do?
  - Pass torch.Tensors to my function by splitting them as a list.
-  What did you expect will happen?
   - Expected the tensors to be passed to my function, have the function run, and get the result for each tensor.
   - if I had a function that just adds 2 to the values in a tensor, inputting the list of tensors [torch.Tensor([1]), torch.Tensor([2])] to be split into this function should create an output of [torch.Tensor([3]), torch.Tensor([4])]
- What actually happened?
  - Results "repeat" such that the output of a function (tested with multiple functions) match the result of the first split
  - To continue the example from above, [torch.Tensor([1]), torch.Tensor([2])] gives out [torch.Tensor([3]), torch.Tensor([3])]
  - Similarly, [torch.Tensor([1,2]), torch.Tensor([3,4])] gives out [torch.Tensor([3,4]), torch.Tensor([3,4])]
- Can you replicate this behavior?
  - yes, the toy example below matches the above description (this is simplified from what I was actually trying to do but I imagine fixing on this toy example will fix other issues)

```
import torch
import pydra

@pydra.mark.task
def test_task(fake_audio_input):
  return fake_audio_input+2

wf = pydra.Workflow(name='wf_test', input_spec=['audio_input'])

wf.split('audio_input', audio_input =[torch.tensor([1]),torch.tensor([2])]).combine('audio_input')

wf.add(test_task(name='testing', fake_audio_input=wf.lzin.audio_input))
# wf.combine('audio_input')
wf.set_output([('wf_out', wf.testing.lzout.out)])


with pydra.Submitter(plugin='cf') as sub:
    sub(wf)
print(wf.done)
# results = wf.result(return_inputs='val')
results = wf(plugin='cf')
print(results)
# output: [Result(output=Output(wf_out=tensor([3])), runtime=None, errored=False), Result(output=Output(wf_out=tensor([3])), runtime=None, errored=False)]
```
Similarly, using other tensor structures doesn't change the behavior:
```
wf.split('audio_input', audio_input =[torch.tensor([1,2]),torch.tensor([3,4])]).combine('audio_input')
# output: [Result(output=Output(wf_out=tensor([3, 4])), runtime=None, errored=False), Result(output=Output(wf_out=tensor([3, 4])), runtime=None, errored=False)]
```
or 
```
[...]

PR #762, opened 2024-07-18
Registered serializer for common classes of additional array-like objects

wilke0818:
## Types of changes

- Bug fix (non-breaking change which fixes an issue), see #761 

## Summary

- Registers serializers for 3 common array-like packages: torch, tensorflow, and pandas
- Copies existing methodology for serializing numpy in `bytes_repr_numpy`

## Checklist

  - No tests exist for `bytes_repr_numpy` which all changes are based off of
  - No, let me know if there is a specific place that references this hash code that needs updating

### DIFF
--- a/pydra/utils/hash.py
+++ b/pydra/utils/hash.py
@@ -585,6 +585,7 @@ if HAVE_NUMPY:
         else:
             yield obj.tobytes(order="C")
 
+
 if HAVE_PYTORCH:
 
     @register_serializer(torch.Tensor)
@@ -607,5 +608,6 @@ if HAVE_PANDAS:
     def bytes_repr_pandas(obj: pandas.DataFrame, cache: Cache) -> Iterator[bytes]:
         yield f"{obj.__class__.__module__}{obj.__class__.__name__}:".encode()
         yield from bytes_repr_numpy(obj.to_numpy(), cache)
+
 
 NUMPY_CHUNK_LEN = 8192
--- a/pydra/utils/hash.py
+++ b/pydra/utils/hash.py
@@ -43,6 +43,27 @@ except ImportError:
 else:
     HAVE_NUMPY = True
 
+try:
+    import pandas
+except ImportError:
+    HAVE_PANDAS = False
+else:
+    HAVE_PANDAS = True
+
+try:
+    import torch
+except ImportError:
+    HAVE_PYTORCH = False
+else:
+    HAVE_PYTORCH = True
+
+try:
+    import tensorflow
+except ImportError:
+    HAVE_TENSORFLOW = False
+else:
+    HAVE_TENSORFLOW = True
+
 __all__ = (
     "hash_function",
     "hash_object",
@@ -564,5 +585,27 @@ if HAVE_NUMPY:
         else:
             yield obj.tobytes(order="C")
 
+if HAVE_PYTORCH:
+
+    @register_serializer(torch.Tensor)
+    def bytes_repr_torch(obj: torch.Tensor, cache: Cache) -> Iterator[bytes]:
+        yield f"{obj.__class__.__module__}{obj.__class__.__name__}:".encode()
+        yield from bytes_repr_numpy(obj.numpy(), cache)
+
+
+if HAVE_TENSORFLOW:
+
+    @register_serializer(tensorflow.Tensor)
+    def bytes_repr_tensorflow(obj: tensorflow.Tensor, cache: Cache) -> Iterator[bytes]:
+        yield f"{obj.__class__.__module__}{obj.__class__.__name__}:".encode()
+        yield from bytes_repr_numpy(obj.numpy(), cache)
+
+
+if HAVE_PANDAS:
+
+    @register_serializer(pandas.DataFrame)
+    def bytes_repr_pandas(obj: pandas.DataFrame, cache: Cache) -> Iterator[bytes]:
+        yield f"{obj.__class__.__module__}{obj.__class__.__name__}:".encode()
+        yield from bytes_repr_numpy(obj.to_numpy(), cache)
 
 NUMPY_CHUNK_LEN = 8192


djarecka:
Thank you for your contribution! do you think you can create a simple test that shows the issue you are addressing?

effigies:
I don't think we should add these as-is, and would strongly suggest avoiding adding anything but numpy directly. However, we could use protocols to try to identify these and similar types that can be coerced to numpy arrays. Perhaps looking for an `__array__()` method, or one of these `.to_numpy()` methods?

djarecka:
@effigies - you mean not adding `bytes_repr_tensorflow`at all, and just try to re-use `bytes_repr_numpy` for tensorflow objects?

wilke0818:
@effigies I'm not aware of a common method like `__array__` that many of these libraries use. I considered the `to_numpy()` and `numpy()` route as you mention, but this would require changing `bytes_repr_numpy` to accept more than just numpy types, right? Doing so might then make the optional dependencies weird? I fallback on y'all's judgement though. @djarecka I will try to look at creating a test in the meantime.

effigies:
I mean something like:

```python
class ArrayLike(ty.Protocol):
    def __array__(self, *args, **kwargs): ...

@register_serializer
def bytes_repr_arraylike(obj: ArrayLike, cache: Cache) -> Iterator[bytes]:
        yield f"{obj.__class__.__module__}{obj.__class__.__name__}:{obj.size}:".encode()
        array = np.asanyarray(obj)
        if array.dtype == "object":
            yield from bytes_repr_sequence_contents(iter(array.ravel()), cache)
        else:
            yield array.tobytes(order="C")
```

As a quick proof of concept:


```python
from functools import singledispatch
import typing as ty

import numpy as np
import pandas as pd
import tensorflow as tf
import torch

class ArrayLike(ty.Protocol):
    def __array__(self, *args, **kwargs): ...

@singledispatch
def identify(obj: object) -> str:
    return obj.__class__.__name__

@identify.register
def _(obj: ArrayLike) -> str:
    return "ArrayLike"

print(f"{identify([0, 0])=}")  # list
print(f"{identify(np.array([0, 0]))=}")  # ArrayLike
print(f"{identify(pd.DataFrame({"a": [0, 0]}))=}")  # ArrayLike
print(f"{identify(tf.constant([0, 0]))=}")  # ArrayLike
print(f"{identify(torch.tensor([0, 0]))=}")  # ArrayLike
```

wilke0818:
@effigies I tried something like you mention, or at least how I understood it, and copied it below. I haven't written official test cases, but have tried it in a colab notebook where I was originally getting the error this was meant to fix. The issue is that though your proof of concept works, this doesn't work for DataFrames (which throw an unserializable error) nor torch Tensors which continue to fail silently. For some reason it did seem to work for tensorflow.

```python
class ArrayLike(ty.Protocol):
    def __array__(self, *args, **kwargs): ...

    def __array_interface__(self, *args, **kwargs): ...


if HAVE_NUMPY:
    @register_serializer
    def bytes_repr_arraylike(obj: ArrayLike, cache: Cache) -> Iterator[bytes]:
            yield f"{obj.__class__.__module__}{obj.__class__.__name__}:".encode()
            array = numpy.asanyarray(obj)
            yield f"{array.size}:".encode()
            if array.dtype == "object":
                yield from bytes_repr_sequence_contents(iter(array.ravel()), cache)
            else:
                yield array.tobytes(order="C")
```

effigies:
This works for me in a quick IPython session:

```python
class ArrayLike(ty.Protocol):
    def __array__(self, *args, **kwargs): ...

@register_serializer
def bytes_repr_arraylike(obj: ArrayLike, cache: Cache) -> Iterator[bytes]:
    yield f"{obj.__class__.__module__}.{obj.__class__.__name__}:".encode()
    array = np.asanyarray(obj)
    yield f"{array.dtype.str}[{array.shape}]:".encode()
    if array.dtype == "object":
        yield from bytes_repr_sequence_contents(iter(array.ravel()), cache)
    else:
        yield array.tobytes(order="C")
```

Testing with the test objects above:

```python
print(f"{b''.join(bytes_repr([0, 0], Cache()))}")
print(f"{b''.join(bytes_repr(np.array([0, 0]), Cache()))}")
print(f"{b''.join(bytes_repr(pd.DataFrame({"a": [0, 0]}), Cache()))}")
print(f"{b''.join(bytes_repr(tf.constant([0, 0]), Cache()))}")
print(f"{b''.join(bytes_repr(torch.tensor([0, 0]), Cache()))}")
```

```python
b'list:(\xb8\xa7\xf7\x08q;\x12\x80rw^\xc4\x12L:-\xb8\xa7\xf7\x08q;\x12\x80rw^\xc4\x12L:-)'
b'numpyndarray:2:\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
b'pandas.core.frame.DataFrame:<i8[(2, 1)]:\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
b'tensorflow.python.framework.ops.EagerTensor:<i4[(2,)]:\x00\x00\x00\x00\x00\x00\x00\x00'
b'torch.Tensor:<i8[(2,)]:\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```

Can you share a pandas dataframe that fails? Mixing and matching types, I see:

```python
>>> b''.join(bytes_repr(pd.DataFrame({'a': [0, 0], 'b': [1.5, 2.5], 'c': ['one', 'two']}), Cache()))
b'pandas.core.frame.DataFrame:|O[(2, 3)]:\xb8\xa7\xf7\x08q;\x12\x80rw^\xc4\x12L:-\xd1\x94\x8a?2\x95a\x85=\x8c\xb8|06\xb2\xfe\xa7\xce\xcb+\xd1\xbbS\xd2\xd1\xe5\xa9\x8c|O\xed\xf9\xb8\xa7\xf7\x08q;\x12\x80rw^\xc4\x12L:-\xe3\xcd\x89$_K\x88\x98\xf8\x19l\x95fk\xf1\xee;\xf5\xfdW|\xf5>\xf8\xf7\x94a\x8c\xfa\xff\x15\x04'
```

Another approach could be to use `yield pickle.dumps(obj)` (via @satra), which will work for pickle-able objects, but if a dataframe is not serializable, that may not work either.

wilke0818:
Running your above code in a Colab notebook produces the following for me:
image
and when including the DataFrame I get
image

I don't see where there should be a difference in the way iPython runs from using a Colab but perhaps there is.
https://colab.research.google.com/drive/1f-XYbg2Gr4m2YyzFj8zmoYzoNCLA0_jN?usp=sharing

effigies:
That's really weird. Can you activate `pdb` and see what object is actually causing the problem?

wilke0818:
apologies I am not sure how to use pdb in this context. I have added made the link to the colab to allow editing. will continue to try and figure it out if you don't have the time currently to look into it.

effigies:
I've never used colab and don't have time to figure it out. I don't understand why you're getting weak references in your environment, hence why I suggest using pdb. It appears to be a Jupyter notebook, so you should be able to run the `%pdb` magic before calling the erroring function and walk through the call stack to find where you got a weak reference. `pd.DataFrame([[0,0],[1,1]])` gives me a normal Pandas dataframe and serializes to numpy without problem.

wilke0818:
In hash.py, when calling with the DataFrame to bytes_repr, we go to the general function where it breaks the dataframe into 
`{'_is_copy': None, '_mgr': BlockManager
Items: Index(['b'], dtype='object')
Axis 1: RangeIndex(start=0, stop=2, step=1)
ObjectBlock: slice(0, 1, 1), 1 x 2, dtype: object, '_item_cache': {}, '_attrs': {}, '_flags': <Flags(allows_duplicate_labels=True)>}` using the \_\_dict__ function in pandas which is then passed to `bytes_repr_mapping_contents` and this whole process continues for a while until we try to hash the key/value pair of `'_flags': <Flags(allows_duplicate_labels=True)>`. The issue is that when `Flags.__dict__` is called, it returns

`{'_allows_duplicate_labels': True, '_obj': <weakref at 0x7eac8b3931f0; to 'DataFrame' at 0x7eac888e9a20>}`

and when the key/value `'_obj': <weakref at 0x7eac8b3931f0; to 'DataFrame' at 0x7eac888e9a20>}` is attempted to be hash, we try to run `__dict__` on the weakref which fails.

PR #768, opened 2025-02-17
Implementation of proposed syntax changes

tclose:
## Types of changes

## Summary

Implements proposed syntax changes in #692 

## Test modules (to fix)

- [x] test_python (design)
- [x] test_shell (design)
- [x] test_workflow (design)
- [x] test_boutiques
- [x] test_dockertask
- [x] test_environments
- [x] test_functions
- [x] test_graph
- [x] test_helpers_file
- [x] test_helpers_state
- [x] test_helpers
- [x] test_node_task
- [x] test_numpy_examples
- [x] test_profiles
- [x] test_shelltask_inputspec
- [x] test_shelltask
- [x] test_singularity
- [x] test_specs
- [x] test_state
- [x] test_submitter
- [x] test_task
- [x] test_tasks_files
- [x] test_workflow (engine)
- [x] test_hash
- [x] test_messenger
- [x] test_typing

## Checklist


satra:
@tclose - at this point, does it make sense to still keep two separate areas (design/engine) in the codebase? or simply merge them? i think we are all in agreement that this new design is the path we want to take.

tclose:
> @tclose - at this point, does it make sense to still keep two separate areas (design/engine) in the codebase? or simply merge them? i think we are all in agreement that this new design is the path we want to take.

Thanks :) although my plan was to keep the design area separate in the final state. Actually, in the refactor I had in mind I was planning to pull out environments and workers into separate namespaces, e.g. `pydra.workers`, `pydra.environments`, and have `design`, `tasks`, `workers` and `environments` (do we like plural or non-plural, i.e. `task`, `worker`, `environment`?) as expansion points where plugins can add support for new task-types, tasks, workers and environment classes.

I also wanted to pull out the helper functions into `pydra.utils` to avoid some circular import issues, leaving the only user-facing API within `pydra.engine` the `Submitter` class

pydra/engine/state.py
```diff
@@ -77,7 +164,15 @@ class State:
 
     """
 
-    def __init__(self, name, splitter=None, combiner=None, other_states=None):
+    def __init__(
+        self,
+        name,
+        definition: "TaskDef",
```
djarecka, line 170: `State` used to be independent on the task. Not sure if I understand why you decided to add the task here. 
I could see that you're using it in one place instead of passing input, but wondering if it's really worth it

tclose: You are right, I could probably just pass it into the method where it is used

djarecka: i believe you don't even need to pass it there, you just need input variables

pydra/design/base.py
```diff
@@ -244,12 +244,14 @@ class Arg(Field):
...
-        List of allowed values for the field.
     requires: list, optional
         Names of the inputs that are required together with the field.
-    xor: list[str], optional
-        Names of the inputs that are mutually exclusive with the field.
+    allowed_values: Sequence, optional
+        List of allowed values for the field.
+    xor: Sequence[str | None], optional
+        Names of args that are exclusive mutually exclusive, which must include
+        the name of the current field. If this list includes None, then none of the
```
satra, line 253: including the current field was not the case in nipype or the previous version of pydra i believe. reducing this kind of boilerplate would be helpful, and also if `xor` is commutative without specification (`a` xor `b` implies `b` xor `a` and should not have to be specified in both places - just one should do the other should be computed). also any checks to make sure xor combinations are valid or invalid (e.g. `a` xor `b`, `b` xor `c`, `c` xor `a`)

tclose: I was thinking it would be better as an argument to the `shell.define` decorator/function, so it would only need to be defined once and is really a "task-level" attribute

satra: so a list of lists? i.e. there could be multiple different `xor`s in a task.

tclose: Yes, maybe list[str] | list[list[str]] could be acceptable inputs

new_file_2.txt
djarecka, line 1: you have a few empty files, I'm guessing they were added by mistake

tclose: ah yes, some of the tests generate stray files in the CWD. Should try to track down which ones and clean them up

new-docs/source/tutorial/1-getting-started.ipynb
```diff
@@ -0,0 +1,340 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "\n",
+    "## Running your first task\n",
+    "\n",
+    "Pre-defined task definitions are installed under the `pydra.tasks.*` namespace by separate\n",
+    "task packages (e.g. `pydra-fsl`, `pydra-ants`, ...). To use a pre-defined task definition\n",
```
djarecka, line 92:
```suggestion
    "Pydra allows you to install independent packages with pre-defined tasks (e.g., `pydra-fsl`, `pydra-ants`). The task from the packages are installed under the `pydra.tasks.*`. To use a pre-defined task definition\n",
```

new-docs/source/tutorial/1-getting-started.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "and run it over every NIfTI file in the directory using the `TaskDef.split()` method"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from pydra.tasks.mrtrix3.v3_0 import MrGrid\n",
```
djarecka, line 154: not sure if this is meant to be fixed before the merge, but this import doesn't work right now

new-docs/source/tutorial/1-getting-started.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "It is straightforward to apply the same operation over a set of inputs using the `split()`\n",
+    "method. For example, if we wanted to re-grid all the NIfTI images stored in a directory,\n",
+    "such as the sample ones generated by the code below"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Then we can by importing the `MrGrid` shell-command task from the `pydra-mrtrix3` package\n",
```
djarecka, line 144:
```suggestion
    "such as the sample ones generated by `Nifti1` before, we can import the `MrGrid` shell-command task from the `pydra-mrtrix3` package\n",
```

new-docs/source/tutorial/1-getting-started.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "pprint(outputs.out_file)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "It is also possible to iterate over inputs in pairs/n-tuples. For example, if you wanted to use\n",
+    "different voxel sizes for different images, both the list of images and the voxel sizes\n",
+    "are passed to the `split()` method and their combination is specified by a tuple \"splitter\"\n",
```
djarecka, line 173: probably it should be unified how the splitters are called, since on the main page for splitter, it is called scalar

new-docs/source/tutorial/1-getting-started.ipynb
tclose: I created a prerelease version of pydra-mrtrix3, https://pypi.org/project/pydra-mrtrix3/3.0.4a17/, that should be compatible. Do you have that installed?

new-docs/source/tutorial/1-getting-started.ipynb
tclose: I think I would vote for "pairwise" or something like that (although I suppose you could have N>2 "pairs")

tclose: maybe "zipped" to match the Python function?

new-docs/source/tutorial/1-getting-started.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "When a task runs, a unique hash is generated by the combination of all the inputs to the\n",
+    "task and the operation to be performed. This hash is used to name the task cache directory\n",
+    "within the specified cache root. Therefore, if you use the same cache\n",
+    "root and in a subsequent run the same task is executed with the same\n",
+    "inputs, then the path of its cache directory will be the same, and if Pydra finds\n",
+    "existing results at that path, then the outputs generated by the previous run will be\n",
+    "reused.\n",
+    "\n",
+    "This cache will grow as more runs are called, therefore care needs to be taken to ensure\n",
+    "there is enough space on the target disk. Since the cache will be constantly To specify\n",
```
djarecka, line 271:
```suggestion
    "there is enough space on the target disk.\n",
```

new-docs/source/tutorial/2-advanced-execution.ipynb
```diff
@@ -0,0 +1,428 @@
...
+    "with submitter:\n",
+    "    result2 = submitter(mrgrid_varying_vox2)\n",
+    "\n",
+    "# Check that the output directory is the same for both runs\n",
+    "assert result2.output_dir == result1.output_dir\n",
+    "\n",
+    "# Change the voxel sizes to resample the NIfTI files to for one of the files\n",
+    "mrgrid_varying_vox2.voxel[2] = [0.25]\n",
+    "\n",
+    "# Result from previous run is reused as the task and inputs are identical\n",
```
djarecka, line 245: that sentence probably should be changed. However I don't remember now if pydra still collects the old results for elements of the splits that have not changed the voxel value

new-docs/source/tutorial/1-getting-started.ipynb
djarecka: yes, I have `3.0.4a17`, and get an error:
```
>>> import pydra.tasks.mrtrix3.v3_0
  File "<stdin>", line 1, in <module>
  File "/Users/dorota/miniconda3/envs/pydra/lib/python3.11/site-packages/pydra/tasks/mrtrix3/v3_0/__init__.py", line 8, in <module>
    from .fivettgen_fsl import FivettGen_Fsl
  File "/Users/dorota/miniconda3/envs/pydra/lib/python3.11/site-packages/pydra/tasks/mrtrix3/v3_0/fivettgen_fsl.py", line 13, in <module>
    class FivettGen_Fsl(specs.ShellDef["FivettGen_Fsl.Outputs"]):
  File "/Users/dorota/miniconda3/envs/pydra/lib/python3.11/site-packages/pydra/tasks/mrtrix3/v3_0/fivettgen_fsl.py", line 66, in FivettGen_Fsl
    mask: ImageIn | None = shell.arg(
                           ^^^^^^^^^^
TypeError: arg.__init__() got an unexpected keyword argument 'xor'
```

new-docs/source/tutorial/2-advanced-execution.ipynb
tclose: If I understand what you mean old results should be collected shouldn't they? Because they will be stored in a separate "job" directory, which should match the hash of subsequent jobs

new-docs/source/tutorial/1-getting-started.ipynb
tclose: ah ok, I need to make a new release to match the changes to the xor handling

new-docs/source/tutorial/4-python.ipynb
```diff
@@ -0,0 +1,260 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "\n",
+    "## Docstring parsing\n",
+    "\n",
+    "Instead of explicitly providing help strings and output names in `inputs` and `outputs`\n",
+    "arguments, if the function describes the its inputs and/or outputs in the doc string, \n",
+    "in either reST, Google or NumpyDoc style, then they will be extracted and included in the\n",
+    "input or output fields\n"
```
djarecka, line 155: the output doesn't have `c` and `d`, just `out` and no `help`. Not sure if this is by design, or it is not working properly. If it's by design, perhaps it could be explained

tclose: no, that is a bug. Will have a look

new-docs/source/tutorial/5-shell.ipynb
```diff
@@ -0,0 +1,442 @@
...
+    "# Shell-tasks"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Command-line templates\n",
+    "\n",
+    "Shell task specs can be defined using from string templates that resemble the command-line usage examples typically used in in-line help. Therefore, they can be quick and intuitive way to specify a shell task. For example, a simple spec  for the copy command `cp` that omits optional flags,"
```
djarecka, line 16:
```suggestion
    "Shell task specs can be defined using string templates that resemble the command-line usage examples typically used in in-line help. Therefore, they can be quick and intuitive way to specify a shell task. For example, a simple spec  for the copy command `cp` that omits optional flags,"
```

new-docs/source/tutorial/1-getting-started.ipynb
tclose: I think this suggestion deletes some preceding lines unintentionally

new-docs/source/tutorial/2-advanced-execution.ipynb
djarecka: yes, I would say that most of the results should be collected from previous jobs. Just want to confirm, since the description was focused on the main directory.

pydra/design/tests/test_python.py
```diff
@@ -0,0 +1,394 @@
...
+        return a * 2
+
+    SampleDef = python.define(func)
+
+    assert issubclass(SampleDef, PythonDef)
+    inputs = sorted(list_fields(SampleDef), key=sort_key)
+    outputs = sorted(list_fields(SampleDef.Outputs), key=sort_key)
+    assert inputs == [
+        python.arg(name="a", type=int),
+        python.arg(name="function", type=ty.Callable, hash_eq=True, default=func),
```
djarecka, line 26: I know that probably it is not as common, but what if the argument is called "function"?

new-docs/source/tutorial/4-python.ipynb
djarecka: yes, I saw the tests, and they seem to work well, but they do have explicit name of output in the decorator, i.e., `@python.define(outputs=["c", "d"])`

pydra/design/tests/test_python.py
tclose: It is a reserved name, it should raise an error when the task is defined

tclose: Have implemented checks (and tests) for this

tclose:
@djarecka @effigies @satra this PR is now ready to merge 😁 

NB: I reduced the number of tests that the slurm workers were being run over to just test_submitter, as I that module it covers all the relevant code pretty well, and they were taking ages to run over the larger test modules

satra:
@tclose - awesome work. what should the next steps be? 

this PR seems to still have the `design` folder in which all the new work has been done. do you want to merge and then reorganize? or reorganize before merging.

tclose:
Most of the new work is in `design` but I have ended up significantly touching a lot of the files in `engine` too. My idea is to break up the package into the following structure

```
└── pydra
    ├── design
    │   ├── boutiques??.py    
    │   ├── matlab??.py
    │   ├── python.py
    │   ├── shell.py
    │   └── workflow.py
    ├── engine
    │   ├── __init__.py
    │   ├── audit.py
    │   ├── graph.py
    │   ├── hooks.py
    │   ├── job.py
    │   ├── lazy.py
    │   ├── node.py
    │   ├── result.py
    │   ├── state.py
    │   ├── submitter.py
    │   └── workflow.py
    ├── environments
    │   ├── docker.py
    │   ├── modules??.py
    │   └── singularity.py
    ├── scripts
    │   └── run_pickled.py
    ├── tasks
    │   ├── common
    │   │   └── __init__.py
    │   ├── fsl
    │   │   └── __init__.py
    │   ├── mrtrix
    │   │   └── __init__.py
    │   ├── spm
    │   │   └── __init__.py
    │   └── testing
    │       └── __init__.py
    ├── utils
    │   ├── __init__.py
    │   ├── general.py
    │   ├── hash.py
    │   └── typing.py
    └── workers
        ├── cf.py
        ├── dask.py
        ├── debug.py
        ├── psij.py
        ├── sge.py
        └── slurm.py
```

noting that `design`, `environments`, `tasks` and `workers` don't have `__init__.py` files, i.e. are namespace packages, and are therefore points at which plugins can add additional options. In fact, I was thinking it would make sense to move `dask`, `psij` and `boutiques` into separate plugins similar to how the tasks are done

tclose:
If that looks good and you guys have finished reviewing the changes (or just want to have a look through the docs in any case), I can restructure it in this PR or create a separate one. I was thinking it would be better in a separate one as most links to the original code will be lost by the renaming.

tclose:
As per @effigies suggestion, I was also planning to rename

- `Task -> Job`
- `TaskDef -> Task`

In addition to this, I also had some suggestions renaming for renaming a few attribute names I found quite confusing when learning the code. No worries if you prefer the originals though.

- `Submitter.cache_dir -> cache_root`
- `Submitter.cache_locations -> readonly_caches`
- `Task.output_dir -> cache_dir`
- `State.cont_dim -> container_depth` or `unravel_depth` or something else (took me ages to work out what `cont_dim` is I read it as "continuation dimension")

satra:
- could `design` be renamed to `wrapper`?
- would like to avoid having task namespaces (spm/fsl/etc) in pydra core if possible. (i'm not strongly opposed, but just a preference to keep pydra clean from neuroimagiing)

the renames proposed seem reasonable, although i don't quite follow Job and Task semantics cleanly, although their former counterparts also need to us to be steeped in pydra lore :)

moving some of the workers into separate plugin packages seems reasonable to me. 

i think we should make as many of these changes and get this PR to be the replacement for main and new starting point for development when merged.

tclose:
> * could `design` be renamed to `wrapper`?

It could, although would that make sense for workflows, they aren't wrapping around anything per se?

> * would like to avoid having task namespaces (spm/fsl/etc) in pydra core if possible. (i'm not strongly opposed, but just a preference to keep pydra clean from neuroimagiing)

Yes, sorry for the confusion, I just put them in for illustration. Yes, the plan is to have them in separate plugin packages

> the renames proposed seem reasonable, although i don't quite follow Job and Task semantics cleanly, although their former counterparts also need to us to be steeped in pydra lore :)

I will add a glossary to the docs to try to explain the concepts behind Task, Node and Job.

A `Task` defines "what" is to be run, whereas a `Job` combines a `Task` with a `Submitter` object (which contains a `Worker` object), i.e. defining both "what" it is to be run with "how" it is to be run. So essentially, a task submitted to a worker becomes a job.

> moving some of the workers into separate plugin packages seems reasonable to me.
> 
> i think we should make as many of these changes and get this PR to be the replacement for main and new starting point for development when merged.

Ok, will append those changes to this PR

satra:
thanks @tclose 

for this:

> although would that make sense for workflows, they aren't wrapping around anything per se?

`wrapper` only in the sense they create an abstraction around some underlying method of execution (matlab script, shell script/command, python function, etc.,.). i can't wrap my head around what `design` means in that context. perhaps `decorators` or some such is easier? i'll keep thinking.

effigies:
Templates? Strategies? Runners?

tclose:
I don't mind the explicitness of `decorator`, although the package also contains the `*.arg` and `*.out` classes, and I was thinking of adding in some more functions into workflow to work with lazy fields (e.g. `cast`, `apply` and `convert`), also `shell.define` won't strictly be used as a decorator all the time.

`design` makes a lot of sense to me, in that if you are wanting to "design" a new task of any sort you import from that package whereas if you just want to use one that someone else has designed you can just import it from tasks, but that could just be an idiosyncrasy of my thinking.

satra:
how about: `taskbase`, `builder`, `middleware`. i have too much architectural design and design docs in my head that `design` seems to be incongruous in that sense.

<details> <summary> gpt suggestions </summary>

Based on your description, you're looking for a word that captures the idea of **constructing, shaping, or defining** new components (like tasks or workflows), without necessarily being limited to decorators, and one that contrasts nicely with `tasks` (as in pre-built or reusable components).

Here are some alternative names to consider for your `design` module:

### Top Picks
These aim to preserve your intent while fitting well semantically in a Pythonic ecosystem.

- **define** – emphasizes explicitly defining new tasks or elements. Works well alongside `shell.define`, and has an elegant, declarative feel.
- **build** – connotes construction of tasks or components. Simple, practical, and slightly more imperative than `define`.
- **create** – clear and user-friendly, very intuitive if your users are "creating" new functionality.
- **compose** – if there's a functional or compositional vibe to your framework, this has a nice ring to it.
- **construct** – slightly more formal than `build`, but clearly communicates task assembly.
- **make** – short and intuitive, although maybe too generic.

### If you want to preserve the "design" vibe but change the word:
- **schema** – implies structure or blueprint, especially if you're dealing with task interfaces or expected IO.
- **blueprint** – a bit verbose, but nicely evocative of reusable design patterns.
- **layout** – if visual or structural composition is part of the metaphor.
- **draft** – emphasizes preliminary or customizable task creation.

### Some bolder or more thematic ideas:
- **forge** – if you want a bit more flair; evokes crafting or forming something new.
- **craft** – similar to `forge`, but softer and perhaps more artisanal.
- **pattern** – highlights reusable structure, could work if you lean on metaprogramming or templates.
- **spec** – short for specification, suitable if your module is where users define behavior contracts.

</details>

`build` or `compose` sounds short and simple to me.

tclose:
`compose` is good I think (good old GPT)

tclose:
Yeah, I actually like `compose` better than `design` :)

satra:
> `compose` is good I think (good old GPT)

let's go with that.

docs/source/tutorial/7-canonical-form.ipynb
```diff
@@ -0,0 +1,262 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Canonical task form\n",
+    "\n",
+    "Under the hood, all Python, shell and workflow tasks generated by the\n",
+    "`pydra.compose.*.define` decorators/functions are translated to\n",
+    "[dataclasses](https://docs.python.org/3/library/dataclasses.html) by the\n",
+    "[Attrs](https://www.attrs.org/en/stable/). While the more compact syntax described\n",
```
effigies, line 12:
```suggestion
    "[attrs](https://www.attrs.org/en/stable/) library. While the more compact syntax described\n",
```

docs/source/tutorial/7-canonical-form.ipynb
```diff
@@ -0,0 +1,262 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Canonical task form\n",
+    "\n",
+    "Under the hood, all Python, shell and workflow tasks generated by the\n",
+    "`pydra.compose.*.define` decorators/functions are translated to\n",
+    "[dataclasses](https://docs.python.org/3/library/dataclasses.html) by the\n",
```
effigies, line 11:
```suggestion
    "[dataclass](https://docs.python.org/3/library/dataclasses.html)-like classes by the\n",
```

docs/source/examples/glm.ipynb
effigies, line 1: Just trying to run this, I get:

```
Task execution failed
Full crash report for 'FullWorkflow' job is here: [/var/home/chris/.cache/pydra/0.25.dev473](http://localhost:8888/var/home/chris/.cache/pydra/0.25.dev473)+g23e187b2[/run-cache/workflow-e6c509179a1a9fb0452dd4b4ef6a25ca/_error.pklz](http://localhost:8888/run-cache/workflow-e6c509179a1a9fb0452dd4b4ef6a25ca/_error.pklz)

Result(output_dir=PosixPath('/var/home/chris/.cache/pydra/0.25.dev473+g23e187b2/run-cache/workflow-e6c509179a1a9fb0452dd4b4ef6a25ca'), outputs=None, runtime=None, errored=True, task=FullWorkflow(output_dir=PosixPath('/var/home/chris/Projects/pydra/pydra/docs/source/outputs/6_glm')))
```

I haven't tried to debug.

docs/source/explanation/typing.rst
```diff
@@ -0,0 +1,78 @@
...
+Pydra is designed so that strict and specific typing can be used, but is not
+unnecessarily strict, if it proves too burdensome. Therefore, upstream fields that are
+typed as super classes  (or as ``typing.Any`` by default) of the task input they are
+connected to will be automatically cast to the subclass when the task is run.
+This allows workflows and tasks to be easily connected together
+regardless of how specific typing is defined in the task definition. This includes
+file format types, so a task that expects a ``fileformats.medimage.NiftiGz`` file can
+be connected to a task that outputs a ``fileformats.generic.File`` file.
+Therefore, the only cases where a typing error will be raised are when the upstream
+field can't be cast or coered to the downstream field, e.g. a ``fileformats.medimage.DicomSeries``
```
effigies, line 74:
```suggestion
field can't be cast or coerced to the downstream field, e.g. a ``fileformats.medimage.DicomSeries``
```

effigies (commented):
I've gone through most if not all of the docs, including reading through and attemting to run the notebooks.

My main problem is that I can't build the documentation without setting `nbsphinx_allow_errors = True`, or run all of the notebooks without error. In almost all cases, the problem is that `mrtrix3` depends on `libtiff.so.5`, but only `libtiff.so.6` can be installed with conda-forge, and the Fedora 41 environment I'm working in also can't readily install `libtiff.so.5`. Just a warning that the current situation may be brittle and we should be prepared to find a longer-term fix. Possibly just using the `Environment`s to run mrtrix3 tools with Docker, as in tutorial 6.

docs/source/tutorial/2-advanced-execution.ipynb
```diff
@@ -0,0 +1,437 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import tempfile\n",
+    "from pydra.tasks.mrtrix3.v3_0 import MrGrid\n",
+    "from pydra.engine.environments import Docker\n",
```
effigies, line 327:
```suggestion
    "from pydra.environments.docker import Docker\n",
```

effigies (commented):
This should resolve the current test failure in docs.

effigies (commented):
This has moved to:

```python
from pydra.environments import docker

docker.Environment(...)
```

from something more like:

```python
from pydra.environments import Docker

Docker(...)
```

What's the reasoning here? If we want to be able to so something like:

```python
env_mod = __import__(f'pydra.environments.{env_type}')
env_mod.Environment(...)
```

I would instead use

`docker.py`:
```py
class Docker(base.ContainerEnv):
    ...

# Alias `Docker` to `docker.Environment`
Environment = Docker
```

`__init__.py`:
```py
import docker
from docker import Docker

__all__ = [
  'Docker',
  'docker',
  ...
]
```

(We could do something lazier if we're worried about import time.)

docs/source/tutorial/2-advanced-execution.ipynb
```diff
@@ -0,0 +1,437 @@
...
+    "test_dir = tempfile.mkdtemp()\n",
+    "\n",
+    "nifti_file = Nifti1.sample(test_dir, seed=0)\n",
+    "\n",
+    "# Instantiate the task, \"splitting\" over all NIfTI files in the test directory\n",
+    "# by splitting the \"input\" input field over all files in the directory\n",
+    "mrgrid = MrGrid(in_file=nifti_file, operation=\"regrid\", voxel=(0.5, 0.5, 0.5))\n",
+    "\n",
+    "# Run the task to resample all NIfTI files\n",
+    "outputs = mrgrid(environment=Docker(image=\"mrtrix3/mrtrix3\", tag=\"latest\"))\n",
```
effigies, line 338:
```suggestion
    "from pydra.environments import docker\n",
    "\n",
    "test_dir = tempfile.mkdtemp()\n",
    "\n",
    "nifti_file = Nifti1.sample(test_dir, seed=0)\n",
    "\n",
    "# Instantiate the task, \"splitting\" over all NIfTI files in the test directory\n",
    "# by splitting the \"input\" input field over all files in the directory\n",
    "mrgrid = MrGrid(in_file=nifti_file, operation=\"regrid\", voxel=(0.5, 0.5, 0.5))\n",
    "\n",
    "# Run the task to resample all NIfTI files\n",
    "outputs = mrgrid(environment=docker.Environment(image=\"mrtrix3/mrtrix3\", tag=\"latest\"))\n",
```

effigies (commented):
Okay, I've looked through what I think I'm going to be able to. I've left some comments earlier and a very minor nit-pick that I happened to look at closely.

Overall, I think the API is good. I do think simple things like `pydra.environments.Docker` and `pydra.workers.ConcurrentFutures` are more conventional than `pydra.environments.docker.Environment` and `pydra.workers.cf.Worker`. I'm okay providing alternative access, especially when that simplifies writing generic code, but most people won't be doing that.

I also think the `workers.base_worker` and `environments.base_environment` are excessively specific, when they are frequently import with `as base`... Why not just `workers.base` and `environments.base`?

docs/source/examples/glm.ipynb
tclose: Ah yes, I haven't got the GLM example working yet. Was thinking it could wait for a later PR

tclose:
> This has moved to:
> 
> ```python
> from pydra.environments import docker
> 
> docker.Environment(...)
> ```
> 
> from something more like:
> 
> ```python
> from pydra.environments import Docker
> 
> Docker(...)
> ```
> 
> What's the reasoning here? If we want to be able to so something like:
> 
> ```python
> env_mod = __import__(f'pydra.environments.{env_type}')
> env_mod.Environment(...)
> ```
> 
> I would instead use
> 
> `docker.py`:
> 
> ```python
> class Docker(base.ContainerEnv):
>     ...
> 
> # Alias `Docker` to `docker.Environment`
> Environment = Docker
> ```
> 
> `__init__.py`:
> 
> ```python
> import docker
> from docker import Docker
> 
> __all__ = [
>   'Docker',
>   'docker',
>   ...
> ]
> ```
> 
> (We could do something lazier if we're worried about import time.)

Essentially yes. It was also for aesthetic consistency with `shell.Task`, `python.Task` and `workflow.Task`. However, I wasn't completely comfortable with it to be honest, so happy to go with your alias approach.

The only thing is that to make it an extension point for new plugins we need to drop the `pydra/environments/__init__.py` module so you will need to do 

```python
from pydra.environments.docker import Docker

Docker(...)
```

tclose:
> Okay, I've looked through what I think I'm going to be able to. I've left some comments earlier and a very minor nit-pick that I happened to look at closely.
> 
> Overall, I think the API is good. I do think simple things like `pydra.environments.Docker` and `pydra.workers.ConcurrentFutures` are more conventional than `pydra.environments.docker.Environment` and `pydra.workers.cf.Worker`. I'm okay providing alternative access, especially when that simplifies writing generic code, but most people won't be doing that.
> 
> I also think the `workers.base_worker` and `environments.base_environment` are excessively specific, when they are frequently import with `as base`... Why not just `workers.base` and `environments.base`?

I had `workers.base` and `environments.base` but changed them as it was causing an issue with how pytest imports the modules, but I think this is something I can address with the pytest configuration intsead

tclose:
> I've gone through most if not all of the docs, including reading through and attemting to run the notebooks.
> 
> My main problem is that I can't build the documentation without setting `nbsphinx_allow_errors = True`, or run all of the notebooks without error. In almost all cases, the problem is that `mrtrix3` depends on `libtiff.so.5`, but only `libtiff.so.6` can be installed with conda-forge, and the Fedora 41 environment I'm working in also can't readily install `libtiff.so.5`. Just a warning that the current situation may be brittle and we should be prepared to find a longer-term fix. Possibly just using the `Environment`s to run mrtrix3 tools with Docker, as in tutorial 6.

I'm happy to switch to a different toolkit other than Mrtrix3 if it is easier once we have the Nipype interfaces ported across. It is just the only one that is in the new format at this point

docs/source/tutorial/5-shell.ipynb
```diff
@@ -0,0 +1,442 @@
...
+    "Cp = shell.define(\n",
+    "    \"cp <in_fs_objects:fs-object+> <out|out_dir:directory> \"\n",
+    "    \"-R<recursive> \"\n",
+    "    \"--text-arg <text_arg?> \"\n",
+    "    \"--int-arg <int_arg:int?> \"\n",
+    "    \"--tuple-arg <tuple_arg:int,str*> \"\n",
+    ")\n",
+    "\n",
+    "pprint(fields_dict(Cp))\n",
+    "pprint(fields_dict(Cp.Outputs))"
```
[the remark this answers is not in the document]
tclose: I have switched this to use the `task_help(Cp)` helper function instead. Will modify it so that help functions and factory defaults are printed

satra:
@tclose - instead of mrtrix, would it be helpful to just create a lightweight demo package to demonstrate and do tests with. my personal preference is that pydra should demonstrate a general framework rather than a neuroimaging specific framework. this demonstration package could rely on common standard unix shell commands and python standard library.   

these changes don't have to be done immediately and can wait after the PR is merged, unless you can see a very quick package build and deployment. also, we should update the the pydra task package template. in another template, our group convinced me to use poetry (and now poetry 2).

tclose:
Ok, I believe this is all good to go!! I have added a brief glossary to the docs explaining some of the new concepts and a few of the attributes I renamed (see docs/source/reference/glossary.rst).

Let me know if there is anything else, otherwise it would be great to get this merged and a 1.0a released so I can start updating the rest of my packages to point to that 😁

satra:
thanks tom. i will leave it to you and others to pull the trigger. 

regarding prereleases and auto (which we use), we may need to setup a few things: https://intuit.github.io/auto/docs/generated/next (may not be a bad thing to setup).

alternatively, we could merge into `master` (should we rename to `main`) and make a manual pre-release.

satra:
one last question is the hashing of arbitrary objects, especially given the issues we had with torch arrays. is there a default serializer that should handle any arbitrary object? it did seem from that issue that we had to write and register our own serializer, and that the default backstop was not sufficient. (see: https://github.com/sensein/senselab/blob/5f1af5b5305579c8bdd698837dc8048d5284229d/src/senselab/utils/data_structures/pydra_helpers.py#L17).

tclose:
> one last question is the hashing of arbitrary objects, especially given the issues we had with torch arrays. is there a default serializer that should handle any arbitrary object? it did seem from that issue that we had to write and register our own serializer, and that the default backstop was not sufficient. (see: https://github.com/sensein/senselab/blob/5f1af5b5305579c8bdd698837dc8048d5284229d/src/senselab/utils/data_structures/pydra_helpers.py#L17).

The backstop object bytes-repr is here 
https://github.com/nipype/pydra/blob/bc6b0a943ab1c8aaa8c207d5aa15988b90ee90af/pydra/utils/hash.py#L308-L346
I did make some changes to it so that it can handle __slots__ classes (as well as special handling for attrs-generated objects), so would be interested if it can now handle the PyTorch arrays. If not, I might be worth adding those bytes-repr functions to the hash module with guards for the case where torch is not installed

satra:
perhaps i'm doing something wrong, but let me know if this makes sense to you. it doesn't seem to be reflecting the object fully, i.e. the output doesn't change for the two different tensors.

```python
In [15]: a = torch.tensor([1, 2])

In [16]: list(bytes_repr(a, cache=cache))
Out[16]: [b'torch.Tensor:{', b'}']

In [17]: a = torch.tensor([3, 4])

In [18]: list(bytes_repr(a, cache=cache))
Out[18]: [b'torch.Tensor:{', b'}']
```

for a backstop i was hoping it would be similar to the cloudpickle/pickle option, where as long as it is pickleable we can potentially hash it.

tclose:
> https://intuit.github.io/auto/docs/generated/next

I had a look at the shipit pre-release docs, and I think that would work better > 1.0, as we don't really want to keep the current master as the default until 1.0 is released.

So if you guys are ok with it, I will pull the trigger on the merge and manually release a 1.0a. I'm also happy to rename the `master` branch to `main` in the process if you like.

satra:
whereas the `bytes_repr` works fine for numpy:

```python
In [21]: a = np.array([3, 4])

In [22]: list(bytes_repr(a, cache=cache))
Out[22]: 
[b'numpyndarray:2:',
 b'\x03\x00\x00\x00\x00\x00\x00\x00\x04\x00\x00\x00\x00\x00\x00\x00']

In [23]: a = np.array([1, 2])

In [24]: list(bytes_repr(a, cache=cache))
Out[24]: 
[b'numpyndarray:2:',
 b'\x01\x00\x00\x00\x00\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00']
```

tclose:
> perhaps i'm doing something wrong, but let me know if this makes sense to you. it doesn't seem to be reflecting the object fully, i.e. the output doesn't change for the two different tensors.
> 
> ```python
> In [15]: a = torch.tensor([1, 2])
> 
> In [16]: list(bytes_repr(a, cache=cache))
> Out[16]: [b'torch.Tensor:{', b'}']
> 
> In [17]: a = torch.tensor([3, 4])
> 
> In [18]: list(bytes_repr(a, cache=cache))
> Out[18]: [b'torch.Tensor:{', b'}']
> ```
> 
> for a backstop i was hoping it would be similar to the cloudpickle/pickle option, where as long as it is pickleable we can potentially hash it.

Interesting, `torch.Tensor` doesn't have slots and its `__dict__` is empty. We could just fallback to dir/getattr, although we will need to enclose it in try/except to catch invalid properties (I'm wondering whether we could just check `set(dir(obj)) - set(dir(type(obj)))` or whether that would miss things.

Cloudpickle would be guaranteed to be unique, but it is not guaranteed to produce the same representations between equivalent objects so you would get a lot of false-negative cache lookups, e.g.

```
import torch
import cloudpickle as cp

a = torch.tensor([1, 2])
cp1 = cp.dumps(a)

b = torch.tensor([1, 2])
cp2 = cp.dumps(b)

print(cp1 == cp2)  # False
```

tclose:
> > perhaps i'm doing something wrong, but let me know if this makes sense to you. it doesn't seem to be reflecting the object fully, i.e. the output doesn't change for the two different tensors.
> > ```python
> > In [15]: a = torch.tensor([1, 2])
> > 
> > In [16]: list(bytes_repr(a, cache=cache))
> > Out[16]: [b'torch.Tensor:{', b'}']
> > 
> > In [17]: a = torch.tensor([3, 4])
> > 
> > In [18]: list(bytes_repr(a, cache=cache))
> > Out[18]: [b'torch.Tensor:{', b'}']
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > for a backstop i was hoping it would be similar to the cloudpickle/pickle option, where as long as it is pickleable we can potentially hash it.
> 
> Interesting, `torch.Tensor` doesn't have slots and its `__dict__` is empty. We could just fallback to dir/getattr, although we will need to enclose it in try/except to catch invalid properties (I'm wondering whether we could just check `set(dir(obj)) - set(dir(type(obj)))` or whether that would miss things.
> 
> Cloudpickle would be guaranteed to be unique, but it is not guaranteed to produce the same representations between equivalent objects so you would get a lot of false-negative cache lookups, e.g.
> 
> ```
> import torch
> import cloudpickle as cp
> 
> a = torch.tensor([1, 2])
> cp1 = cp.dumps(a)
> 
> b = torch.tensor([1, 2])
> cp2 = cp.dumps(b)
> 
> print(cp1 == cp2)  # False
> ```

Maybe this could be addressed in a separate PR though...

satra:
thanks for the cloudpickle example - fine with addressing this in a separate PR.

satra:
woohoo!

PR #769, opened 2025-02-17
Develop: fixes for test_environments

tclose:
## Types of changes

## Summary
Fixes `test_environments` after syntax changes in #768 

## Checklist


satra:
@tclose - just a quick note that instead of building singularity you can install apptainer directly from binaries. example here: https://github.com/ReproNim/neurodocker/blob/eb4720e1e04f7347488d63fe05fcdafabaaac9bd/.github/workflows/pull-request.yml#L20

tclose:
> @tclose - just a quick note that instead of building singularity you can install apptainer directly from binaries. example here: https://github.com/ReproNim/neurodocker/blob/eb4720e1e04f7347488d63fe05fcdafabaaac9bd/.github/workflows/pull-request.yml#L20

Thanks, should I include it in the main test build and drop the separate singularity one?

satra:
> Thanks, should I include it in the main test build and drop the separate singularity one?

your call with the refactor as it depends on where singularity/apptainer gets used. it's a relatively quick install, so if it's helpful to bundle into a single test that's fine.

PR #782, opened 2025-04-02
Update CI to use build-and-inspect-python-package and tox-uv

effigies:
This PR is aimed at centralizing testing and modernizing the CI workflow. It:

1) Adds a `tox.ini` that allows testing on the latest versions of all dependencies (`py*-latest`), the lowest stated dependencies (`py*-min`) and pre-releases (`py*-pre`).
2) Builds the packages with https://github.com/hynek/build-and-inspect-python-package, an action that encapsulates current best practices, including cryptographic attestation and listing package contents to the CI summary.
3) Uploads using PyPI's [trusted publishing](https://docs.pypi.org/trusted-publishers/using-a-publisher/), which removes the need for a token and permits uploading cryptographic attestations.
4) Uses `tox-uv` to run the tests on CI, making it easier to reproduce issues seen in CI locally and vice-versa.

Some small issues were fixed in the process, including updating minimum versions based on running the `py311-min` environment. Apart from numpy, which I set at `>=1.26` so we don't become unintentionally incompatible with 1.x, the test dependencies are mostly quite recent.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,2 +1,0 @@
-[run]
-branch = True
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,7 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    groups:
+      actions-infrastructure:
+        patterns:
+          - "actions/*"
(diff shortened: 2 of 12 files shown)


tclose (approved):
Looks great :)

PR #790, opened 2025-04-09
Handle-none-outputs

tclose:
## Types of changes

## Summary

Handles the case of tasks and workflows without any outputs (i.e. return type is set to None). In this case the outputs are empty.

## Checklist

### DIFF
--- a/pydra/compose/base/helpers.py
+++ b/pydra/compose/base/helpers.py
@@ -185,10 +185,15 @@ def extract_function_inputs_and_outputs(
             input_defaults[p.name] = p.default
     if inputs is not None:
         if not isinstance(inputs, dict):
-            raise ValueError(
-                f"Input names ({inputs}) should not be provided when "
-                "wrapping/decorating a function as "
-            )
+            if non_named_args := [
+                i for i in inputs if not isinstance(i, Arg) or i.name is None
+            ]:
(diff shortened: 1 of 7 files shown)


pydra/compose/base/helpers.py
```diff
@@ -183,7 +183,7 @@ def extract_function_inputs_and_outputs(
         input_types[p.name] = type_hints.get(p.name, ty.Any)
         if p.default is not inspect.Parameter.empty:
             input_defaults[p.name] = p.default
-    if inputs:
+    if inputs is not None:
         if not isinstance(inputs, dict):
             raise ValueError(
                 f"Input names ({inputs}) should not be provided when "
```
effigies, line 189: Just looking at this, it seems like `inputs` could be a `list[str | Arg]`, but it isn't turned into a dict before this. This means we'll now error on an empty list as well as a populated list, but maybe we should just remove the list annotation?

effigies (approved):
Makes sense to me, good tests. One small thing I noticed that could be addressed here or later.

pydra/compose/base/helpers.py
tclose: Now test for list[Arg] and convert to dict[str, Arg] if they are named (raise an error otherwise)

PR #792, opened 2025-04-10
Serializing Task classes to/from dictionaries

tclose:
## Types of changes

## Summary
Adds two functions `serialize_task_class` and `unserialize_task_class` to enable you to serialise task definitions to dictionaries (i.e. to be written to JSON/YAML if desired)

## Checklist


effigies (approved):
Makes sense. `task_as_dict()` is a somewhat confusing name, though...

pydra/utils/general.py
```diff
@@ -351,7 +352,7 @@ def task_fields(task: "type[Task] | Task") -> _TaskFieldsList:
     )
 
 
-def task_dict(obj, **kwargs) -> dict[str, ty.Any]:
+def task_as_dict(obj, **kwargs) -> dict[str, ty.Any]:
```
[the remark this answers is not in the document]
tclose: I see what you mean, I was going for an analogy with `attrs.asdict`. We already have `task_fields` to extract the fields from both tasks and task outputs, should I change that to `get_fields` and `task_as_dict` to `get_values`? `task_class_as_dict` could be `serialize_task_class` or something. 

Not feeling inspired, open to suggestions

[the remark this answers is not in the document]
tclose: Ok, so I have gone with 

`task_fields` -> `get_fields`
`task_as_dict` -> `asdict`
`task_class_to_dict` -> `serialize_task_class`
`task_class_from_dict` -> `unserialize_task_class`

Sound good?

pydra/utils/tests/test_general.py
```diff
@@ -0,0 +1,87 @@
...
+
+def test_python_serialize_task_class(tmp_path):
+
+    dct = serialize_task_class(Add)
+    Reloaded = unserialize_task_class(dct)
+    assert get_fields(Add) == get_fields(Reloaded)
+
+    add = Reloaded(a=1, b=2)
+    assert add(cache_root=tmp_path / "cache").out_int == 3
+
```
djarecka, line 41: perhaps you could add a simple test that also check the `dct` itself, or at least what fields it should contain, i.e., "type", "name", "inputs", ...

pydra/utils/general.py
```diff
@@ -584,3 +585,132 @@ def get_plugin_classes(namespace: types.ModuleType, class_name: str) -> dict[str
...
+
+    Parameters
+    ----------
+    task_class : type[pydra.compose.base.Task]
+        The Pydra task class to convert.
+    filter : callable, optional
+        A function to filter out certain attributes from the task definition passed
+        through to `attrs.asdict`. It should take an attribute and its value as
+        arguments and return a boolean (True to keep the attribute, False to filter it out).
+    value_serializer : callable, optional
```
djarecka, line 610: I don't think I understand what is `value_serializer` and when do you use it

pydra/utils/general.py
```diff
@@ -584,3 +585,132 @@ def get_plugin_classes(namespace: types.ModuleType, class_name: str) -> dict[str
...
+) -> ty.Dict[str, ty.Any]:
+    """Converts a Pydra task class into a dictionary representation that can be serialized
+    (e.g. to a file), then reread and passed to an appropriate `pydra.compose.*.define`
+    method to recreate the task.
+
+    Parameters
+    ----------
+    task_class : type[pydra.compose.base.Task]
+        The Pydra task class to convert.
+    filter : callable, optional
```
djarecka, line 606: can we have one test that uses it.

djarecka (commented):
i left just a couple of questions/suggestions. But it's a great addition

pydra/utils/general.py
tclose: It gets passed through to [attrs.asdict](https://www.attrs.org/en/stable/api-attr.html#attr.asdict), will write a test

pydra/utils/tests/test_general.py
tclose: I didn't want to overspecify the test. I think as long as it can be serialized to a dictionary that should be enough. However, I will write a check to see that all Pydra objects have been serialized

pydra/utils/general.py
tclose: added

pydra/utils/general.py
tclose: Have rewritten the test to hopefully make it easier to follow

pydra/utils/general.py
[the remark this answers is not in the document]
tclose: as discussed

`task_class_to_dict` -> `unstructure`
`task_class_from_dict` -> `structure`

PR #795, opened 2025-04-14
Adds check to parse_command_line_template to catch options that don't have trailing field

tclose:
## Types of changes

## Summary

Catches options passed to parse_command_line_template that don't have an associated variable with them

## Checklist

### DIFF
--- a/pydra/compose/shell/tests/test_shell_fields.py
+++ b/pydra/compose/shell/tests/test_shell_fields.py
@@ -984,6 +984,13 @@ def test_shell_missing_executable_dynamic():
         )
 
 
+def test_shell_trailing_option_fail():
+
+    with pytest.raises(ValueError, match="Found an option without a field"):
+
+        shell.define("ls <directory:generic/directory> --unparameterized-option ")
+
+
 def test_shell_help1():
 
     Shelly = shell.define(
--- a/pydra/compose/shell/builder.py
+++ b/pydra/compose/shell/builder.py
@@ -512,6 +512,9 @@ def parse_command_line_template(
                 f"Found unknown token {token!r} in command line template: {template}"
             )
 
+    if option:
+        raise ValueError(f"Found an option without a field: {option!r}")
+
     remaining_pos = remaining_positions(arguments, len(arguments) + 1, 1)
 
     for argument in arguments:


pydra/compose/shell/tests/test_shell_fields.py
```diff
@@ -984,6 +984,13 @@ def test_shell_missing_executable_dynamic():
         )
 
 
+def test_shell_trailing_option_fail():
+
+    with pytest.raises(ValueError, match="Found an option without a field"):
+
+        shell.define("ls <directory:generic/directory> --unparameterized-option ")
```
effigies, line 991: What if I wanted to create an interface where some fields were non-optional?

```py
rmdir = shell.define("rm -r --preserve-root=all <directory:generic/directory>")
```

You can image a lot of other cases, like `fslmaths`, where you carve out some specific functionality to expose to the task caller.

tclose: You could do this currently with the "canonical" task form by explicitly setting the executable = ["rm", "-r", "--preserve-root=all"], but I can see how it might be convenient to use the template form to construct such tasks

tclose: The trouble is that there isn't a way as far as I can see to distinguish between an unparameterized flag followed by an argument from an option with a parameter.

tclose: Using `<ro|>` would fit in with the current parsing logic more easily (i.e. like `<out|output_field>` and `<modiify|modifiable_files>`, e.g. ```rmdir = shell.define("rm -r<ro|> --preserve-root=all<ro|> <directory:generic/directory>")```

tclose: You could do something like that currently with 

```python
rmdir = shell.define("rm <directory:generic/directory>", inputs=[shell.arg(name="recursive", argstr="-r", readonly=True), shell.arg(name="preserve_root", argstr="--preserve-root=all", readonly=True)])
```

The question is how often this use case comes up and how streamlined it needs to be I suppose

satra: i agree that this becomes a lower priority issue. it would be nice to be able to streamline but not the highest need right now in pydra. i'll let @effigies describe whether the current formulation would be sufficient for now.

tclose:
```python
rmdir = shell.define("rm |-r| |--preserve-root=all| <directory:generic/directory>")
```
as a shorthand for it (I like `<`, `>` and `|` for special characters as they can't be used in the command syntax)

effigies: Eh, I'm hesitant to make the spec language even more complicated. Default values might do the job. e.g.

```python
rmdir = shell.define("rm -r<recursive:bool=True> --preserve-root=<preserve_root:Literal['all']='all'> <directory:generic/directory>")
```

PR #817, opened 2025-04-29
moved psutil from test dep to core dep

### LINKED ISSUE #760: `psutil` dependency
## What version of Pydra are you using?
0.23
(Python 3.10.11)

## What were you trying to do?
Following the tutorial (commit [e1c91e3](https://github.com/nipype/pydra-tutorial/commit/e1c91e3c13cd310714b04623dc08c1c13a19eec6)) for [FunctionTask#Using Audit](https://github.com/nipype/pydra-tutorial/blob/master/notebooks/2_intro_functiontask.md#using-audit) part.

## What did you expect will happen?
I expected the provided code to run nicely.

## What actually happened?
```
ModuleNotFoundError: No module named 'psutil'
```
The error is due to the fact that `psutil` is included in the `tests` optional-dependencies only, although it seems like it is needed by the `Audit` features.

## Can you replicate the behavior? If yes, how?
Yes, here is how : 

Create a `test_pydra_tutorial.py` file containing:
```python
# Contents of test_pydra_tutorial.py
from pydra.mark import task
from pydra.utils.messenger import AuditFlag, PrintMessenger

@task
def add_var(a: int, b: int) -> {'out': int}:
	return a+b
task0 = add_var(
	a = 4, b = 5,
	audit_flags = AuditFlag.RESOURCE # This line causes the error
	)
task0()
print(task0.result())
```

Create a virtualenv and activate it
```bash
virtualenv pydra_test_venv
source pydra_test_venv/bin/activate
```

Then inside the virtualenv
```bash
pip install pydra
python test_pydra_tutorial.py
```

tclose:
## Types of changes

## Summary
Adds `psutil` as core dependency instead of just test. When resources are audited it imports the `pydra.utils.profiler` module, which has psutil as a dep.

Fixes #760

## Checklist

### DIFF
--- a/pydra/engine/tests/test_audit.py
+++ b/pydra/engine/tests/test_audit.py
@@ -302,3 +302,15 @@ def test_audit_all(
     # commented out to speed up testing
     collect_messages(tmpdir / funky._checksum, message_path, ld_op="compact")
     assert (tmpdir / funky._checksum / "messages.jsonld").exists()
+
+
+def test_audit_resource():
+
+    @python.define(outputs=["out"])
+    def AddVar(a: int, b: int) -> int:
+        return a + b
+
+    task0 = AddVar(a=4, b=5)
+    outputs = task0(audit_flags=AuditFlag.RESOURCE)
+
+    print(outputs)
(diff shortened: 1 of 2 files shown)


tclose:
I'm easy either way. An 'audit' extra sounds like a reasonable idea, we just need to make sure that audit isn't imported unless it is used

PR #819, opened 2025-05-02
added guard against missing implicit helps in custom compose packages

tclose:
## Types of changes

## Summary
In custom `compose` plugins, there may not be a way to implicitly pull out help strings, in which case they will be None by default in `ensure_field_objects`, but this was not being checked.

## Checklist

### DIFF
--- a/pydra/compose/base/helpers.py
+++ b/pydra/compose/base/helpers.py
@@ -66,7 +66,7 @@ def ensure_field_objects(
                 )
             else:
                 arg.name = input_name
-            if not arg.help:
+            if not arg.help and input_helps:
                 arg.help = input_helps.get(input_name, "")
         elif is_type(arg):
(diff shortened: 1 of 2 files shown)


pydra/compose/base/helpers.py
```diff
@@ -66,7 +66,7 @@ def ensure_field_objects(
                 )
             else:
                 arg.name = input_name
-            if not arg.help:
+            if not arg.help and input_helps:
```
effigies, line 69: If `input_helps` is `{}`, `arg.help` will no longer go from `None` or something Falsey to the specific value `""`. Is that okay?

tclose: Not sure, but will change so it does

effigies (approved):

PR #827, opened 2025-05-20
allow sequence types for fields with requires (can be empty)

tclose:
## Types of changes

## Summary

Allows task fields of Sequence type to have non-empty requires (as they can be empty if requirements are not met)

## Checklist


tclose:
This feature was already implemented within the `is_truthy_falsy` function

PR #870, opened 2026-08-02
[FIX] Submitt workflows to task queue

### LINKED ISSUE #869: Parallel execution lost when wrapping split tasks in a nested workflow
I'm having some problems splitting nested workflows to multiple processes. I'm using the `cf` worker and trying to have a workflow with a nested workflow. I'm new to Pydra, so it's possible I'm misunderstanding how nested workflows are expected to behave, but based on the documentation I expected split tasks to execute in parallel regardless of whether they are wrapped in a second workflow or not.

I'm using Pydra 1.0a9 with Python 3.14 on Windows.

To illustrate the behavior, I created the minimal example below. The square task includes a `sleep(10)`  so that it's easy to observe whether the split tasks are executing concurrently.

Without the nested workflow (square_numbers), the work appears to be distributed across multiple workers and completes in roughly 13 seconds. However, when the same task is wrapped in a nested workflow (square_numbers_nested), execution appears to become sequential, taking roughly 56 seconds.

```python
from time import sleep, time
from pydra.compose import workflow, python

@python.define
def square(value: int) -> int:
    sleep(10)
    return value**2

@workflow.define
def square_wrapped(value: int) -> int:
    squared = workflow.add(square(value=value))
    return squared.out

@workflow.define
def square_numbers(values: list[int]) -> list[int]:
    squared = workflow.add(
        square()
        .split(value=values)
        .combine("value")
    )
    return squared.out

@workflow.define
def square_numbers_nested(values: list[int]) -> list[int]:
    squared = workflow.add(
        square_wrapped()
        .split(value=values)
        .combine("value")
    )
    return squared.out

if __name__ == "__main__":

    values = [1, 2, 3, 4, 5]

    print("Running unnested workflow...")

    start = time()
    square_numbers_workflow = square_numbers(values=values)
    result = square_numbers_workflow(worker="cf", rerun=True)

    print(f"Total time: {time() - start} s")

    print("Running nested workflow...")

    start = time()
    square_numbers_workflow_nested = square_numbers_nested(values=values)
    result = square_numbers_workflow_nested(worker="cf", rerun=True)

    print(f"Total time: {time() - start} s")
```

The output is:

```
Running unnested workflow...
Total time: 13.332271099090576 s
Running nested workflow...
Total time: 55.96804141998291 s
```

I expected that the nested workflow would exhibit similar parallel execution to the unnested version.

Is this expected behaviour, or am I missing something about how nested workflows and spliting works?

svirvel:
## Types of changes

## Summary
Changes expansions of asyncronus workflows so that nested workflows are submitted as `Asyncio.Task`. This allows nested workflows to be split across multiple processes instead of waiting for a single nested workflow to finish before continuing. 

This closes #869.

## Checklist

### DIFF
--- a/pydra/engine/submitter.py
+++ b/pydra/engine/submitter.py
@@ -433,15 +433,16 @@ class Submitter:
                                 )
                             raise RuntimeError(msg)
                 for job in tasks:
-                    if job.is_async:  # Only workflows at this stage
-                        await self.worker.submit(
-                            job, rerun=rerun and self.propagate_rerun
-                        )
-                    elif job.checksum not in futured:
-                        asyncio_task = asyncio.Task(
-                            self.worker.run(job, rerun=rerun and self.propagate_rerun),
-                            name=job.checksum,
-                        )
+                    if job.checksum not in futured:
+                        if job.is_async:  # Only workflows at this stage
+                            coroutine = self.worker.submit(
+                                job, rerun=rerun and self.propagate_rerun
+                            )
+                        else:
+                            coroutine = self.worker.run(
+                                job, rerun=rerun and self.propagate_rerun
+                            )
+                        asyncio_task = asyncio.Task(coroutine, name=job.checksum)
                         task_futures.add(asyncio_task)
                         futured[job.checksum] = job
                 task_futures, completed = await self.fetch_finished(task_futures)


pydra/engine/submitter.py
```diff
@@ -433,15 +433,16 @@ async def expand_workflow_async(
                                 )
                             raise RuntimeError(msg)
                 for job in tasks:
-                    if job.is_async:  # Only workflows at this stage
-                        await self.worker.submit(
-                            job, rerun=rerun and self.propagate_rerun
-                        )
-                    elif job.checksum not in futured:
```
svirvel, line 440: My suggested change will add the future job check for nested workflows as well. I'm not sure if this has any implications. I've run all the tests, and none (which weren't already failing and seem to be related) are failing. It would be great to have some input here.

